PR #22, opened 2015-01-29
Update GPSR.tex

egeerts:
Two points that I would like to see different:
1.: About the type of actions.
Last year at the WC only the action 'Bring an object from location1 to location2" was possible and that give already lots of troubles to most teams. Now there are lots of new action examples introduced which makes it even harder to understand all the possible actions. Some of them are quite unclear and therefore I propose some adjustments in the file.

Remarks:
Example-action1: Go to the kitchen, find a person and follow her (missing object-interaction part).
Remark on example-action1: The last part 'follow her', until the time ends..? or until she says stop? Very unclear.

Example-action2: Go to the dinner-table, grasp the crackers, and take them to the TV.
Remark on example-action2: Should the robot place the crackers on the TV in this example? Very unlikely that this must happen in real life. Since when does a TV like to have some crackers?

Example-action3: Offer a drink to the person at the door.
Remark on example-action3: I assume that the robot needs to know where the drinks are stored. This information will be given? 

2.: In the score system the robot has only one attempt to understand the operator correctly. Otherwise no points. To my opinion, a robot should have more chances to score points, especially since speech recognition is such an important task in this challenge.

### DIFF
--- a/GPSR.tex
+++ b/GPSR.tex
@@ -30,13 +30,13 @@ The command is composed by three actions, which the robot has to show it has rec
 \paragraph{Command examples}
 \begin{itemize}
 \item Go to the bedroom, find a person and tell the time (missing object-interaction part).
-\item Go to the kitchen, find a person and follow her (missing object-interaction part).
-\item Go to the dinner-table, grasp the crackers, and take them to the TV.
+\item Go to the kitchen, find a person and follow her to the livingroom (missing object-interaction part).
+\item Go to the dinner-table, grasp the crackers, and put them on the tv-table.
 \item Go to the shelf, count the drinks and report to me.
 \item Take this object and bring it to Susan at the hall.
-\item Bring a coke to the person in the living room and answer him a question.
+\item Bring a coke to the person in the living room and answer his question.
 % \item Bring me the crackers to the TV.
-\item Offer a drink to the person at the door.
+\item Take a drink from the kitchentable and offer it to the person at the entrance.
 \end{itemize}
 
 \subsection{Additional rules and remarks}
@@ -68,6 +68,8 @@ The maximum time for this test is 8 minutes.
 	\textbf{Action} & \textbf{Score} \\ \hline
 	\textbi{Performing the task}  \\
 	Understanding the set of actions on the first attempt & 4.0 \\
+	Understanding the set of actions on the second attempt & 2.0 \\
+	Partly understanding the set of actions & 2.0
 	Performing the first command correctly & 4.0 \\
 	Performing the first and second command correctly & 6.0 \\
 	Successfully solving the complete task & 10.0 \\


kyordhel:
Dear Erick,
Let me add some light to your questions.
Bring an object from A to B was considerd problem solved since 2009. Thereore, it was included to GPSR which demanded the execution of 3 actions in less than 5 minutes, or to demand additional information and solve a simple task.
The reason on which the possible actions are unclear is because we expect a human-like behaviour on robots (they may ask questions if they do not understand). Giving too many details will turn us back to state machines, what is exactly what we are trying to get rid off.
Example action 1: Go to the kitchen, find a person and follow her.If I was the robot, what should I do? To follow that person until she asks me to stop (and perhaps continuing until being threaten or slap), or stp when I get bored. Both options are valid.
Example action 2: Go to the dinner table, grasp crackers and take them to the TV.Well, I was watchin the TV when suddenly felt hungry, then I pause the movie, address to the kitchen, and suddenly: a robot (or your little brother, suits the same for the example). You just smile and ask to bring you the crackers to the TV. How do you solve this task?
Example action 3: Offer a drink to a person at the door.As is stated on the rulebook since 2011 (perhaps before, in the current rulebook under General Rules->Scenario->Objects on 3.2.5) each object has a category and each category has a placement. All robots already know where take drinks from when no more information is given. 
Regarding scoring. I don't see where it states for all test that robots have only one chance to understand the operator correctly (other than the Speech Recognition & Audio Test). Point 5.1.3 of GPSR is clear enough. If the robot does not understand the operator o the first attempt, you use custom operator and get half points for the speech recognition part (2 instead of 4). If the robot does not understand the custom operator (3 attempts), it may request the Continue Rule but the overall score will be reduced to half.
Your comments will be discussed on the following TC meeting. If we found it suitable, it wil be the TC who make the ammends on the rulebook.
With best regards,Mauricio Matamoros,In behalf of RoboCup @Home Technical Committee

  
    
    
  
  
 		 	   		  =

egeerts:
Dear Mauricio,

Thanks for your response!

I understand that in general the idea is to get rid of the statemachine behaviour. The robot should be more generic in its actions and be capable of interleaving actions more fluently. I agree, and therefore this challenge is very complex. To my opinion there is still a middle way to reach this goal and not make the step too big compared to last year where only one type of action was possible (see https://github.com/RoboCupAtHome/gpsr_command_generator/blob/master/cat2Sentences.txt). 

I have still some remarks/questions:

1. Last year not a lot of points were scored on the EGPSR challenge, although only one type of action had to be heard. Now lots of type of actions need to be heard in one long sentence and the challenge is moved to Stage 1. I expect lots of robots that cannot understand the sentence and cannot execute the action, even with the CONTINUE-possibility. To my opinion this challenge can therefore not be compared with the other challenges of Stage 1.

2. With the given examples I can expect (hundreds of) thousands of possible ways to ask a robot/someone to do something: e.g. example in gpsr: "Go to the shelf, count the drinks and report to me." 
Could also be asked in following way:
"Report the number of drinks on the shelf to me."
or
"Tell me how many drinks can be found on the shelf."
etc.
All these possibilites makes the speech recognition very complex. Google already solved this problem mainly. Their speech recognition is very good, but during a tournament a solid internet connection can stil not be guaranteed. 
I expect that the current speech command generator will be updated intensively to generate all these possibilities?

3. Last years the robot had to show that it has recognized the command by repeating the command.
In the current rulebook is stated: "The robot may repeat the understood command and ask for confirmation." Does this mean that the robot does not need to repeat the command anymore? Just say, "I understood your command"? If not, then I would suggest to adjust the rules and give points to the robot if the robot repeats the command in its own formulation.
E.g. command is: Give me a coke from the shelf.
Robot repeats: I will get you your coke. 
If the robot then goes directly to the shelf and gives you your drink, then this repeating of the sentence is okay to my opinion. What are your thoughts?

Best regards,
Erik Geerts

kyordhel:
Dear Erik,
Before continuing with this discussion, pleasse refer to RoboCup @Home Rulebook 2011, Section 4.5 and review it. For your convenience you may find the link below
https://docs.google.com/viewer?a=v&pid=sites&srcid=cm9ib2N1cGF0aG9tZS5vcmd8cm9ib2N1cC1ob21lfGd4Ojc1MGI5Nzk0ZTA2ZjdiZjM
You will notice that we are re-taking a 4 years old test, and granting the robot will perform navigation, object-interaction and HRI as tasks. Also, commands may not be given explicitly, but requiring some inference.
A new generator will be developed for this test which may not be based in the old one. As with a human, it is always possible for the robot to ask for missing information, propose a solution or explain the operator what it is going to do.

  
    
    
  
  

About Google's speech recognition system. Remember that robots mus be self-contained and all the computers required to perform the task must be inside the arena during the test (Rulebook 2015, Section 3.4, subsection 7). Unless you can move google servers inside the arena, this is not a valid option for processing.
Again, your comments will be mentioned during the next TC meeting. I stress the point that this competition is intended to be challenging, not to stay on the state-of-the-art on average teams capabilities.
With best regards,Mauricio Matamoros. 		 	   		  =

LoyVanBeek:
Hey @egeerts, we discussed the following today about the GPSR:
- We will only ask the robot to do sensible things. Bring crackers to the tv is indeed uncommon. We want the robots to have common sense, we will have common sense too. 
- Indeed, there are lots of ways to state the same command. This is exactly the point. We know it is difficult too. We want natural interaction though and thus push the state of the art. 
- Interaction should be natural

PR #134, opened 2015-10-29
Feature/qr code continue

LoyVanBeek:
A standardized alternative input will make it easier for the TC to let the robots do something. 
It is not intuitive, but that is the reason for a high penalty. 

Perhaps another alternative to ASR may also be provided that is intuitive to use with a smaller penalty.
That is difficult to do, while adding QR-capabilities to the robot is relatively simple.

### DIFF
--- a/ExampleSkills.tex
+++ b/ExampleSkills.tex
@@ -42,7 +42,7 @@ Please note that these examples are to illustrate the level of complexity and ap
 	\item Collaborative robot-human manipulation.
 	\item Maintaining a conversation.
 	\item Learning actions on-the-fly.
-	\item Learning objects from humans e.g. ``This object is a ...'' with an open vocubulary.
+	\item Learning objects from humans e.g. ``This object is a ...'' with an open vocabulary.
 	\item Following a human by grasping its hand.
 	\item Explain the robot abstract concepts (why people love sunny days).
 	\item Arrange unknown random people for a nice photo (no occlusions).
--- a/Setup.tex
+++ b/Setup.tex
@@ -116,7 +116,8 @@ Safety is the most important issue when interacting with humans and operating in
 		\item voice of the robot (it must be loud and clear)
 		\item robot speaker system (plug for RF Transmission)
 		\item use of external devices (including wireless network)
-		\item usage of continue rule \label{rule:asrcontinue}. The robot must stop anywhere between the entry and exit and wait for someone to say ``Continue''. Instead of actually saying anything, the team must demonstrate to the TC how to use the continue rule for the robot. 
+		\item usage of (standardized) continue rule (\ref{rule:asrcontinue}). The robot must stop anywhere between the entry and exit and wait for someone to say ``Continue''. 
+		  Instead of actually saying anything, the robot will be shown a QR-code encoding the text ``Continue'' as per the standardized Continue-rule. 
 		\item other safety issues (cables hanging loose etc.)
 	\end{itemize}
 	\item \textbf{Re-inspection:} If the robot is not approved in the inspection, it is the responsibility of the team to get the approval (later). Robots are not allowed to participate in any test before passing the inspection by the TC.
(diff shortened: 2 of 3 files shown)


kyordhel:
I'm correcting the conflicts and merging your pull request, however, as discussed only QR and intuitive ASR alternatives are allowed. Therefore in file *general_rules/Procedure.tex* starting in line 231

The alternative methods are listed but stating that no keyboard typing/plugin-laptop will be allowed since pluging hardware is not intuitive. Same applies to the use of *rostopic* and *ssh*.

We should add a rule for the operator having all devices for the continue Rule before the test to avoid cheating.

balkce:
I'm not comfortable making QR mandatory. I'm fine with providing that as an option to the teams, and the penalty to use it is good, but I don't think a lot of teams will like the idea of spending precious development time in creating a module with 20 year-old vision technology.

In addition, if we're not letting the teams use keyboards because "they are not intuitive", then I could make the case that neither are QR codes. I'm fine with letting the keyboard be used, with a severe penalty obviously. Lets remember that the continue rule is so that the test can proceed, it doesn't REQUIRE to be pretty. And, I want the part of the penalties for alternative continue interfaces to emphasize that there is a chance that the penalties will be taken away if the TC decides its intuitive/natural. Lets make this continue rule something that the teams welcome, not force onto them.

LoyVanBeek:
The only time I want to use the QR-codes is as a last result for a challenge where we're not concerned about speech, so in a lot of stage 1 tests except Speech Recognition.
The reason to require this is that its standard. And I've seen challenges messed up because typing on the commandline or wherever under time pressure on a weirdly positioned keyboard also sucks. 

Maybe not make it required indeed but a highly recommended standard. Adding it is simple: there's a bunch of ROS nodes and otherwise ZBar to do the QR-reading.

balkce:
Agreed then to not make QR reading a requirement, but a recommendation.

Actually, if the idea is to standardize it, the league should develop the QR-reading module such that teams only need to download it, compile and run. Or maybe we should point inside the rule book to those ROS nodes you're referring to, such as: http://wiki.ros.org/zbar_ros

kyordhel:
We can always rollback to previous commit. 

About being intuitive: I think virtually all volunteers will find showing a piece of paper with a QR to the robot, or a tablet we provide with the QR on screen. The system is ready and operational. On the other hand, having people (team members) inside the arena plugging a keyboard and typing 

rostopic ...

Or from a remote computer:

ssh user@doman...

Is for me, unacceptable. There will be BIG penalties for QR, while other interfaces need to be discussed on place (e.g. pointing to an object instead of saying go get it may even have no penalty). 

Still, I think that if this is going to be a standard alternative (I didn't found in the rulebook it was mandatory) we must also provide an official open-source QR module/library integrated with ROS.

Best!

balkce:
What about when the keyboard is already plugged in? There are quite a bit of robots that have a laptop with a keyboard installed. In the last version of the continue rule it was made clear that if typing it out is required, it should only be the literal command and pressing the Enter button at the end; no coding or additional command should be required. In my experience there have been several robots that carried out the continue rule in this manner and it took them more or less the same time that I suspect a TC member will need to generate the QR code, since they would need to type it out into the official QR-code-generating PC anyway.

This is, of course, unless we have a pre-made dictionary of most popular QR codes, but that would mean that we would need to be shuffling for them whenever we need them (either the print-outs or in the image files inside a computer). And even then, we will need to generate many QR codes on the spot (EGPSR for example) every time there is an ASR command, meaning that we would need to always have a TC member be near a laptop specifically just to generate QR codes. Yeah, not a fan; I think we already have enough to do in the arena as referees.

About it being mandatory in the rule book: Section 3.7.15 (Continue rule: Bypassing Automatic Speech Recognition), 2nd paragraph: "All robots are required to provide QR-codes as a standardized alternative to ASR."

To be fair, I had a look into the zbar_ros module, and, as Loy indicated, it seems pretty straightforward. That module could be the way to go if we decide to QR-code the world.

dirkholz:
Guys,  what about just adequately integrating the "alternative intuitive
interface" (TouchPad etc.)? I'm pretty sure that will be a game changer. No
more need for continue rule: "use speech or the interface, if both don't
work,  it's your system that sucks ..." :)

balkce:
Not it for developing it. 

I do like the idea of having a standard alternative way for ASR, but I'm curious of what teams may come up with too and having a "standard" may stifle that.

kyordhel:
+1 to Dirk's comments. We are making a mess of it. If is simple, it works!

  
  



Caleb, I'm OK with an onboard keyboard, I just don't want teams running all around looking for a keyboard/computer to plug-in. Just type and enter is OK for me as well.

LoyVanBeek:
If we're going to use QR-codes, all the commands we say should be generated and a QR-code simultaneously. Typing on a commandline under pressure is error prone, I don't want to do that to teams. 
I prefer a custom intuitive continue rule over QR-codes, and thats why QR-codes should be heavily penalized. With the custom continue-rule, you get, say, 80% of the score, with QR-codes only 20%.

PR #136, opened 2015-12-07
Feature/datarecording

LoyVanBeek:
Record data during challenges for benchmarking purposes. 

Not sure yet how to incentivize this, because giving a fixed amount of points and a percentage both have their downsides. 
Open to ideas :confused:


kyordhel:
I think we should add a section containing all data formats and how data should be provided.

Per team and for each test I find useful (when applicable):
- PCM wave files AND transcript (transcript is for us).
- Map (topological?) showing the path of the robot and the action performed at keypoints.
- RGB-D data on where the robot performed object recognition.

Per team (once)
  - RGB-D models of the Known/Alike objects.
  - Map of the arena (point-clouds?)

More information than that, I think would be useless. We may encourage participation by offering an award for providing:
  - All wave files.
  - RGB-D models of objects.
  - Maps.

After that, the feature would be complete and we can proceed with the merge, I think.

As final remark, I'm not sure I would like to force teams to share data, but only those with something yo share, so we may have a database with good maps, decent audio files and lots of object models with scenarios for training. I find no use (yet) on logging action planning since is up to the teams how to solve a task and can't see TC doing desktop work to decide if a plan is good or not.

balkce:
So, basically, we're asking the teams to run rosbag. Sounds simple enough.

kyordhel:
A little bit more complicated than that @balkce. We also need to provide the infrastructure for non-ROS teams to be able to provide data (RoCKIn people already did that).

balkce:
We could base it on their work. I think they have it open source'd somewhere.

kyordhel:
They have it @balkce, what we are doing is trying to incorporate RoCKIn stuff into @Home

LoyVanBeek:
@balkce indeed, running rosbag :-). Not everybody can do this though and too much data may not be helpful either. We also have to store and upload it somewhere too. 

My main use-case for this data is to test and train speech recognition, object recognition and maybe mapping better outside RoboCup. 

For an incentive: I'd give a small percentage bonus for delivering all listed data per challenge.

LoyVanBeek:
Slightly updated the pull request. 

- I want to have images for every recognition to get images for a lot f different situations, so not once per team.
- Same goes for the maps, because we will **shuffle** the furniture and so the map is sligly different every time
- Data logging format is clear enough for now, I think. I don't know how to make it more precise right now.

kyordhel:
I don't like shuffling furniture for every test. Small changes is ok though. Maybe push the coach, move a chair...

About the data logging formt, just copy specs from RoCKIn

LoyVanBeek:
Thats exactly what I want: move the furniture around a little bit, e.g. turn a table 90 degrees or so. 
Regrding datalogging specs: their rulebook does not state such specs, says they will be announced before the competition

PR #137, opened 2015-12-08
Reworked EEGPSR

kyordhel:
Changes as follows:
- Major changes on test description, structure and spirit remains.
  - Test structure modified accordingly to previous years Enhanced GPSR and Endurance GPSR
  - Added EEGPSR appendix including example commands and abilities to be evaluated.
- EEGPSR Scoresheet updated
  - Max score set to 500 since is replacing 2 tests

I tried to include all suggestion from the Ideas GDoc and issue #133 

Please review the test before merging.

Issue #133, opened 2015-10-29
Enhanced Endurance GPSR in Stage 2


kyordhel:
*Wake me up* & *RoboNurse* will be replaced with a *Enhanced Endurance GPSR* test.

- 2 robots in parallel
- Endurance: 
  -Duration: ca. 40 minutes
  - Take into account history: if an object was moved A to B during the challenge, next time grab the object from B instead of A.
    -For now, only if the robot moved from A to B for some command
- Enhanced: Large set of commands
- Robots must respond to events happening (we might restrict this to event happening only when the robot is looking, later extend to robot just being in the room, robot being anywhere in the apartment)
  - Events: 
    - Falling
    - Walking away
    - Standing up
    - Sitting down
    - Waking up
    - ...
- Skills to show in EE-GPSR
  - Waking someone up
  - Describing objects
  - Grab described object
  - Natural handover
  - Pour into a bowl

There is no story described to the teams beforehand. For the audience, we make a story using the commands. E.g. user wakes up his partner

GPSR can be boring when robots are unable to do a task. 
- Let robots tell if they cannot do a task
- Different levels of difficulty:
  - Stage-1 GPSR test 1 command in different difficulties of Natural Language Understanding with skills shown in Stage-1
  - Stage 2 EE-GPSR tests NLU and skill complexity of Stage-2. 

Giving all robots the same commands (to make it fair) is also boring. Instead, score on separate skill-categories instead of scoring all points or not for the whole command. (Ow, if you achieve 90% but fail in the last 10% of the command)
Manipulation, Gesture recognition, Object Recognition, Navigation, …. If the whole command was completed, e.g. + 10%. 

- Three commands will be given to the robot.
  - Command 1 belongs to old Cat 1
  - Command 2 belongs to old Cat 2 and will require information from Command 1
  - Command 3 belongs to old Cat 3 and will require information from Command 1 or 2

kyordhel:
It is done!

PR #138, opened 2015-12-10
Reworked EEGPSR according to #133 & #137

kyordhel:
Changes as follows:

- Major changes on test description, structure and spirit remains.
  - Test structure modified accordingly to previous years Enhanced GPSR and Endurance GPSR
  - Added EEGPSR appendix including example commands and abilities to be evaluated.
- EEGPSR Scoresheet updated
  - Max score set to 500 since is replacing 2 tests

I tried to include all suggestion from the Ideas GDoc and issue #133 and corrections from #137

Please review the test before merging.


tests/EEGPSR.tex
```diff
@@ -1,129 +1,168 @@
...
+		\item If the robot appears to consistently not be able to understand the operator, the referees ask the team to continue with a custom operator, reducing the obtained score for that command to 75\% (\refsec{asrcontinue}).
+		\item If the robot appears to consistently not be able to understand the custom operator, the referees ask the team to continue with the alternate method, reducing the obtained score for that command to 25\% (\refsec{asrcontinue}).
+	\end{itemize}
+
+	\item \textbf{Inoperative robots:} If a robot gets stuck while trying to accomplish a task during a reasonable amount of time (e.g.~30 seconds), the referee may ask the team to move back the robot to its designated position, proceeding with the next robot. \\
+
+	\item \textbf{Restarts/Changing/Charging batteries:} The team may install a charging station at the designated position of the robot, if it does not hinder the other robots. However, the robot must connect itself with the charging station after carrying out a command. Changing batteries or manually connecting the robot with the charging station is allowed during a restart. The general rules for restart apply. \\
```
LoyVanBeek: I like the idea, but only recharging during a restart is kinda harsh. On the other hand: a robot should last 30 to 40 minutes on a charge. Maybe give the teams 1 opportunity for a manual battery change taking up to max. 1-2 minutes.

kyordhel: I don't mind. I just copied the text from previous Endurance GPSR "as is", but as @dirkholz once said: *we don't want teams running with cables and bateries inside the arena and we should push teams to have robots that last more than 10 minutes running*

PR #180, opened 2016-04-09
No guiding in nav but door opening

LoyVanBeek:
Door opening instead of guiding in Navigation test. Guiding is also tested in 2 other tests still. 
The way it is defined now is not difficult and not optional. The 40 points should be a big enough incentive. 

We discussed this at RoboCup European Open 2016 with Loy, Mau and Sven.

### DIFF
--- a/scoresheets/Navigation.tex
+++ b/scoresheets/Navigation.tex
@@ -23,7 +23,8 @@ The maximum time for this test is 5 minutes.
 	% Up to 50 points
 	\scoreheading{Waypoint 4-3: Go back back home}
 	\scoreitem{5}{Reaching the door via which the robot went out the appartment again. The handle must be within manipulator reach}
-	\scoreitem{40}{Opening the door.}
+	\scoreitem{20}{Pushing the door handle/knob/lever with manipulator}
+	\scoreitem{20}{Opening the door far enough to drive in}
 	\scoreitem{5}{Reaching waypoint 3 after reentering the arena}
 	
 	% Up to 50 points
--- a/scoresheets/Navigation.tex
+++ b/scoresheets/Navigation.tex
@@ -3,10 +3,10 @@ The maximum time for this test is 5 minutes.
 
 \begin{scorelist}
 
-	% Up to 60 points
+	% Up to 50 points
 	\scoreheading{Waypoint 1: Path planning}
 	\scoreitem{5}{Planing new trajectory after the door has been shut}
-	\scoreitem{50}{Opening the door and continue instead of planning a new trajectory}
+	\scoreitem{40}{Opening the door and continue instead of planning a new trajectory}
 	\scoreitem{5}{Reaching waypoint 1}
 
 	% Up to 15 points
@@ -20,9 +20,11 @@ The maximum time for this test is 5 minutes.
 	\scoreitem{20}{Reaching waypoint 4}
 	\scoreitem[10]{-2}{Penalty for asking \textit{Professional Walker} to wave to resume following.}
 
-	% Up to 10 points
+	% Up to 50 points
 	\scoreheading{Waypoint 4-3: Go back back home}
-	\scoreitem{10}{Reaching waypoint 3 after reentering the arena (i.e. after reaching waypoint 4)}
+	\scoreitem{5}{Reaching the door via which the robot went out the appartment again. The handle must be within manipulator reach}
+	\scoreitem{40}{Opening the door.}
+	\scoreitem{5}{Reaching waypoint 3 after reentering the arena}
 	
 	% Up to 50 points
 	\scoreheading{Obstacle avoidance}
@@ -35,7 +37,7 @@ The maximum time for this test is 5 minutes.
 	\scoreheading{Leaving the arena}
 	\scoreitem{5}{Leaving the arena}
 
-	\setTotalScore{170}
+	\setTotalScore{200}
 \end{scorelist}
 
 % Local Variables:
--- a/scoresheets/Navigation.tex
+++ b/scoresheets/Navigation.tex
@@ -1,5 +1,5 @@
 
-The maximum time for this test is 5 minutes, 5 more for the optional guiding-phase.
+The maximum time for this test is 5 minutes. 
 
 \begin{scorelist}
 
@@ -20,12 +20,9 @@ The maximum time for this test is 5 minutes, 5 more for the optional guiding-pha
 	\scoreitem{20}{Reaching waypoint 4}
 	\scoreitem[10]{-2}{Penalty for asking \textit{Professional Walker} to wave to resume following.}
 
-	% Up to 40 points
-	\scoreheading{Waypoint 4-3: Take me back home}
-	\scoreitem{15}{Locate the operator and resume guiding after two people interference}
-	\scoreitem{10}{Reentering the arena}
-	\scoreitem{5}{Reaching waypoint 3 after reentering the arena (i.e. after reaching waypoint 4)}
-	\scoreitem{10}{Reaching waypoint 3 with the operator (i.e. not having lost the operator)}
+	% Up to 10 points
+	\scoreheading{Waypoint 4-3: Go back back home}
+	\scoreitem{10}{Reaching waypoint 3 after reentering the arena (i.e. after reaching waypoint 4)}
 	
 	% Up to 50 points
 	\scoreheading{Obstacle avoidance}
@@ -38,7 +35,7 @@ The maximum time for this test is 5 minutes, 5 more for the optional guiding-pha
 	\scoreheading{Leaving the arena}
 	\scoreitem{5}{Leaving the arena}
 
-	\setTotalScore{200}
+	\setTotalScore{170}
 \end{scorelist}
 
 % Local Variables:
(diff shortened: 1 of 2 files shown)


kyordhel:
I'm not mergingthe pull request, some changes need to be done:
- Five minutes is quite short for obstacle avoidance, following human, navigating back, opening door and exiting. I could do that in 3 minutes average so robots need no less than 8. I would advice even 10.
- Score: 40 points for opening the door is quite ambiguous.
  - 20~25 pts for pushing the door handle/knob/lever with the manipulator. No need for grasping, just functional touching & pushing as we discussed it in eindhoven. Thats where we are pushing towards.
  - 20~25 pts for opening the door quite enough to get in.
- Remarks on functional touching must be more specific and state that **it must be clear the robot is trying to solve the task, not just brute-force entering**

About touching the handle. I know this test is about navigation, but opening a door requires both skills so we grant score for navigation part, and also a reward for those teams which dare to do more.

Those are my comments.

balkce:
I agree on removing the following/guiding phase in navigation. It's
redundant with other tests.

C

LoyVanBeek:
@kyordhel like in 25a6961 ?

semeyerz:
I would vote for removing following and guiding from navigation as well to remove redundancy.

In my opinion we should really think of a way to give more points for fast and save navigation.
Especially in Leipzig with the 200qm² arena we can test navigation over long distances pretty good.

Version 459737c gives half of the points for person recognition and version 25a6961 gives half the points for door opening.
Why not give more points for navigation itself?
Even by keeping the time short and thus penalizing slow navigation.

My Suggestion:
- more points for navigation/waypoints
- multiple small objects with penalties when touched during navigation
- Giving points for door opening only ones

Reason for that is that I think navigation is something most of the robots can do somehow and it would be nice if the points represent that.
Especially for new teams it could be quite confusing if they worked a lot on getting the robot navigating good and in the end they receive only 10 points as there robot was not able to solve all the tasks that are not related to navigation.
For that we have manipulation, person rec and so on.

My suggestion: https://github.com/semeyerz/RuleBook/blob/fd8659995005da8ac2960821f2307c17240a3eb7/scoresheets/Navigation.tex

kyordhel:
@LoyVanBeek Exactly like in 25a6961 !!!

@semeyerz Navigation is problem solved. We are not scoring robots for moving around. New challenges are avoiding real objects, online mapping and localization. On the other hand, taking times and normalizing makes scoring very difficult, specially in a test where multiple path can be taken.

There is a rule: safety first. Pushing team to move faster and just deducting points for each collition doesn't seem to go in that direction. Pumas' robot Justina can drive up to 1.5m/s which is fast as hell for a 80kg robot, At that speed it could easily finish the test in seconds by ramming over the arena. 

IMO: Precision first, then speed. Once no robot touches an obstacle, we start to speed up things.

LoyVanBeek:
This kind of simple navigation is mostly solved. Moreover, if you can open a door, sometimes navigation becomes easier, faster or simply possible. 
I do agree its not the primary focus, therefore its at the end. If your not good at navigation already, you won't even get to this door opening.

semeyerz:
> Navigation is problem solved.

In general I agree with that.
There is still a long way to go and a lot to improve.

> We are not scoring robots for moving around.

But we could score them for moving efficient and without pumping into something.
This is at least what the name of the test suggests?

>  taking times and normalizing makes scoring very difficult

Of course. This is why I suggest to keep the time short. Then scoring comes by not getting points if your robot is not doing the right thing in time.

> There is a rule: safety first.

And again I completely agree.

> Precision first, then speed

This is why I suggested to add different kinds of small objects.

> open a door [...] its not the primary focus, therefore its at the end.

Sounds good.  25a6961 suggests having it in the beginning for 40 points and at the end for another 40 points. Giving 80 points in nav for door opening 2 month before the competition might lead to a lot of hardcoded door opening and not so much real contribution.

I think the rulebook already favors manipulation by giving 300 points for manipulation task.
Now giving 80 more in the nav task that already only gives 200 really favors teams/robots that focus on manipulation.

Even if navigation includes some solved tasks I still like to represent that in the points.
This makes it easier for the public to understand what robots already can do and what they can't.

LoyVanBeek:
Good point on the relative scoring.
We can do the following things:

1. - Remove the opening of the internal door for 40 points
2. - Leave the external door open if the robot already opened the internal one. Makes no sense story-wise and conditionals in the scoring sucks.
3. - Leave it in! In a big arena, opening a door rather than traveling a lot can be beneficial, especially when later speed becomes a factor. 

With the door opening, I think we do show that robots can do navigation mostly, but door opening and manipulation not so much. 
I think given the conditions, hardcoding (open loop trajectory of the gripper) will fail, I think. If not yet, we should add conditions to require closed loop. 

I'm mostly concerned with adding this so late to the competition. 

BTW: the manipulation & object recognition also includes a lot of points for object recognition, the 300 are not all for manipulation.

kyordhel:
Ok, be prepared for (very) long answer.

First, I would like to ask you all to stay calm. Let's not change a test that has been discussed and studied for months just two months before the competition. And please don't think in terms of points. Points are irrelevant. Think in terms of what do we want to achieve. I want the following:

- Robot uncapping a bottle of beer (already done by NimbRo).
- Robot serving breakfast (already done by NimbRo).
- Robot cleaning tables using a cloth (kind of done by Pumas).
- Robot watering plants with real water (already done by NimbRo).
- Robot flipping pancakes.
- Robot cleaning using a broomstick and a mop.
- Robot real time mapping first time enters the arena.
- Robot safely navigating in a room with mirrors, paintings and transparent furniture.
- to name some...

I agree with Sebatian @semeyerz in that the rulebook is mostly oriented in vision and manipulation, in that few points are given for navigation, in that the test is boring and robots are slow. I alo agree in that following and going back is redundant. However, robotics is not a bunch of isolated abilities but all about coordination. Navigate is necessary but no task can be solved just by moving around. Even a RoboCop needs more than patrol, to recognize people and shoot to the bad guy while protecting the owner.

So, going back to the Navigation Test. _What's the goal of the test?_ Having robots **navigating safely while avoiding obstacles and moving through rooms**. Sorry, that involves dealing with doors, we do it every day that way. In my opinion, following and guiding is more vision/perception than navigation since involves object (people) tracking and object (environment) recognition, but let it be. _What is the goal of this year?_ I've been pushing towards the **serving the breakfast** in a tray and navigate while **transporting the tray**, and that's tested in Restaurant and in EEGPSR. Also, **opening the door** has been in the schedule since 2009, so is time to solve it.

This test was revisited during European Open by Sven, Loy and I, and we agreed on reducing the guiding to going back and make mandatory opening the door at the end of the test. Nonentheless, lets revisit the test again, lets do a Brainstorm! Specifically,_ what do you want the robots to do and how can we test it with a minimum setup_?

Right now the test is:
- Navigate to A
- Navigate to B
- Navigate to C
- Follow to D
- Go back to C (requires open door)
- Leave.
- Do it while avoiding all obstacles on your path.

And that's why there is no predefined order for the obstacles. The robot must have all obstacle avoidance and path planning software running all the time. Can the robot detect an obstacle is a person? Awesome, interact! No? Take another route. Can the robot detect an object is functional-pushable? Awesome, interact! No? Take another route. Add more obstacles sounds easy since we can jump over it, but we could also create easily an scenario in which a robot is unable to go through because the (small) obstacles are too close one from each other. Plus, setting the arena this way is hard to rearrange and teams already know the setup before start.

So, what can we change then? What else can we test?
**Brainstorm here**

If we are going to change anything, it needs to be done before May, but I'm against.
@swachsmu @balkce @robotijn @dirkholz @dragonjack Feedback please.

balkce:
Wow. You guys really like discussing. 

Let's just remove the following part at the end of the test and have the robot optionally open a door to exit the arena with some big points.

This is Stage 1, we can't force teams to do something that no robot has done in an international competition.

LoyVanBeek:
OK, the verdict is: this pull-request is going in. 

1. Navigation is mostly solved: all robots in RoboCup@Home can do it and most other tasks are more difficult, so the relative scoring should reflect that. 
2. Opening doors has been optional for a long time without anyone trying even. I've not seen this happen at RoboCup in the 5 years I attend. The only change made now is making it more mandatory and accurate specification. 
3. Guiding indeed is redundant, this is tested elsewhere.

balkce:
I just checked the merge. There is still a guiding phase at the end of the test. I thought we were taking that out.

LoyVanBeek:
Yes we are, thnx for the remark. I am re-reading the rulebook just to spot this kind of errors.

balkce:
All gone now as far as I've checked. 

I'm still actually uncomfortable with the following phase, since we are also testing that elsewhere. But I take what I can get.

kyordhel:
The follwing phase is tested redundantly here becuse we want to be sure there will be no problems when robots go into the wild, surrounded by normal people, screaming children, and the gods know which other fauna.

I don't like eaither but is  necessary.

balkce:
Wait. Are we putting this as a pre-requisite for the robots in the wild test? Meaning if robots don't do at least decently here they don't get to do the robot in the wild test? Didn't read that. 

If so, sounds reasonable then.

kyordhel:
Is not written since we can't make mandatory to _pass_ the Navigation test before doing the RoboSafari (Follow & Guide) because all are Stage 1 and it heavily depends on the schedule, but we will schedule Navigation first and try to convince teams of not participating because of security concerns.

Also, they need to guide the robot from the entrance of the arena through the team area and to the wild, so we have robustness & fault tolerance there also.

Issue #173, opened 2016-04-05
Networking setup


LoyVanBeek:
MSL does this as well, see the last pages of their rule book on http://wiki.robocup.org/images/0/0f/Robocup-msl-rules-2016.pdf We should do the same thing and announce this beforehand.

komeisugiura:
Nagoya LOC would like to prevent the cases where robots are not working because of the wifi interferences. We will strictly assign the channels, and pay money to an external company to manage this. We also pay money to hire "wifi police" to find and stop unofficial wifi APs including iPhone tethering brought by spectators.

Both Pepper and HSR are strongly dependent on network connections, especially Pepper is required to be connected during its booting sequence. 

The details are still under discussion. If we are to use VLAN, the specifications are required to ask the company to estimate the cost. @LoyVanBeek , do you have some ideas on the VLAN specifications?

LoyVanBeek:
Lets assume 36 teams in total. 

- We'll need a VLAN for every team, so the switch/router will need to be capable of handling that many VLANs. 
- In total, we'll have a few devices on the wireless network for each team, so the total network needs to be able to handle all those devices. There is also a bit of configuration to do for each, I suppose. If each teams has 1 robot and just 1 laptop in the arena (thus wireless) that is already 72 hosts on the WiFi (over 3 very busy arena's). 
- Each arena must be covered by the WiFi acces points. Depending on the size of the arenas, that will be 1 or maybe 2 APs per arena. 
- The APs need to be capable of routing robot MAC addresses into their team's VLAN or broadcast separate SSIDs that each map to the VLANs. 
- Each VLAN must have internet access
- Reasonable bandwidth and latency between team desk and robot. The faster the better, as always.

warp1337:
Reviving this after the German Open, where Tim from TU/e found a nicely working solution for the wireless 'situation'. However, we also discovered/discussed some problems concerning a potential setup for >30 teams.

@LoyVanBeek, can you bring Tim into this discussion?

Noticed problems at the German Open (7 teams in the @Home league):

[...]

PR #182, opened 2016-04-21
Networking set up

LoyVanBeek:
See #173

### DIFF
--- a/Appendices.tex
+++ b/Appendices.tex
@@ -10,6 +10,8 @@
 \input{tests/EEGPSR_appendix.tex}
 \newpage
 \input{images/standard_QR_codes.tex}
+\newpage
+\input{general_rules/ip_assignments.tex}
 
 \end{appendices}
 
--- a/general_rules/ip_assignments.tex
+++ b/general_rules/ip_assignments.tex
@@ -1,0 +1,62 @@
+\chapter{Table of network addresses}
+
+\section{General Setup}
+\begin{itemize}
+ \item WEP encryption is turned off.
+ \item Broadcast of SSID is turned on.
+%  \item Subnet mask normal PC: 255.255.255.0.
+%  \item Subnet mask of a PC connected to the Refbox: 255.255.0.0.
+%  \item Access Point Beacon Interval should be set to 20-30.
+%  \item Access Point DTIM Interval should be set to 2-3.
+\end{itemize}
+
+\section{Organisation network setup}
+\begin{table}
+    \begin{tabular}{|l|l|l|}
+\hline
+    ~              & Arena A        & Arena B        \\ \hline
+    SSID (2.4GHz)  & AtHome\_A\_2.4 & AtHome\_B\_2.4 \\ \hline
+    SSID (5GHz)    & AtHome\_A\_5   & AtHome\_B\_5   \\ \hline
+    ~              & ~              & ~          s    \\ \hline
+    \end{tabular}
+\end{table}
+
+\section{Team network address ranges}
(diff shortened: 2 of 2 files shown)


tik0:
Here you assume, that every team shares the same physical switch and (wireless-) network, right?
Shouldn't we even go a step further:

- Setup a vlan for every team to have at least a logical separation
  - This is absolutely necessary due to the promiscues and unsecure software  (ROS, etc.) the teams are using
  - With a vlan, every team can setup their own IP range and even DHCP servers
- WLAN setup
  - The APs should promote a unique SSID for every team, which then routes to the specific vlan
  - This has the nice benefit of switching the SSIDs on/off for the corresponding teams in the current task
- Punishment and fall-back
  - There should be some punishment for not obeying the rule, that a team has to use the official network
  - A decision at the **end of the first setup day** should be made, if the official network is usable for all teams, or not. And if not, are the teams allowed to use their own WLAN and network?
  - If we think practical, the upper point answers it self: Different teams come from different countries, which allows only particular WLAN frequency channels which might not work with the local setup. Thus, if one team is **technically** unable to use the official WLAN, then no one needs to use it, which leads to the question: Is it possible to setup and support an official network/WLAN/VLAN?

LoyVanBeek:
Hi @tik0, good points made there. The setup as it is now is mimicked from the Middle Size League. They can do it and we should be able as well. 

I totally agree on the decision on the end of the first setup day. 

VLANs sound very useful, but I'm not a network administrator and my networking knowledge is not that deep. I asked the OC for some detail yesterday on how they plan to do things. 
If we can use DD-WRT on the wireless routers, VLANs should possible. I'm not sure whether multiple SSIDs is possible on a single channel as well. 

Preferably, there's a person who can arrange and manage this for us :smiley:

semeyerz:
Despite VLANs/Unique BSSIDs are a great idea, we have to decide whether we want to provide Internet and be able to have a Referee-Box soon.

If so it could be quite hard to set up routing to the web and to the ref-box if teams share the same IP-Adresses even in different VLANs.
Teams are allowed to announce external computing they should be able to access Internet from the wifi.
This means we should assign ip-ranges now even if we are able to get a good vlan setting running that separates the teams from each other.

LoyVanBeek:
So, we had a chat about the @Home network today, the TC (me) and some people of Homer, TechUnited and ToBi. 

The setup decided on is using a VLAN with a managed switch:

- The teams register the MAC addresses of their WiFi-enabled devices: with a reasonable max per team. We will not register all the team laptops, just the bare minimum: their robot(s) and 1 or 2 laptops and perhaps a tablet. 
- There is a central switch, with ca. 40 Gigabit ethernet ports
- Each arena has a wireless access point with its own SSID
- From the central switch, there goes 1 cable to each team table. They can do whatever they want on that cable. Everything on that cable (so of that particular team) always gets routed to the MAC-addresses associated with that team, via a Virtual LAN. 
- The refbox (to be, in the future) is also connected to every VLAN so all the teams can access it. 
- Teams teams can put  a switch, router with DHCP server on their table and plug the cable from the central switch into it. And also plug a different cable with internet access into the same router. 

For this, we'll need a managed switch, with the following requirements:

1. VLAN-capable
2. 40 ports or so
3. Perhaps a firewall to limit internet access of the robots

The estimated cost of such a thing is in the 5000-10000 euro range.

rventura:
The network setup proposal looks great. I'd only advise to make sure the WiFi APs are capable of segregating different MAC addresses to different VLANs. It should be the AP to split traffic among different VLANs, not the switch (that'd be too late).

PR #211, opened 2016-12-24
2017/set a table

LoyVanBeek:


LoyVanBeek:
As this is a stage 2 challenge, giving a restart is the least we can do.

PR #224, opened 2017-01-28
Reduce points in sattu

LoyVanBeek:
Setting a table and cleaning it up yielded 990 points, which is out of line with the other stage 2 tests. 
The score is reduced to 390 points by:
- Not yielding points for recognizing objects. That is simply required to achieve the task
- Wiping the table yields less points (40 instead of 100) as it is not 10x harder than grasping
- Less points for hard to grasp objects
- Less grasps and placement actions by adhering to the tests own rules where there are only 3 hard and 3 easy to grasp objects, 6 in total. Those need to be handled twice though.


rventura:
Looks good, Loy. Do I have to press Merge from this side?

LoyVanBeek:
Nope, if you agree, I'll press the button :-)

rventura:
Agreed!

Thank you,
Rodrigo

> On 30/01/2017, at 10:36, Loy <notifications@github.com> wrote:
> 
> Nope, if you agree, I'll press the button :-)
> 
> —
> You are receiving this because your review was requested.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

Rodrigo Ventura
Institute for Systems and Robotics
Instituto Superior Técnico
Lisbon, Portugal
http://users.isr.tecnico.ulisboa.pt/~yoda

Issue #231, opened 2017-02-09
[Help me carry] Standard bag and box


kyordhel:
The rulebook states that

> The operator gives the robot a command to carry something, e.g. a box or a bag.  E.g. "Carry this box"

Box and bag are not listed within Section **3.2.5 Objects**, therefore teams don't know what to expect (nor do I). In some discussion I read that for Pepper a lightweight paper gift bag with rigid handle (See figure below) might be used, since Pepper's arms are not strong, however, I find it difficult for any robot to locate that handle unless is wrapped with a bright and contrasting color.

Whatever solution we provide, the objects should be described in both, the test and the Objects section.

Figure 1: Lightweight paper gift bag
![birthday_paper_gift_bags](https://cloud.githubusercontent.com/assets/8583949/22788426/1643adbe-eee0-11e6-9d8a-6e086b597df5.gif)

dirkholz:
while i like the intention of this test, i doubt that you will see "a nice
solution" but instead hacky stuff that is neither impressive nor coming the
intended capability any closer ... my two cents.

For now I'd say go ahead, but i think we really need to change the overall
concept of all tests to be a "solvable problem" but requiring a real
solution rather than a hack

LoyVanBeek:
That sort of paper bag is more or less what I had in mind indeed. 
For this challenge, I'm less concerned with the manipulation and I'd be OK with giving the robot the bag when it is sticking out it's arms zombie-style without paying attention. 
Same for a box. 

Picking it up is a nice extra, but not core to the challenge.

kyordhel:
Let's be congruent then. 

**Manipulation and Object Detection/Recognition (120pts)**

      60 Bag handover (3 x 20)
      15 Place object (3 x 5)
      30 Moving moveable object
      10 3D Object recognition
    +  5 Small object recognition
    ----------------------------------------------------------------------
     120 Total

**Navigation (90pts)**

      30 Follow Operator (Reach the car 3 × 10)
      30 Online mapping (Reach inside the arena again 3 × 10)
    + 30 Indoor navigation (Reach destination 3 x 10)
    ----------------------------------------------------------------------
      90 Total

**Human-Robot Interaction (30pts)**

      15 Fetch destination (Speech Recognition 3 × 5)
      10 Move aside from person (Person Recognition)
    +  5 Ask person step aside (Person Recognition)
    ----------------------------------------------------------------------
      30 Total

In summary we have:

| Ability                  | Points  | Perc. |
[...]

PR #234, opened 2017-02-13
2017/Help Me Carry

kyordhel:
Help me carry rework as in #231

- Added guide phase
- Special tasks for each league
  - SSPL focused more in HRI
    - Find person in large room
    - Harder people following
    - Harder people guiding
    - Smart obstacle mandatory
  - DSPL/OPL focused in manipulation
    - Bag pickup (no box)
    - Door opening mandatory
    - Moveable object mandaroty
  - Test is no longer stoped if robot pushes (not crashes) small object.
- **Scoring**
  - Up to 200 pts
  - DSPL/OPL and SSPL grant the same points on test completion.
  - Balanced scoring based on the difficulty of achieving the goal.

### Remark
The _smart_ object should be stepping under the door frame or in a corridor  to make this make sense.


tests/HelpMeCarry.tex
```diff
@@ -1,77 +1,137 @@
...
+  \item \textbf{Start:} The robot starts at a designated starting position in the arena, and waits for the \textit{professional} operator. The operator steps in front of the robot and tells it to follow (e.g. by saying \quotes{follow me}). The team is \emph{not} allowed to instruct the operator.
+
+  \item \textbf{Memorizing the operator:} The robot has to memorize the operator. During this phase, the robot may instruct the operator to follow a certain setup procedure.
+
+  \item \textbf{Following the operator:} When the robot signals that it is ready to start, the operator starts walking --in a natural way-- towards the car. Upon arrival, the operator will indicate the robot when they have reached their destination as instructed by the robot (e.g. by saying \quotes{here is the car} or \quotes{stop following me}).
+
+  \newcounter{enumTemp}
+  \setcounter{enumTemp}{\theenumi}
+  \item {[DSPL \& OPL]} \textbf{Bring the groceries in} \\
+  The robot is asked to deliver a bag with groceries to a specific location (e.g. \quotes{Take this bag to the kitchen table}).
```
furushchev: @kyordhel Does it mean that a robot already has a map for entire building and just does not know the car location in the map? Or we must find kitchen table by exploring environment?

tests/HelpMeCarry.tex
```diff
@@ -1,77 +1,137 @@
...
+    \end{itemize}
+
+    \item \textbf{Find a person:} After reaching the designated room, the robot needs to find a person (there is only one person in the room, the name is meaningless).
+  \end{enumerate}
+
+  \item \textbf{Memorizing the \emph{new} operator:} The robot has to memorize the operator that will help. During this phase, the robot may instruct the operator to follow a certain setup procedure.
+
+  \item \textbf{Guiding the operator:} When the robot signals that it is ready to start guiding, the robot guiding the operator to the car. The robot must clearly announce when the destination.
+  \begin{itemize}[leftmargin=3cm]
+    \item[DSPL \& OPL] \textbf{Closed door:} Along it's path to the car, the robot will find a closed door (most likely the entrance to the house) that will need to open to reach the destination.
```
furushchev: @kyordhel Just to clarify, a robot moves the same path twice from home to the car, right? If so, does this mean that at first the door is open and at the second time the it is closed and need to be open?

tests/HelpMeCarry.tex
LoyVanBeek: During the setup days, the teams can map the entire arena, including the kitchen. 

Only the car must is taught to the robot by guiding it there (the robot follow the human to the car)

tests/HelpMeCarry.tex
LoyVanBeek: Yes.

tests/HelpMeCarry.tex
furushchev: @LoyVanBeek OK. Thank you for clarifying! :+1:

tests/HelpMeCarry.tex
furushchev: @LoyVanBeek OK! Thanks!

tests/HelpMeCarry.tex
kyordhel: I'm answering because I was mentioned.

@furushchev as @LoyVanBeek said, your robot knows the arena but nothing else. This part of the test is to prepare robots for Restaurant (safe navigation in unknown environments). You may map the entire venue if you wish, but such workaround goes against the goal of the test.

PR #239, opened 2017-02-24
Added Tour Guide test (SSPL only)

kyordhel:
- **No manipulation**
- Focus:
  - **Mainly HRI**
  - Outdoor safe navigation
  - People detection
  - Gesture recognition
  - Unconstrained NLP
- Up to 390 points (same as Set a Table... test).

### DIFF
--- a/tests/TourGuide.tex
+++ b/tests/TourGuide.tex
@@ -16,7 +16,7 @@ This test focuses in safe outdoor navigation, people detection, gesture recognit
 \subsection{Task}
 
 \begin{enumerate}
-	\item \textbf{Start:} The robot starts at a designated starting position, and waits for referee to give the start signal. When the referees start the time, the team is allowed to (briefly) provide some remarks about the robot's operation. After the instruction, the referee gives the start signal.\\
+	\item \textbf{Start:} The robot waits at a designated starting position for the referee to give the start signal. When the referees start the time, the team is allowed to (briefly) provide some remarks about the robot's operation. After the instruction, the referee gives the start signal to the robot.\\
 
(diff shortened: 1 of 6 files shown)


tests/TourGuide.tex
```diff
@@ -0,0 +1,85 @@
...
+
+	\item \textbf{Other people:} There are no restrictions on other people walking by or standing around throughout the complete task.
+
+	\item \textbf{In Parallel:} This test can run in parallel, with several teams tested simultaneously.
+\end{itemize}
+
+\subsection{Task}
+
+\begin{enumerate}
+	\item \textbf{Start:} The robot starts at a designated starting position, and waits for referee to give the start signal. When the referees start the time, the team is allowed to (briefly) provide some remarks about the robot's operation. After the instruction, the referee gives the start signal.\\
```
LoyVanBeek: Seems a bit contradictory: first a start signal, then some explanation and then another start signal. Perhaps the referee can offer the team  to express some last words before starting the robot :-)

LoyVanBeek:
Nice test! The score is much dependent on the willingness of the public, which may be reduced after a number of robots, but I think that is sufficiently solved by volunteers.

tests/TourGuide.tex
kyordhel: Indeed, I should say:

> The robot waits at a designated starting position for the referee to give the start signal...

This is meant to prevent delays in explanations. The test time is running so the longer (and less intuitive) the instructions are given to the referee, the less time the robot has to accomplish the task. Our ideal is that robot has no issue if a human is following it.

kyordhel:
@LoyVanBeek Thank You!

Indeed, it depends on the amount of people in the venue and their willingness to participate. The test is designed for using volunteers (1~3) for the worst case of an empty venue. However, I preferred to write down the case of a super-full venue, for teams cannot complain about leading with real people.

Issue #246, opened 2017-03-22
Placement of an object - Storing groceries


nohernan:
Hello there,

reading the score sheet for the Storing Groceries task, I noticed that the robot can put an object anywhere in the cupboard and more points are given when the placement is near objects of the same class. I think 
this is part of the task description and should be clearly stated in section 5.1.4 Task. Can you add this note to the task description, please? Or, probably, you don't intend a placement anywhere in the cupboard?

Regards,
Noe - Golem Team

LoyVanBeek:
The idea is indeed to place e.g. apples and apples together in the cupboard. 

Unfortunately, I don't have much time the coming days to put this in. Could you perhaps make a pull request with this fix?

nohernan:
Hello Loy,

I can do the pull request. First, clarify some questions, please. Does the placement of an object in the cupboard have to be exclusively near objects of the same class? I don't quite agree with this idea, since the objects in the cupboard might be in a shelf that is out of reach for the robot, I think the placement should be anywhere in the cupboard and, preferably, near objects of the same class. Of course, those placements near objects of the same class get more points. Do you want me to  add this thought to the rulebook and make the pull request?

Cheers

LoyVanBeek:
Hi Noe,

The idea of the challenge is to also demonstrate both recognition and grasping at the same time, by performing the useful task of storing groceries. 
That is done by placing two objects of the same class together. You can place it somewhere else, but you'll get less points (this is already in the scoresheet). 
Of course, the corresponding item can be out of range of the robot, but that should be an incentive to build robots with a larger reach. Robots with a larger reach are better suited for this task and may get more points if they have their software in order as well, thus are a well integrated system.

PR #247, opened 2017-03-24
Storing Groceries reworked to make it more clear

kyordhel:
Based on #246 

Still I don't like the following.
- 10 Points for grasping + 5 for placing anywhere times five is almost 75pts for nothing. I would consider granting points only for a successful move. 
- There is no difference between known, alike, and unknown objects which are incrementally harder to categorize.
- No points for starting new categories. For instance the robot find an unknown (square pills box) and in the Cupboard there is only an unknown round one. The robot decides there is none of the same and places the box apart because is square. That's another good strategy.

### DIFF
--- a/tests/StoringGroceries.tex
+++ b/tests/StoringGroceries.tex
@@ -1,11 +1,5 @@
 \section{Storing Groceries [DSPL \& OPL]}
-
-The robot must help put the newly bought groceries in the right place.
-The owner of the robot will put the items on a table and the robot helps the owner by moving the groceries from a table to the right place in the cupboard.
-What the right place for an item is defined by the objects already in the cupboard: objects of the same type must be placed together.
-For example, the new pack of cookies from the table must be placed by the almost empty pack cookies in the cupboard.
-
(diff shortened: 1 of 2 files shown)


tests/StoringGroceries.tex
```diff
@@ -1,11 +1,5 @@
 \section{Storing Groceries [DSPL \& OPL]}
-
-The robot must help put the newly bought groceries in the right place.
-The owner of the robot will put the items on a table and the robot helps the owner by moving the groceries from a table to the right place in the cupboard.
-What the right place for an item is defined by the objects already in the cupboard: objects of the same type must be placed together.
-For example, the new pack of cookies from the table must be placed by the almost empty pack cookies in the cupboard.
-
-In the cupboard and on the table, there will be both known, alike and unknown objects. 
+The robot helps by storing newly bought groceries in the cupboard next to the objects of the same kind that are already there; for instance by placing fresh apples where the other apples (or fruits) are.
```
LoyVanBeek: The "(or fruits)" makes less clear again.

Issue #249, opened 2017-05-03
About handover in Help-me-carry


arturo8602:
I have a question about Help-me-carry. In section 5.4.4.4.1 the line “The operator hands a bag to the robot to carry” I see at least three posible scenarios:
a) The easiest.- the robot only puts its hand in a position to receive passively the bag (in this case it is not necessary to see the bag). The person places the bag in the robot's hand.
b) The hardest.- the robot searches the bag and estimates its position to grasp it. The person doesn't colaborate with the robot. 
c) An intermediate option.- Both the robot and the person colaborate. The robot actively searches the bag, but the person also delivers the bag moving his hand with the intention of placing the bag in the robot's hand.
I suposse it will be a scenario similar to b or c considering that this test is focused in object detection and manipulation, but I think it should be specified clearly which kind of scenario is expected in this test adding a section to define the term 'handover' and when it will be considered as correct. 

In section 5.4.6.6 the term handover is wrong: “Bag pick-up: If the robot can't handover the bag from the operator”, because the robot doesn't deliver the bag to the person. The robot is the one who receives the bag.

Arturo Rodriguez, Team Golem

kyordhel:
@arturo8602 let's assume for a second you are not interacting with the robot but with another human being, for instance... me. What would you expect me to do? Of course you wouldn't hang the bag in my hand passively waiting, do you? Often there is some sort of collaboration, human A extends its hand, B grasps the object, and A releases the object when feels B pulls or somehow knoews the object is secured.

Following criteria will be used:

- Robot puts up hand begging for a bag and closes after some time: 0 points
- Robot puts up hand and closes it after bag-insertion: 1 point maybe?
- Robot picks up bag from ground: 5 points
- Robot takes it from my hand, human only puts hand forward slightly as it would do with another human: 20 points

LoyVanBeek:
This links to https://github.com/RoboCupAtHome/RuleBook/blob/master/tests/HelpMeCarry.tex#L39 and https://github.com/RoboCupAtHome/RuleBook/blob/master/scoresheets/HelpMeCarry.tex#L11 

This could indeed be clearer. I've also identified another way to do this. A clarification is in https://github.com/RoboCupAtHome/RuleBook/pull/250

PR #250, opened 2017-05-03
Clarify the bag pickup (#249)

LoyVanBeek:
Merge only after German Open

### DIFF
--- a/general_rules/Scenario.tex
+++ b/general_rules/Scenario.tex
@@ -127,9 +127,18 @@ The following general rules for objects apply:
 \paragraph*{Important note:} It is not allowed to modify any of the objects provided for training. Also, teams are not allowed to keep more than 5 the objects provided for training at a time nor retaining it for more than one hour.
 
 \subsubsection{Containers}
-The TC will provide at least two containers (a transport container such as a tray and a pouring container such as a bowl) which will be available for training during the setup days.
+The TC will provide at three different types of containers to be used in the tests.
 
-There are no restrictions on a container size, appearance or weight; however, it can be expected that the selected containers be lightweight, with handles, and easily manipulable by a human using both hands.
+\begin{itemize}
+	\item \textbf{Pouring containers:} Such as a bowls, glasses, or other objects in which liquids and grains can be poured.
+
+	\item \textbf{Storage containers:} Such as bags or boxes in which objects can be stored or retrieved.\\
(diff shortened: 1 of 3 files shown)


kyordhel:
Will review this after German Open for security reasons.

HideakiNagano21:
handover

I've made this figure in order to make clear the scenario for Japan Open which is going on today.
In Japan Open, human did not extend his hand actively but kept the posture.
One team succeeded scenario-A' in the above figure.
Several teams selected scenario-C on purpose.

LoyVanBeek:
How difficult B is here really depends on the bag: 

- a paper bag is quite still and keeps the handle upward and clear --> relatively simple
- A plastic or cloth bag is soft and the handle will 'crumble'. --> very hard

Currently at the GermanOpen, we also have paper bags for this stask so I think the rulebook should fix this in the rulebook for now. We might make this more difficult next year.

general_rules/Scenario.tex
```diff
@@ -127,9 +127,18 @@ \subsection{Objects}
 \paragraph*{Important note:} It is not allowed to modify any of the objects provided for training. Also, teams are not allowed to keep more than 5 the objects provided for training at a time nor retaining it for more than one hour.
 
 \subsubsection{Containers}
-The TC will provide at least two containers (a transport container such as a tray and a pouring container such as a bowl) which will be available for training during the setup days.
+The TC will provide at three different types of containers to be used in the tests.
 
-There are no restrictions on a container size, appearance or weight; however, it can be expected that the selected containers be lightweight, with handles, and easily manipulable by a human using both hands.
+\begin{itemize}
```
kyordhel: For *Help-me-carry* there is no other container unless we fully incorporate the box (I will remove the reference in a PR later today). However, other tests (¿?) use other types of containers, such as:
- Bag
- Basket (you used one in German Open)
- Bowl
- Tray
- Plate

Bowl and plate are containers in the very moment you can pour milk/cereal and bread on them, respectively. Basket is useful for placing bread in it (e.g. serving the table), but it may also be used along with the tray to transport 2+ objects in a test like *Set the table...* and *Restaurant*. A smart robot could place all cutlery and dishes in the basket, transport the basket to the table, and do the setting with one single transport instead of 3+.

LoyVanBeek: Yeah, I didn't notice the change was in the general rules and not specific for this test and deleted my comment on this. 
The changes are fine.

Issue #252, opened 2017-05-03
Set a table and clean it up


arturo8602:
On page 68 section 6.2.1.Example, there is a contradiction about the location of the items, first says:
“Several items are placed on the table: a plate, cutlery, a cup, napkins, a basket with bread, a ...”. But after says: “The basket with bread, a napkin, and the cereal box are on the kitchen counter...”, and then says “the plate, the cutlery, and the cup are inside a closet”. 
I suppose the objective of the first line I mentioned was to set the objective of the robot, but the redaction is not clear. Maybe “Several items will be placed on the table by the robot: a plate, cutlery, a cup, napkins, a basket with bread, a ...” could be more appropriate. 

I also found a contradiction between the line 6.2.6.4 where it is mentioned that an order is composed by 6 objects (3 easy and 3 difficult) and the score sheet in the section grasping objects where there is assigned 6x10 for easy-to-grasp objects and 6x15 to hard-to-grasp objects.   

Some typos:
1) Page 68 section 6.2.1. Example:
The robot starts moving these items, one by one, from their original position to the table.
It should be:
The robot starts moving these items, one by one, from their original location to the table.

2) Page 69 section 6.2.2. Goal sentence 3:
locations and cleanup dirt ansd spills from the table.
It should be:
locations and cleanup dirt and spills from the table.

3) Page 69 section 6.2.5. Task 3. Searching for objects:
... on the kitchen counter on inside the closet.
should be:
... on the kitchen counter or inside the closet.

4) Page 69 section 6.2.6. Additional rules and remarks:
3. Collisions: Slightly touching the table. Driving...
should be:
3. Collisions: Slightly touching the table is allowed. Driving...
5)  Page 69 section 6.2.6. Additional rules and remarks:
...stop the robot (Section Section 3.8.1)
should be:
...stop the robot (Section 3.8.1)

6)  Page 70. 6.2.9. OC instructions:
... the objects, as specified in item ?? of section 6.2.6
I don't know how it should be, but the answer marks should be replaced by something

7) Page 71
For each successful grasp of an hard-
It should be:
For each successful grasp of a hard-
(This typo is repeated three times in the same page). 

Arturo Rodríguez and Ivette Vélez, Team Golem

kyordhel:
@arturo8602 thanks for the typos and inconsistencies between the test and the scoresheet.

[...]

PR #262, opened 2017-05-10
Set a Table and Tidy Up reworked

kyordhel:
- Addressed #252
- Removed long test in favor of a boring ans structured schema
- Test structure was preserved
- Added images on how table should look like.
- Several issues to be solved before merging
  - [ ] Test requires cleaning before and after setting the table
    - This is redundant
    - Also a waste of time
    - Too complex for the first time. Instead robot should just set the table and clean afterwards.
  - [ ] It is not specified what is the (default) table setting
  - [ ] When the option of serving a meal is used, table setting must depend on it. E.g.
    - Bowl and spoon for cereal
    - Dish, fork, and glass for toasts and juice
    - Knife, spoon, fork, cup, dish, and glass for English Breakfast
    - Dish and mug for croissant and coffee
    - Napkin and carpet always present
  - [ ] After the meal, the robot should take dirty utensils to the sink or dishwasher
    - In a first run, all to the sink (OC reorganizes)
    - Later on, robot should take dirty appliances to the sink and clean ones back to the cupboard.


tests/SetATableTidyUp.tex
```diff
@@ -1,57 +1,74 @@
...
+\begin{itemize}
+	\item[Part I:] Setting up the table
+	\begin{enumerate}
+		\item \textbf{Fetching command:} The operator requests the robot to set up the table. Optionally, the robot may ask the operator \textit{what has to be served}. The operator replies with the randomly selected (and undisclosed) option.
 
-In phase~II, the operator requests the robot to clean up the table. This includes both returning the placed objects to their original location, and cleaning up dirt and spills using a cleaning cloth. This requires the robot to memorize the location the objects were originally recognized.
+		% I find this unnecessary, the table must be clear already or this test will take forever.
+		\item \textbf{Clearing the table:} The robot must analyze the table and remove all unnecessary objects.
```
LoyVanBeek: This seems redundant with Phase II and was not in the original description. I think we can just leave this bit out

LoyVanBeek:
As for the meal options and what they entail for the table setting, this depends on the objects available and can thus only be specified at the event. We can provide some examples though

LoyVanBeek:
Further, having a robot separate out clean and dirty dishes is something I'd like in my home but is maybe something for next year. We can make it optional now, but nobody ever does optionals.

tests/SetATableTidyUp.tex
kyordhel: Agreed, I'm fixing that one right now.

kyordhel:
### Default table setting
I think we can get *cereals* and sets of *fork, spoon, and knife* all around the world without any problem. If the option is cereal, a bowl and spoon is enough. Appliances are easy to remove but hard to add.

### Separating clean and dirty
No way for this year, that's for the years to come, but utensils should be taken to the sink afterwards assuming all are dirty after use. No fork has ever complained for being washed when clean.

PR #264, opened 2017-05-10
Simplifying EEGPSR

kyordhel:
To be merged ONLY if you approve these changes

- Reduced number of categories to 3
	- Category I: Three at once (former Cat VIII)
	- Category II: People related tasks (former Cat III and IV)
	- Category III: Object related tasks (former Cat I and II)
- Complete, incomplete, and erroneous commands are mandatory but given in random fashion
- Evaluation and scoring
	- Score is by executed command
	- Evaluated aspects
	- Command retrieval (3)
	- Successfully executed explicit command
	- Successfully executed command with missing information
		- Merged former Category V with chosen new category
	- Successfully executed command with erroneous information
		- Merged former Categories VI and VII with chosen new category
	- Max score clamped to 300
- Updated test
- Updated appendix
- Updated score sheet

### DIFF
--- a/scoresheets/EEGPSR.tex
+++ b/scoresheets/EEGPSR.tex
@@ -1,51 +1,32 @@
-The maximum time for this test is 40 minutes.\\
-{\scriptsize \textbf{Note 1:} Related abilities with similar score are packed together for brevity. Rreferees must mark them and score accordingly.}\\
-{\scriptsize \textbf{Note 2:} Abilities marked with * score \textit{at most} the suggested points based on the difficulty and performance (following TC criteria).}
-
-{\footnotesize
-
+The maximum time for this test is 10 minutes.
+%
+% MAURICIO 2017
+% Compact Scoresheet
+%
 \begin{scorelist}
-	\scoreheading{Command retrieval \& Categories}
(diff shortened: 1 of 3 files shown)


balkce:
I'm actually for it. They're some issues here and there, but the idea definitely simplifies the scoring and idea behind the test.

However, I'm not comfortable doing this big of a change to a test (any test) this close to the competition (two months).

LoyVanBeek:
I have to agree it is too short before the World Cup to do this. 
Further, we split up the commands to make the grammar easier as well, which we would undo with this. 

As a middle road, we could simplify the score: one score for succeeding a Cat. 1 assignment, a different for Cat 2 etc. But: the downside is that if a robot gets 99% of a command correct, that would be 0 points and very demotivating. 
But: we specifically fine-grained this scoresheet. Maybe we just have to bite the bullet and run the test mostly as it is, with complex scoring. With proper preparation and enough people to help run it, I think we can manage. 

This means: commands generated, all relevant objects ready, 1 person setting up the arena while another handles the robots and 1 team member accompanying it. Outside the arena is someone to set up robots for the next attempt.

PR #266, opened 2017-05-10
Feature/spr clarify

LoyVanBeek:


LoyVanBeek:
For questions about persons, objects, the arena, eg. "Is the smiling person a boy or a girl", the robot may say "The smiling person is a boy" although there was only a smiling dude. This means it understood what I asked _but_ that it did not extract the correct answer.
Since we are testing both speech understanding and person recognition, I think it makes sense to separate these scores: points for ASR/NLP and points for person recognition.

kyordhel:
You are flipping right!!!!

In fact I think is a bug that the robot gets score for counting girls+boys as correct but not for the size of the crowd regardless gender (sum might be right). Yet, I think this is ok unless we get androgynous people around.

Issue #278, opened 2017-05-17
Enhance description of External Computing requirements


kyordhel:
It has been made of my knowledge that the SPL community is concerned about software running on External Computing

> Many people are wondering whether it is possible to put some modules they developed on the external PC. [...] they tend to imagine "publicly available" only means that the se[r]vice should be available anytime for everybody through internet like cloud service. So I would like you to add the explanation of your conditions, especially A).  Actually not only new team members but also experienced members in Japan were not sure whether the rulebook includes the conditions or not.

formerly I explained that software must meet either A or B of the following requirements
A) Software is be open source (BSD/GPL/etc)
B) Team provides product detailed information (e.g. Vendor, patent number, etc.) and interfacing must be published for scientific use

PR #279, opened 2017-05-17
External Computing explicitly allows own software

kyordhel:
- Addressed #278
- Cloud computing and other propietary solutions with closed code are allowed
  - Details on how to get the product must be provided.
- Open source software is allowed
  - The source must be specified.

### DIFF
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -4,9 +4,12 @@ Robots are allowed to use some form of external computing, for example in the fo
 \begin{enumerate}
 	\item \textbf{Definition:} Computing resources that are not physical part of the robot are \iterm{external computing resources}. 
 	\item \textbf{Inspection:} In general, external computers are not allowed unless explained to and allowed by the \iaterm{Technical Committee}{TC}.
-	  A team must announce to the TC at least 1 month in advance the external computing resources they want to use, for what purpose and how to reach the resources.
-	  E.g. specify the URL or IP-address. 
-	  This must be specified in the team description paper. 
+	  A team must announce to the TC at least 1 month in advance the external computing resources they want to use, for what purpose and how to reach the resources (e.g. specify the URL or IP address and port). Inspected software must meet the following \textbf{requirements:}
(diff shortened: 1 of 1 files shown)


LoyVanBeek:
I'm not sure if "Open source is not mandatory" is compatible with the goals and spirit of RoboCup actually. But then again, we even have closed-source robot platforms in the SPLs. 
If the RoboCup federation has no strict requirement on (team) software being open source, I'm fine with this change. 

I'm also OK with other teams being able to deploy their very own server with sources provided by another team, which could simplify things s bit

kyordhel:
@LoyVanBeek | There is no way we can enforce the use of open source. Many successful solutions used are not open source and yet used in the scientific community (at least for reference).
- Cloud Services such as Google's, Wolfram's Amazon's, and Microsoft's to name some, have public API's (e.g. Speech Recognition) but software is not open source.
- Embedded Solutions. Microsoft SAPI, Loquendo, Nuance's Speech Recognizer, and Verilook Face Recognition are some of the proprietary solutions used in the past by some teams, and maybe still used.
- Environments. Matlab/Simulink is the one that comes to my mind. None really knoww how matlab works, and I am not against some crazy dude have running that thing on a robot despite is slow as snail.
- Databases. Teams may choose Oracle or SQL Server to store their knowledge bases (I've no idea why choose a monster instead of a lightweight MySQL or SQLite, yet is possible).
- Operating Systems. Microsoft Windows and iOS are not open sourced, yet some teams can use a Windows Machine or iPad on the robot.
- Sponsored: Maybe the condition for a team to get sponsored is to use the not-open-sourced proprietary solution developed by the sponsor. I can't find a reason to prevent such teams from participating.

But teams are more concerned in running their own algorithms on a big computer because the one in Pepper/HSR is something close to an old 80's VCR. I would say: go and try! Nowadays many teenagers freak out and turn unable to do anything if they get disconnected from the internet (well, their phones, but there is slight difference), so in a few years will be an acceptable behavior in a robot. Who knows? Maybe Amigo will upload some selfies to instagram of the pretty table  it just set up.

Issue #283, opened 2017-05-19
GPSR: Score Sheet Clarification


johaq:
**GPSR**
The score sheet suggests that completing the third given command gives more points than completing the first. I think the intention was of course to award robots to complete multiple tasks so I think it should state clearly that the additional points for the later commands are only awarded if the previous command was completed successfully. Thats how it was scored at the German Open anyway but I still think this should be clearer in the rulebook.

kyordhel:
#284 attempts to fix this issue. Hope is clear enough.

PR #284, opened 2017-05-19
GPSR with incremental scoring

kyordhel:
- Addressed #283
- Added additional rule explicitly explaining scoring is incremental.
- Is required to solve a first command to get points for the second
- Is required to solve a first and second command to get points for the third.
- The referee determines whether the command was successfully solved or not.

### DIFF
--- a/tests/GPSR.tex
+++ b/tests/GPSR.tex
@@ -64,7 +64,8 @@ This test particularly focuses on the following aspects:
 	\end{itemize}
 
 	\item \textbf{Retrieving the command:} The robot must show it has understood the given command by repeating the command (i.e.~stating all the required information to accomplish the task.).
-	\item \textbf{Incremental scoring:} Scoring depends on the category chosen by the team leader and the previous successfully accomplished command. Thereby, scoring for a second and third command is possible only if the robot has already solved (not understood) a first and second command respectively. Referees determine when a command is considered accomplished or not.
+
+	\item \textbf{Incremental scoring:} Scoring depends on the category chosen by the team leader and the previous successfully accomplished command. Thereby, scoring for a second and third command depends on how well the robot solved (not understood) a first and second command respectively. Referees determine how well the command was accomplished and its impact on the incremental scoring of subsequent commands.
 \end{enumerate}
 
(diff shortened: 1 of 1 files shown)


balkce:
Seems simple enough. However, this will put some weight over the judges when they need to establish if the robot "solved" the first part of the command and throwing out the points for the rest of the test if they considered it didn't. That's a big punishment.

I get the objective of this merge: if the robot was not able to pick up the object, it shouldn't get the points for delivering it. But, I would like to remind you that this is a first stage test, so I think it's fine to give points here for delivering the object without picking it up in the first part since: 1) it shows navigation skills; 2) if the second part was "deliver it to Steve", points should be awarded if the robot found Steve even if it didn't deliver the object; and 3) the robot could also excuse himself to the user that it wasn't able to pick it up in the second part of the test, which seems appropriate and worthy of some points.

I propose to change the text to:

> Scoring depends on the category chosen by the team leader and the previous successfully accomplished command. Thereby, the scoring for a second and third command depends on how well the robot solved (not understood) a first and second command respectively. Referees determine how well the command was accomplished and its impact on the incremental scoring of subsequent commands.

Basically instead of going binary in the first command (not accomplished first command, no points for the rest of the test), go fuzzy in the second and third (not accomplished first command, half points for the rest of the test, or something like that). And I'm leaving it vague and "up to the referees", so as to not require a change in the already complex score sheet.

Finally, having said all this, I believe that this binary style could be implemented in the EEGPSR since it is in the second stage where the better teams are competing, so I'm fine being harsh to them.

kyordhel:
Well, the scoring wasn't modified from how it was executed in German Open. The only clue for this rule was in the scoresheet stating: 
- First Command Successfully Solved.
- Second Command Successfully Solved.
- Third Command Successfully Solved.

giving the impression of dependency among the executions. This way, a robot can't just pretend it solves the first two commands (e.g. retrieving the command, navigating, say it cant and going back) and then try to get the big score for the first. When a command was solved? Well, that depends a lot on the referee and on the command! As you mentioned, if the robot drops the object in the middle of the delivery, the tasks was completed at least in the mind of the robot, so it can go for a second one.

Issue #290, opened 2017-05-24
SSPL deviation from the "Standard"


warp1337:
Dear all,

I was just browsing the rule book (real quick) for a **definition** of a SS(!)PL robot. I mean, the term standard inherently defines that no changes to the robot are/should be allowed, like adding or changing hardware, since this then cannot be considered "a standard" anymore. 

However, I have a specific question: let's assume one wants to implement visual servoing. In order to do this one could take a shortcut and place ar-markers (i.e. stickers) on the robots end-effectors. Is this considered a deviation from the standard? @komeisugiura, @LoyVanBeek

LoyVanBeek:
That is an interesting case, I can't truly judge this, @komeisugiura hopefully can.

My preference is towards a no, but not sure yet.

warp1337:
Thxs Loy. We'll see.

komeisugiura:
We should not accept such an AR-marker. It introduces differences in hardware, which is not based on the standard platform's spirit.

Btw, introducing markers on walls or objects were prohibited in @Home long time ago, which I think is written in the rulebook.

warp1337:
Yes, agree. I read that markers _on walls or obejcts_ are forbidden, thanks for the confirmation again. I was just wondering, thus asking 👍 .

However, to make it clear to everyone, maybe we can include something like:

"In the SPL, changes to the robot hardware are not allowed in order to keep the 'standard' an actual standard. This includes changes to its outward appearance as well. Thus, e.g., ar-markers, stickers or any kind of artificial landmarks on the robot are forbidden."

Again, thank you for clarification.

kyordhel:
@warp1337 
I'm confirming @komeisugiura's statement. Using artificial markers is strictly forbidden since 2009 (Section 3.2.3: External devices). If memory fails me not, having artificial marks in the robot would make it fail the inspection, have its participation in a test denied, or test termination followed by team disqualification from that test.

Maybe is a non-written rule (I saw it in previous rulebook but I can't find it anymore), but OpenCV/QR codes and other types of artificial patterns in robots, walls, floors, furniture, people, etc, have been always considered forbidden and as cheating. In contrast, there is no rule against having one red hand an one green hand in your robot as long as it goes with the robot's amazing fashion style, and not as an evident marker.

warp1337:
Thank y'all. Marking this closed.

PR #295, opened 2017-05-24
Explicit rules against artificial marks added.

kyordhel:
- SPL Robots are allowed to wear sponsor marks and stickers
- OPL Robots are advised against prototypr appearance and T-Shirt cover
- Fixed several typos
- Addressed #290

### DIFF
--- a/general_rules/Robots-OPL.tex
+++ b/general_rules/Robots-OPL.tex
@@ -11,6 +11,8 @@ Robots competing in the RoboCup@Home Open Platform League must comply with secur
 	\item \textbf{Transportation:} Team members are responsible for quickly moving the robot out of the arena.	If the robot cannot move by itself (for any reason), the team members must be able to transport the robot away with an easy and fast procedure.
 \end{enumerate}
 
+
+
 \subsubsection{Emergency stop button}
 \label{rule:robots_emergency_button}
 
@@ -20,7 +22,6 @@ Robots competing in the RoboCup@Home Open Platform League must comply with secur
 	\item \textbf{Robot behavior:} When pressing this button, the robot and all parts of it have to stop moving immediately.
 	\item \textbf{Inspection:} The emergency stop button is tested during the \iterm{Robot Inspection} test (see \refsec{sec:robot_inspection}).
 \end{enumerate}
-
 
 
 
@@ -37,7 +38,9 @@ Robots competing in the RoboCup@Home Open Platform League must comply with secur
 
 
 
-\subsubsection{Audio output plug}\label{rule:roobt_audio_out}
+
+\subsubsection{Audio output plug}
+\label{rule:roobt_audio_out}
 
 \begin{enumerate}
 	\item \textbf{Mandatory plug:} Either the robot or some external device connected to it \emph{must} have a \iterm{speaker output plug}. It is used to connect the robot to the sound system so that the audience and the referees can hear and follow the robot's speech output.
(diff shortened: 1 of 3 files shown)


general_rules/Robots-OPL.tex
```diff
@@ -37,7 +38,9 @@ \subsubsection{Start button}
 
 
 
-\subsubsection{Audio output plug}\label{rule:roobt_audio_out}
+
+\subsubsection{Audio output plug}
+\label{rule:roobt_audio_out}
```
LoyVanBeek: Is this correct?

general_rules/Robots-OPL.tex
```diff
@@ -47,4 +50,11 @@ \subsubsection{Audio output plug}\label{rule:roobt_audio_out}
...
 \end{compactitem}
-\end{enumerate}
\ No newline at end of file
+\end{enumerate}
+
+
+
+
+\subsubsection{Appearance}
+\label{rule:robots_appearance}
```
LoyVanBeek: I don't really agree with this rule for OPL robots. AR-markers are not allowed for OPL robots?

general_rules/Robots-OPL.tex
kyordhel: I think so, I just added a new line

general_rules/Robots-OPL.tex
kyordhel: AFAIK no artificial markers of any kind are allowed in robots since... 2009 at least. Plus I must say that I haven't seen them in a while in @Home.

However, it is not my point to forbid something wasn't forbidden, but to clarify the rules. If we are unsure, we should ask @dirkholz, @komeisugiura, @swachsmu, and @iocchi to name some of the EC/Trustees who have been in the league longer than I.

LoyVanBeek: I know marking any else than the robot has indeed been disallowed but markings on the robot itself have been fine, AFAIK.

kyordhel: I remember it wasn't because several teams in 2010~2012 tried to stick OpenCV/AR-markers on the manipulator's terminal effector for closed loop control while grasping and TC required to remove the stickers to pass RIPS (my former team was one of the affected ones in 2010, so we decided to have a blue hand, and a pink hand). And those were pre-kinect times.

Again, I am not against allowing markers on the robot (although in my opinion it looks pretty nasty), I am just trying to promote consistency.

kyordhel: @balkce @komeisugiura @swachsmu @dirkholz your opinions would help here.

balkce: I remember that Pumas (@kyordhel's former team) used AR-markers for closed-controlled manipulation, but I remember it more recently (pretty sure it was Netherlands 2013). I did not know that TC back in those days asked teams to remove them.

In any case, I really don't mind if the stickers are on the robot's body in OPL, since the philosophy in that league is that they can do whatever they want with the body of the robot (as long as they don't modify the arena in any way). However, in both standard platforms, putting a sticker that isn't for promotional/decorative use and obviously for ease of hand tracking, is walking the line of not being appropriate:

CON PUTTING THE STICKER: the philosophy behind the standard platform leagues is that the software solutions that they come up with can be used in any instance of that hardware platform. If they move that solution to another robot that doesn't have the sticker, their software doesn't work anymore.

PRO PUTTING THE STICKER: it's just a bleeping sticker, which is easily installable.

Ugh... I hate these kinds of issues.

Compromise: let them use stickers, but it is up for the TC during the competition if they are within the "gentleman rule" in the rule book.

kyordhel: Ok, Now artificial markers are set for OPL and forbidden for SPL

LoyVanBeek:
I approve, but am not listed as reviewer. @balkce ?

balkce:
Weird. I'm approving the merge, but @LoyVanBeek should really look into this not-being-a-reviewer issue. You're the owner of this git project, for Turing's sake!

LoyVanBeek:
Hmm, I am listed as reviewer but didn't have an approve-button for some reason.

PR #302, opened 2017-05-25
Removed bonus in Open Challenge for multi-league teams

kyordhel:
The rule Team-team-interaction would grant a bonus to one single team participating in two leagues as they are registered as two different teams.

A remark has been added specifying that interleague collaboration is possible (SPL->OPL only), but teams competing in several leagues can't benefit from the bonus.

### DIFF
--- a/tests/OpenChallenge.tex
+++ b/tests/OpenChallenge.tex
@@ -80,7 +80,7 @@ For sake of clarity, please consider the following example: Let be A, B two team
     \item Team B participates in both SSPL and OPL.
     \item Team A and B have qualified into Stage II.
 \end{itemize}
-Then, by aplying Rule \ref{rule:OC-inter-league-collaboration} the following statements can be concluded:
+Then, by aplying the \textit{Inter-league collaboration Rule} (See \refsec{rule:OC-inter-league-collaboration}) the following statements can be concluded:
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


LoyVanBeek:
Reasonable, but I think this is slightly too broad: 
I'm fine with Team A's SSPL robot cooperating with Team B's OPL robot, even when Team B also has a SSPL robot. 
Any reader will understand this already, but this could throw them off and ask questions.

komeisugiura:
Although the original sentences make sense, @LoyVanBeek 's point sounds reasonable.

kyordhel:
I am not sure to be following your point. 
Team A is [S]SPL
Team B is [S]SPL and OPL

The rule should state clearly the following
- B OPL can **not** participate in A SSPL's open challenge.
- B OPL can **not** participate in B SSPL's open challenge.
- A SSPL can participate in B OPL's open challenge. There *is* bonus because A <> B.
- B SSPL can participate in B OPL's open challenge. There is no bonus because B = B.

Unless you disagree with the former, if this is not 100% understandable, the rule needs extra clarification.

LoyVanBeek:
This is 100% what I mean indeed.

kyordhel:
Ok, hope #d6d77e6 solve it.

Issue #258, opened 2017-05-06
Opening the door in the storing groceries task


swachsmu:
I think it is not a good idea to have the complete shelf covered by a door. For the teams it is very risky to open it autonomously because they may loose all the points for placing and recognition of shelf objects. I would suggest to close only one half of the shelf by a door. Thus, the robot can work on the task without opening the door, first, and solve half of the task. Then, the robot can open the door, gain an extra time bonus, and is able to solve the complete task. The robot would NOT be allowed to ask a human to open the door for him.

A similar thing with time bonus could be done to distinguish team which just do object recognition from teams who a doing the grasping and placing. At the first time the robot has only 2min for visually inspecting the objects and to a first grasp. If the grasp was successful the robot would gain a time bonus for completing and doing more grasps (e.g. extra 2min).

kyordhel:
@swachsmu I agree with your suggestion. Even we might remove one of the doors for safety reasons. In my understanding, the _cupboard_ must contain at least one [small] door with space for performing the test even if the door is closed, like the one in the picture below:

![image](https://cloud.githubusercontent.com/assets/8583949/25799014/ad530f0e-33e3-11e7-9a6c-ae983a585cfb.png)

However, this kind of shelf might not be accessible to all teams since many could not be able to reach the upper part. Therefore, I think the RuleBook should depict something more or less like this

![image](https://cloud.githubusercontent.com/assets/8583949/25799434/3a94831a-33e5-11e7-91e2-b564c1f78c20.png)

Or within our budget (we may not place the right door of this IKEA shelf):

![image](https://cloud.githubusercontent.com/assets/8583949/25799400/15fbf59c-33e5-11e7-818b-c31eebe42031.png)

About the 2 min limit for first grasping, I also give my up vote. Although it makes the test slightly harder to control, it also adds an implicit requirement of action planning. In order to get the extra time, teams will be forced to choose the easier-to-grasp object in sight, and not a random one.

balkce:
@kyordhel pretty much explained all I have to say in the matter.

The last IKEA shelf without one of the doors in place seems ideal. In any case, my guess is that there won't be a lot of teams attempting to open the door.

The only issue I have is that the time addition will make it harder to evaluate, but as long we have enough judges and one of them taking care of only of the timing, I see no problem.

LoyVanBeek:
[...]

PR #306, opened 2017-05-28
Fix/#258 door opening

LoyVanBeek:
- Only half of shelf covered
- Limit to for non-manipulating robots, add a bit of time for those who can. 

Address https://github.com/RoboCupAtHome/RuleBook/issues/258


tests/StoringGroceries.tex
```diff
@@ -16,6 +16,7 @@ \subsection{Focus}
 		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.30m and 1.80m from the ground and contains several objects (See \ref{rule:scenario_objects}).
 		\begin{itemize}
 		 	\item \textbf{Door:} The cupboard has a single door, which is closed initially.
+		 	This door covers half of the objects, e.g. the left/right half of the closet, as indicated by the hatched area in the image.
```
kyordhel: I would rather say ```This door encloses some of the objects, covering up to one half of the cupboard (e.g. the left/right...```

This is because we don't know beforehand how he cupboard will look like.

scoresheets/StoringGroceries.tex
```diff
@@ -1,8 +1,11 @@
 
-The maximum time for this test is 1+3 minutes.
+The maximum time for this test is 1+2+2 minutes.
 The first minute is for the robot to open the Cupboard's door on its own.
 If the robot is not able to open the door within that minute, it will be opened by the referee. 
 In case the robot opens the door within the minute, the robot has a small time advantage. 
+Then, the robot has 2 minutes to recognize objects and manipulate items.
+If an item is manipulated within that time, 2 more minutes are added to the test time.
```
kyordhel: If robot fails to move one object within the first 3 minutes, the test ends. Nice!
However this somehow contradicts the rule for inoperative robots (30sec), so a robot should take no more than 60 seconds to place the first object in the shelf counting from the door opening. Lets say:
1+1+3 = 5

LoyVanBeek: I'm fine with the robot looking and driving around (!= inoperative) to get a better view.

kyordhel: You are right! I didn't thought about that. Two minutes is ok then

tests/StoringGroceries.tex
```diff
@@ -16,6 +16,7 @@ \subsection{Focus}
 		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.30m and 1.80m from the ground and contains several objects (See \ref{rule:scenario_objects}).
 		\begin{itemize}
 		 	\item \textbf{Door:} The cupboard has a single door, which is closed initially.
+		 	This door encloses some of the objects, covering up to one half of the cupboard (e.g. the left/right...), as indicated by the hatched area in the image.
```
kyordhel: You took my proposal way too literal. The ... were because I was too lazy to copy the whole sentence!!!

LoyVanBeek: :-)

Issue #276, opened 2017-05-17
Grammars of questions for Speech and Person Recognition


hirokiyokoyama:
It seems that the repository of GPSR command generator
[https://github.com/kyordhel/GPSRCmdGen](https://github.com/kyordhel/GPSRCmdGen)
has also one for Speech and Person Recognition (SPR) questions,
which is referred neither in the section 5.3 (SPR) nor in Appendix A (Questions for SPR).
Is it going to be used in SPR?
If so, I would like to propose that it should be clarified in the rule book as well as [EE]GPSR.

LoyVanBeek:
That would indeed be an improvement. 

I'm can't fix this at this moment but a pull request fixing this is also very welcome :-)

kyordhel:
Indeed I added a generator for the Speech and Person Recognition test as you mention; but as it was unofficial, I never added the reference to the rulebook. Is important to remark that the list of questions may (and probably will) change during setup days.

*I been not also  too much of free now*, but since Loy approves, a PR will be handy.

PR #307, opened 2017-05-28
Reference GPSRCmdGen also in SPR

LoyVanBeek:
Fix https://github.com/RoboCupAtHome/RuleBook/issues/276


kyordhel:
I really need to update that generator soon.

Issue #310, opened 2017-05-30
Finalization of Speech and Person Recognition test


mtlazaro:
Hi,

we have some question about the finalization of the task "Speech and Person Recognition".
The rulebook says "The robot leaves the room through the designated door", however no points are awarded for this navigation task. 
Furthermore, it says "This test may also be held outside the arena" so in that case there will be no doors.
How to finalize this task is incoherent in our opinion, especially since the focus is speech understanding and not navigation.

LoyVanBeek:
Navigation is such a basic task that no points are awarded in this task for that. No penalty or whatever is given if the robot is not able to do it, it is mostly an inconvenience to the next team coming up.

kyordhel:
Is not incongruent, but maybe nor properly explained. The point is there because, in case that the test is held in the arena (most likely), the robot **must leave autonomously**. It gives a terrible impression to the audience to see robots being dragged or pushed outside, so this requirement is only there to make the test more appealing.

However, if the test is not held in the arena (something possible but undesirable), there is no audience, so no need for appealing behavior.

PR #311, opened 2017-05-30
Remove leaving the arena after SPR.

LoyVanBeek:
It's fine if the operators move the robot away.

Addresses https://github.com/RoboCupAtHome/RuleBook/issues/310

### DIFF
--- a/tests/SPR.tex
+++ b/tests/SPR.tex
@@ -32,7 +32,7 @@ This test focuses on human detection, sound localization, speech recognition, an
     This process is repeated with 5 (possibly) different people. 
     The game will end when the 5th question has been made, following the same distribution of questions as in the riddle game. The robot must answer the question without asking confirmation. Questions may be repeated once.
     
-    \item \textbf{Leave} Th robot must leave the arena/test area to make place for the next robot to be tested. 
+    \item \textbf{Leave} The robot must leave the arena/test area after all questions have been asked or when instructed to do so.
 \end{itemize}
 
 \subsection{Additional rules and remarks}
@@ -92,7 +92,7 @@ The referee needs to
 \begin{itemize}
     \item Announce the placement of the robots
     \item choose the volunteers for the second part of the test, and clearly explain the procedure to them.
-    \item Announce where the robot should leave to after the test. 
+    \item When the test is held outside the arena, announce (way)point the robot to leave to
 \end{itemize}
 
 \newpage
--- a/tests/SPR.tex
+++ b/tests/SPR.tex
@@ -92,7 +92,7 @@ The referee needs to
 \begin{itemize}
     \item Announce the placement of the robots
     \item choose the volunteers for the second part of the test, and clearly explain the procedure to them.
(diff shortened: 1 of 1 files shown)


kyordhel:
There is no need to remove it.

That point is there because Audience don't want to see robots being removed, but autonomous robots.
See [comment in #310](https://github.com/RoboCupAtHome/RuleBook/issues/310#issuecomment-304836036)

LoyVanBeek:
So simply state that that robot must move out of the way is good enough alreayd, right? I was in doubt to put that in or to just remove it.

kyordhel:
Teams will start asking _what is move out of the way?_ and some make stupid decisions like going to the kitchen or bedroom (where there might be running a different test). Leaving the arena is a standard and provides a nice and clean end for all tests, so it should be in all but Restaurant, Tour Guide, and EEGPSR (for obvious reasons).

LoyVanBeek:
Updated

tests/SPR.tex
```diff
@@ -31,8 +31,8 @@ \subsection{Task}
     \end{itemize}
     This process is repeated with 5 (possibly) different people. 
     The game will end when the 5th question has been made, following the same distribution of questions as in the riddle game. The robot must answer the question without asking confirmation. Questions may be repeated once.
-
-    \item \textit{Leave:} The robot leaves the room through the designated door.
+    
+    \item \textbf{Leave} Th robot must leave the arena/test area to make place for the next robot to be tested.
```
kyordhel: ```The robot must leave the arena/test area after all questions have been asked or when instructed to do so. ```

tests/SPR.tex
```diff
@@ -92,6 +92,7 @@ \subsection{OC instructions}
 \begin{itemize}
     \item Announce the placement of the robots
     \item choose the volunteers for the second part of the test, and clearly explain the procedure to them.
+    \item Announce where the robot should leave to after the test.
```
kyordhel: I find this unnecessary as it over-specifies the function of the Exit door in General Rules. In any case I would say:

```When the test is held outside the arena, announce the leaving [way]point```

**Comment by LoyVanBeek** [comment]
Updated once more :-)

File: tests/SPR.tex [hunk 3]
```diff
@@ -92,6 +92,7 @@ \subsection{OC instructions}
 \begin{itemize}
     \item Announce the placement of the robots
     \item choose the volunteers for the second part of the test, and clearly explain the procedure to them.
+    \item When the test is held outside the arena, announce (way)point the robot to leave to
```
kyordhel: I don't get the syntax here. Instead of ```When the test is held outside the arena, announce (way)point the robot to leave to``` shouldn't be either one of

- ```When the test is held outside the arena, announce the leaving (way)point```
- ```When the test is held outside the arena, announce the (way)point through which the robot shall leave```

LoyVanBeek: I really shouldn't be doing this in between normal work :-)

kyordhel: We should be paid full time for doing this.

Issue #358, opened 2017-08-01
New scoring system


iocchi:
Dear all,
another great RoboCup@Home this year with a lot of great novelties (new leagues and new teams) and perfect organization (thanks Execs/TC/OC and LOC). Still we can improve and I think it is now time to discuss about the rules and the scoring system to solve some problems that we have noticed in the last years, but have not been solved yet.

The main problems in my opinion are:
1. Writing the rulebook is very complex and it is left on the hands of only a few people.
2. Implementing the tests by TC is difficult (not easy to delegate to volunteer referees, so requires TC to be present to evaluate every test) and sometimes the spirit of the test is not fully captured by the referees.
3. Teams are often not happy because of some rigid constraints in the rules, and sometimes they decide to not participate at all to some tests (probably because of their complexity). Some team members are not happy because the robot never reaches a point in which their development is tested, due to previous failures.
4. The Open tests (challenge and final) tends to be very conservative (no need to risk some autonomous behaviour or a "real" speech recognition phase) and sometimes seem to not seriously address the functionalities, tasks and scientific objectives of RoboCup@Home. In most cases, we still see more human presentations than robot tasks.
5. PhD students and researchers rarely find the subjects of their research in @Home tests. Also (maybe as a consequence) there are not so many publications from @Home teams.
6. Audience rarely see something interesting and they have difficulty in understanding what's going on.

I think these problems can be attached with a new structure of the tests and a new scoring system.
The current scoring system was introduced in 2008 to replace the previous one that assigned only boolean scores to tests. By introducing scores to intermediate steps we could better evaluate performance of teams. However, this introduced complexity in writing the rules and in evaluating the tests (scoresheets longer than a page) that increased a lot over the years.

On the flight back from Nagoya, I was reading how some sport disciplines that combine different elements are evaluated. For example, evaluation of gymnastics is based on a Code of Points, a Table of Elements describing what is expected in the tests and on a mechanism to compute the score of a performance. See https://en.wikipedia.org/wiki/Code_of_Points_(artistic_gymnastics)

I think RoboCup@Home has a similar concept: we want to integrate basic functionalities in different ways and reward the solution that better integrates many complex abilities.

Since the flight was long enough, I tried to think how to adapt a scoring system like this to @Home and I tried to put down an idea that is summarized below.

## RoboCup@Home score system based on sports (e.g., gymnastics) ##

[...]

Issue #353, opened 2017-08-01
Force manipulation in Storing Groceries


kyordhel:
Storing Groceries replaces the Manipulation (and object recognition) test. It's main goal is to manipulate objects, therefore all points must be scored by solving the problem, not by standing like a tripod and reporting. 

Here the changes:
- The robot needs to open a door
  - There is no help from human anymore
  - It gets additional time for doing so. 
  - the door can be opened at any time. 
- The robot must rearrange objects in the shelf by category
- The robot must move objects from the table to the shelf
- There is no specific order to perform the actions 
- Task driven: we expect empty table and arranged shelf. 
- The robot *may* deliver a pdf report
  - Only manipulated objects are considered.  
  - Allow us to give partial scoring for properly categorized miss recognised objects. 
  - PDF report is just a backup tool
- At least one of
  - cutlery (small)
  - cloth (amorphous)
  - alike (hard to recognise)
  - known light
  - known heavy

swachsmu:
I agree as long as the opening of the door is not mandatory (one half of the shelf is visible and usable). The robot should be able to score with simple objects. I would offer simple and difficult objects at the same time, so that the robot can select what to take. Difficult objects (cutlery or cloth) should score more points. Concentrating on the manipulation makes the test more understandable for the audience.

kyordhel:
@swachsmu I agree with the different object types. We would first need to define the _types_ though, such as small, light, heavy, crystal, cloth, bowl, etc.

Opening the door wouldn't be mandatory but necessary. Half of the objects can be seen and new objects can be placed in the open part of the shelf. If a robot want to see all the objects in the shelf, then it needs to open the door (i.e. no help from human). This, only if we want to make all robots are able to open doors. See #362.

kyordhel:
@airglow, @awesomebytes, @HideakiNagano21, @justinhart, @mnegretev, and rest of TC: your thoughts, comments and proposals, please. We also need a volunteer to update the test.

PR #369, opened 2017-12-04
Include the Filling the Dishwasher test (Stage I)

kyordhel:
A new test is introduced in Stage 1 for OPL and DSPL (mostly manipulation based), the Filling the Dishwasher test, granting up to 350 points for allocating up to 10 objects in a dishwasher.

General rules are modified to mention the Dishwasher in the arena specification, as well as the utensils to be used in this test, for which colourful examples have been chosen to lower the gap in
object recognition in this first try.

Also, the test has been designed to be compliant with issues #358, #353, and with P&G request.

**Remark:** For an enhanced version, please check #370


tests/FillingTheDishwasher.tex
```diff
@@ -0,0 +1,67 @@
...
+	\item \textbf{Location:} This test takes place in the kitchen of the arena. 
+	\item \textbf{Sink:} The sink has 10 utensils (See \ref{rule:tableware_kitchenware}) in it. There are at least two dishes (or bowls) two mugs (or glasses, or cups), one cooking pan (or pot), and one spoon (or any other cutlery).
+	\item \textbf{Start position:} The robot will start somewhere between the \textit{sink} and the \textit{dishwasher}.
+\end{enumerate}
+
+
+\subsection{Task}
+The robot must execute the following tasks in any order as long as there are utensils in the sink.
+\begin{itemize}
+	\item \textbf{Place cutlery:} The robot takes any cutlery from the sink and safely places it in the cutlery basket of the dishwasher.
```
LoyVanBeek: I'm fine with these being on a table. The whole point is to not have to bring the dishes anywhere, the robot does this all for you. 
To speed things up, we can put the table close to the dishwasher (they happen to be < 1.5m apart in my home)

tests/FillingTheDishwasher.tex
```diff
@@ -0,0 +1,67 @@
...
+	\item \textbf{Place cooking pans:} The robot takes any cooking pan (or pot, or utensil) from the sink and safely places it in the the dishwasher.
```
LoyVanBeek: Same for the pans: leave them in their usage locations, (the stove) and try to lay out the kitchen to have the dishwasher and stove close together to not waste time navigating

tests/FillingTheDishwasher.tex
kyordhel: The *sink* is like any other location. We can define a table to be the sink. If you check the General Rules, we are not asking for a real one so far. However, assuming we have one, I was thinking in something like this.

| Side by side  | Around the corner |
| ------------- | ------------- |
| fd0b781a869f6b643c0e916196fbb449  | 
img_3225  |

I don't want to overspecify the rulebook

tests/FillingTheDishwasher.tex
kyordhel: For simplicity, I would like to have everything in the same pace (the sink , which can be a table), to prevent the robot going here and there. Your suggestion makes sense, and I think changing *sink* to *stove* is no big deal.

airglow:
I like this test, may it's possible to include it already for RoboCup next year as a Storing Groceries replacement? I really like to see tasks that people usually avoid at home. Pumps up the motivation for actually solving it.

Some comments:

* We should specify if the stove is closed or open. I would say that it's to hard for Stage 1 to open the stove before being able to grasp the kitchen ware in 5 min (if teams want to do this it could give a time bonus).
* Maybe we could add alternative locations like on the coffee table or kitchen table. Usually the kitchen counter and the sink are high and hardly reachable for some robots. In order to avoid robots just standing in front of a sink those alternative locations are beneficial. Also it's maybe more interesting for the audience if the robot brings stuff from the couch table after a movie night into the dishwasher, than from the sink. It could also be described as a pick and place command for GPSR.
* It's really crucial that there is a dishwasher, already from the first set up day. When it will be available to late there most probably will be no team demonstrating anything or aiming for something not reusable.
* When this is targeting to be  Storing Groceries replacement we have to keep in mind that there is only a subset of objects involved here.
* I see there a chance that we will be able to test it already during RoboCup German Open.

kyordhel:
@airglow +1 to replacing the Storing Groceries with this one. Also helps in the roadmap towards integrating this with Set the Table as @balkce commented in #370 

On your comments:
- Stove is a flat surface with no doors. Not sure how it can be open. As for the dishwasher, it is open and the bottom rack is about 50% outside, so the robot can easily push or pull at convenience). Upper rack is stowed. This is specified in Additional Rules and Remarks.
- Stove and Sink are predefined locations, not an actual sink and stove. TC can define which surfaces are assigned for that purpose, so coffee table would be OK. In years to come, an actual sink should be used.
- At some point, object recognition will be considered as *solved* in the same way navigation is not rewarded with points any more. It makes sense to reduce the number of objects when migrating from boxes and cans to kitchenware and cleaning tools. Manipulation on the other hand, is not a solved problem to humans (my burnt hand confirms it).

balkce:
I'm all for simplifying the rulebook, and at first glance, this test could in theory replace Storing Groceries.

The only issue with this is that this test will be in Stage 1, and Clearing the Table will be in Stage II. Seems odd, since they're suppose to tell different parts of the same story.

What do you think about replacing Storing Groceries with the Dishwasher test, and passing to Stage II? I know there are some issues with that, but I'm just testing the waters.

kyordhel:
I'm opening an issue for that discussion.

kyordhel:

Issue #368, opened 2017-11-13
[DSPL] Official Standard Laptop for Toyota HSR (Ethernet wired)(**again**)


okdhryk:
Deadline: Thursday, November 30th, 2017

To the DSPL community:

Unfortunately #366 has closed
https://github.com/RoboCupAtHome/RuleBook/issues/366

Please let me suggest again using official standard laptop for HSR.

Members of 10 teams who participated in this year's competition should be strongly aware of the necessity of more powerful computing power.
I think that a new team that enters from the next year will feel its necessity if they actually use HSR.

In the original HSR there is a weak point in the ability of the built-in PC.
We can not fully implement the functions of object recognition or speech recognition.
This is boring for the audience and it is a major factor hindering the advancement of our research.

If it is difficult for all teams to purchase and use the same model,
Why do not you decide the upper limit of CPU and GPU capabilities?

Hiroyuki Okada
Team eR@sers, JAPAN

tkelestemur:
I strongly agree with you. Having an external backpack laptop will increase our success significantly. That being said, I think in the last issue, the problem was not finding a common model that can be bought everywhere. The problem was the lack of the participation in deciding the laptop model. I am not quite sure but there might be teams that do not follow Github issues for RoboCup. Also, @kyordhel mentioned 5  of 12 old teams + 9 candidates raised opinion but are 9 candidates aware of RuleBook repo? Should we expect them to raise opinion?

We can create a list of the proposed laptop models in an excel file with specs, prices, and availability and send it as an email to every participant. We can collect votes until the deadline. I can volunteer to do that if it makes sense to everyone.

justinhart:
I think that this is just diffusion of responsibility. Individual teams
don't feel the need to pipe up and assume that we're solving this.

Anyway, 1 step down from the latest laptop processor. 1 step down from the
latest Nvidia processor. That makes the laptop affordable. No team is going
to whine about spending $1400 on a laptop when they're shipping a crate
with a 500 lb robot in it to Canada.

kyordhel:
Thanks @okdhryk for continuing this discussion. I really hope you find consensus soon. Also thanks for the clarification about the number of teams (10 participants instead of 12, although we received 12 applications past year, hence my confusion), It is fixed now.

[...]

Issue #366, opened 2017-10-20
[DSPL] Official Standard Laptop for Toyota HSR (Ethernet wired)


kyordhel:
Deadline: **Sunday, October 29th, 2017**

To the DSPL community:

In response to your demands for a better juice extractor for the sensors of the HSR, the Executive Committe has decided to authorize the use of a **standard laptop** connected to the Toyota HSR via Ethernet cable, and located in the *TOYOTA HSR Backpack* provided by TOYOTA for this purpose.

The EC thinks it must be the DSPL community who provide the technical specifications of such laptop, as well as at least two commercial options. Requirements are as follows:

- Available worldwide
- Budget friendly
- Dimensions:
    - Maximum height: 430mm
    - Maximum width: 430mm
    - Maximum thickness: 48mm
    - Maximum weight: 5.0kg
- Software
    - All its components must be Linux compatible.
    - Same ROS version as the HSR 
- Laptop must fit in the backpack.

Other than that, there are no limits. You choose the rope, we hang you with it.

The EC will choose one of the proposed laptops as a Official Standard Laptop for teams to use during the competition. During the Robot Inspection, the OSL will also be inspected. No additional hardware will be allowed, including, but not limited to: battery, RAM, hard drives, SSD drives, USB devices, wireless cards, adapters, antennas, etc. Random checks might be done by the TC/OC/Referees at any time during the competition.

We are looking forward to read your proposals in this issue or in the OC mailing list.

tkelestemur:
It is great to hear this! Our major problems in Japan 2017 were because of the lack of computational power on the HSR. 

As Team Northeastern, we propose System76  Oryx Pro as the OSL. It is a Linux oriented brand and comes with Ubuntu installed. Some features:

- Thin aluminum alloy design
- NVIDIA GeForce GTX 10-Series GPUs
- 15″ or 17″ matte display
- HiDPI and IPS display options
- Up to 64 GB of memory
- Affordable

For more information : https://system76.com/laptops/oryx

Only disadvantage is battery life, but given the computational power on the laptop that is understandable.

justinhart:
I can get behind this. This is reasonably-priced and quite nice.

moriarty:
Popular choices here in Germany are Tuxedo Computers, and Schenker.

https://www.tuxedocomputers.com/index.php
https://www.mysn.de

They were the official hardware sponsor for RoboCup 2016 in Leipzig, for the simulation leagues.

[...]

PR #374, opened 2017-12-15
Add Official Standard Laptop specs and rules (compulsory for DSPL)

kyordhel:
This patch adds to the rulebook the "Official Standard Laptop for DSPL" section. This section makes mandatory the use of the OSL in DSPL, as well as restricting the use of any hardware attached to the OSL and allowing Referees/TC/OC to run random checks.

Robot Inspection is modified to include the inspection of the OSL. Changes are compliant with what was discussed and decided in #366 and #368

### DIFF
--- a/Setup.tex
+++ b/Setup.tex
@@ -119,6 +119,7 @@ Safety is the most important issue when interacting with humans and operating in
 		\begin{itemize}
 			\item Neat appearance
 			\item No modifications have been made
+			\item Specifications of the \iaterm{Official Standard Laptop}{OSL} (if required)
 		\end{itemize}
 		\item \textbi{Open Platform robots}
 		\begin{itemize}
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -29,9 +29,17 @@ Robots are allowed to use some form of external computing, for example in the fo
 
 \subsection{Official Standard Laptop for DSPL}
 \label{rule:osl_dspl}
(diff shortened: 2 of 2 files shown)


justinhart:
I think that we should discuss the point that Komei brought up before
merging this as-is. It seems like it should be okay for teams to opt out of
the laptop. Some teams may deploy solutions which don't utilize it, at
which point it may simply be a burden. I think that all teams are likely to
use it, as the hardware stock in the HSR is fixed, and therefore the only
way to keep pace with newer computer hardware is to use the laptop, but it
is possible that some teams would prefer to opt out.

Thoughts?

kyordhel:
I'm aware that some teams may not require the backpack at all, so turning
off the laptop would make no difference. However, this is an standard
league and that means all teams using exactly the same robot. I don't know
if the additional weight and volume will affect the HSR mobility or battery
duration, but definitely I don't want to figure it out during the
competition.

In any case, if Komei says it's optional, then it's optional. I'm making
the changes tomorrow to the branch (not merging yet).

kyordhel:
@komeisugiura requested changes have been made in 5f3fa86

balkce:
Frankly, even though it is a standard league, it doesn't seem fair to impose this onto teams that already have a robot for which they are already paying a lease or have spent a lot of money on. So, my vote is on putting it as optional.

Having said that, it has been overwhelming how many teams (if not all) wanted this option, so I'm pretty sure that even if we put that it was required, I don't see teams having issues with it.

kyordhel:
@balkce the last commit made it **optional**.

balkce:
In that case, it seems as though this discussion is over. I'm closing this branch.

PR #375, opened 2017-12-15
201x/tests/objectConfiguration

AMR-:
Suggestions for a new test: Replicating an object configuration

In this task, Pepper is allowed to view a configuration of objects. A human in a different room (or separated by a screen) has access to all of these object arranged arbitrarily. First, Pepper must describe the scene in front of it and instruct the human how to arrange their objects in an identical manner. Second, Pepper moves to the human’s area and views their setup. If the setup is wrong in
some fashion, Pepper will instruct the human to make the appropriate changes.

Details:
When the task begins, Pepper can view objects in front of it. Objects can include furniture such as tables, chairs, and stools, as well as smaller objects such as food containers and blocks. Objects could differ in size, color, and shape. Pepper will describe what is sees, using relational language. For example, “Place the green can on top of the yellow plate that is in the center of the table.” This first part of the task is limited to 5 minutes. Afterwards, a tournament official will go to the human’s configuration and introduce at least one additional error that Pepper will have to correct. At this time, Pepper is led to the human’s area for the second part of the task. After viewing the scene, Pepper should assess the errors present in the configuration and give additional instructions to fix them, such as “There should not be a coke bottle in this scene” or “The water bottle should be on the red plate, not the yellow plate.” Successive rounds might have increasing number of objects.

Scoring
Pepper will have 5 minutes for the first phase (description) and 5 minutes for the second phase (critiquing the human’s results). Teams will be scored according to the number of objects correctly placed in the first phase, both in terms of absolute position and in relationship to each other, and the number of successful corrections in the second phase

### DIFF
--- a/Rulebook.tex
+++ b/Rulebook.tex
@@ -133,7 +133,7 @@ Each ability test is designed to evaluate the average performance of the robot i
 Meanwhile, the integration test has been designed to evaluate how this abilities work together while solving a common task.
 
 The total score for ability and integration tests is the average of the best two performances out of preferably three performances (given the time constraints of a competition).
-The point of this is the both eliminate good and bad luck for the robots/teams and to get a more objective view of the performance,
+The point of this is the both elimate good and bad luck for the robots/teams and to get a more objective view of the performance,
   not to give teams time to tweak the robot between test performances. 
 
 \iterm{Following and Guiding} (demonstration for the audience) goes out of the arena and into the venue between the audience. 
--- a/Rulebook.tex
+++ b/Rulebook.tex
@@ -133,7 +133,7 @@ Each ability test is designed to evaluate the average performance of the robot i
 Meanwhile, the integration test has been designed to evaluate how this abilities work together while solving a common task.
 
 The total score for ability and integration tests is the average of the best two performances out of preferably three performances (given the time constraints of a competition).
-The point of this is the both elimate good and bad luck for the robots/teams and to get a more objective view of the performance, 
+The point of this is the both eliminate good and bad luck for the robots/teams and to get a more objective view of the performance,
   not to give teams time to tweak the robot between test performances. 
 
 \iterm{Following and Guiding} (demonstration for the audience) goes out of the arena and into the venue between the audience. 
--- a/Rulebook.tex
+++ b/Rulebook.tex
@@ -186,6 +186,9 @@ Of course, team A \& B can both perform the test in their own turn.
 \input{tests/EEGPSR}
 
 \newpage
+\input{tests/ObjectConfiguration}
+
+\newpage
 \input{tests/OpenChallenge}
 
 \newpage
--- a/score_sheets.tex
+++ b/score_sheets.tex
@@ -161,6 +161,12 @@
 \input{scoresheets/EEGPSR.tex}
 \end{scoresheet}
 
+% Object Configuration
+\renewcommand{\currentTest}{Object Configuration}
+\begin{scoresheet}
+\input{scoresheets/ObjectConfiguration.tex}
+\end{scoresheet}
+
 % Open Challenge
 \renewcommand{\currentTest}{Open Challenge}
 \begin{scoresheet}
--- a/scoresheets/ObjectConfiguration.tex
+++ b/scoresheets/ObjectConfiguration.tex
@@ -1,0 +1,27 @@
+The maximum time for this test is 6 minutes.
+
+\begin{scorelist}
+	\scoreheading{Describing the objects} % 100
+	%\scoreitem[n]{50 \/ n}{Description touches upon each of the n objects}
+	\scoreitem[10]{5}{Description touches upon each of the n objects}
+	\scoreitem[10]{5}{Description of each object is accurate}
+
+	\scoreheading{Move to Part 2} % 30
+	\scoreitem{30}{Follow official to the human room}
+
+	\scoreheading{Correct Human Error} % 130
+	\scoreitem{10}{Focus on the correct space}
+	\scoreitem[4]{13}{Identifies each of the p errors}
+	\scoreitem[4]{12}{Has an accurate correction for each of the p errors}
+	\scoreitem{20}{Correctly declares when the task is complete}
+
+	\scoreheading{Penalties}
+	\scoreitem[-1]{10}{During human error correction, noting something as an error that is not an error (Note: it has to be a clear issue of misunderstanding the scence, not resulting from the human insufficiently moving the object to the robot's exacting specifications.)}
+	
+	\setTotalScore{260}
+\end{scorelist}
+
+
+% Local Variables:
+% TeX-master: "Rulebook"
+% End:
--- a/scoresheets/ObjectConfiguration.tex
+++ b/scoresheets/ObjectConfiguration.tex
@@ -2,8 +2,7 @@ The maximum time for this test is 6 minutes.
 
 \begin{scorelist}
 	\scoreheading{Describing the objects} % 100
-	%\scoreitem[n]{50 \/ n}{Description touches upon each of the n objects}
-	\scoreitem[10]{5}{Description touches upon each of the n objects}
+	\scoreitem[10]{5}{Description touches upon each of the 10 objects}
 	\scoreitem[10]{5}{Description of each object is accurate}
 
 	\scoreheading{Move to Part 2} % 30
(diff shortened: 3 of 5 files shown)


tests/ObjectConfiguration.tex
```diff
@@ -0,0 +1,75 @@
...
+\begin{itemize}
+	\item \textbf{Robot Starting Room}: any room.
+	\item \textbf{Human Starting Room}: a room adjacent to Robot Starting Room.
+	\item \textbf{Human Collaborator}: The human collaborator is in the Human Starting Room.
+	\item \textbf{Robot Objects} 10 objects are arranged in a space in front of the robot, potentially on top of a table or other furniture.
+	\item \textbf{Human Objects:} In this room there are the identical objects as in the other room, but arranged differently.
+\end{itemize}
+
+\subsection{Task Part 1}
+
```
kyordhel: Task requires complete redesign. It is too fine-grained and the goal of each of the steps is underspecified. It is unclear where the collaboration is leading to.

scoresheets/ObjectConfiguration.tex
```diff
@@ -0,0 +1,27 @@
+The maximum time for this test is 6 minutes.
+
+\begin{scorelist}
+	\scoreheading{Describing the objects} % 100
+	%\scoreitem[n]{50 \/ n}{Description touches upon each of the n objects}
+	\scoreitem[10]{5}{Description touches upon each of the n objects}
```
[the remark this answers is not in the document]
kyordhel: I don't think is a good idea to have $n$ in any scoresheet. All teams should be able to score the same.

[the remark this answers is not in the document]
AMR-: the original idea was to have n objects, with each being worth (50/n) points, so that the "Describing the objects" in total is worth 50 points always, and n constant for the competition.

But perhaps having a constant value is better, so I changed n to 10.   Feel free to let me know which you like better

Rulebook.tex
```diff
@@ -133,7 +133,7 @@ \chapter{Tests in Stage I}
 Meanwhile, the integration test has been designed to evaluate how this abilities work together while solving a common task.
 
 The total score for ability and integration tests is the average of the best two performances out of preferably three performances (given the time constraints of a competition).
-The point of this is the both elimate good and bad luck for the robots/teams and to get a more objective view of the performance,
```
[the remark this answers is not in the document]
AMR-: i'll make a different PR for the typo correction

tests/ObjectConfiguration.tex
AMR-: By "unclear where the collaboration is leading to" do you mean you would like there to be more of a story justification?

Rulebook.tex
```diff
@@ -133,7 +133,7 @@ \chapter{Tests in Stage I}
 Meanwhile, the integration test has been designed to evaluate how this abilities work together while solving a common task.
 
 The total score for ability and integration tests is the average of the best two performances out of preferably three performances (given the time constraints of a competition).
-The point of this is the both elimate good and bad luck for the robots/teams and to get a more objective view of the performance, 
+The point of this is the both eliminate good and bad luck for the robots/teams and to get a more objective view of the performance,
```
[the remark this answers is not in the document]
AMR-: will make new PR

scoresheets/ObjectConfiguration.tex
```diff
@@ -0,0 +1,27 @@
+The maximum time for this test is 6 minutes.
```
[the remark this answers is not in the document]
AMR-: done

scoresheets/ObjectConfiguration.tex
```diff
@@ -0,0 +1,27 @@
...
+	\scoreitem[10]{5}{Description touches upon each of the n objects}
+	\scoreitem[10]{5}{Description of each object is accurate}
+
+	\scoreheading{Move to Part 2} % 30
+	\scoreitem{30}{Follow official to the human room}
+
+	\scoreheading{Correct Human Error} % 130
+	\scoreitem{10}{Focus on the correct space}
+	\scoreitem[4]{13}{Identifies each of the p errors}
+	\scoreitem[4]{12}{Has an accurate correction for each of the p errors}
```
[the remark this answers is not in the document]
AMR-: added in commit 3 (dde38a4)

tests/ObjectConfiguration.tex
```diff
@@ -0,0 +1,75 @@
...
+\begin{enumerate}
+
+	\item \textbf{Analyzing:} Robot views the object configuration in front of it.
+
+	\item \textbf{Description:} Robot describes the configuration to the human collaborator.  The robot can use relational languae or any other means of description.
```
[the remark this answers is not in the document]
AMR-: fixed in commit 3 (dde38a4)

tests/ObjectConfiguration.tex
kyordhel: The test is extremely hard to follow because there is no intuitive goal. *Serve Breakfast*, *Cocktail Party*, *Follow Me*, *Fetch and Carry*, etc., give you a clue about the task and goal from the very title. Any person can easily grasp the context and what's the ultimate goal. In this test I couldn't find a direct and intuitive application for the task, something that people in the audience can easily relate to something they do in daily life.

The closest thing that came to my mind was when I move all objects from one surface to another while cleaning and then I can't remember what was the original configuration. Here It would be useful if the robot reminds me whether my ancient *Ming pot* was left or right of the stand with the *Scroll of Truth*, stuff that is most valuable to me and the robot is not allowed to touch. However:
- I am in the same room
- I am interacting with exactly the same objects
- Either I ask the robot or the robot realizes the mistake.
Therefore, I see no reason why I should replicate a setup that is in another room, nor why should I receive instructions from anyone (human or robot, that's extremely annoying).

However, I am not sure if what I mentioned above actually has something to do with the proposed test based only in the content of the tex file.

Designing a test is a very difficult task for it needs to be cross-culture understandable, so the first advice is give your test *as is* to read to a 3rd person (better if outside the work group) and ask them to explain back to you. Don't explain, just take notes and ask why that person makes that interpretation.

score_sheets.tex
```diff
@@ -161,6 +161,12 @@
 \input{scoresheets/EEGPSR.tex}
 \end{scoresheet}
 
+% Object Configuration
```
kyordhel: Wrong place. The test is a Stage I test and is located between EEGPSR and Open Challenge which are Stage II

scoresheets/ObjectConfiguration.tex
```diff
@@ -2,23 +2,23 @@
 
 \begin{scorelist}
 	\scoreheading{Describing the objects} % 100
-	%\scoreitem[n]{50 \/ n}{Description touches upon each of the n objects}
-	\scoreitem[10]{5}{Description touches upon each of the n objects}
+	\scoreitem[10]{5}{Description touches upon each of the 10 objects}
```
kyordhel: I find "Description touches upon" an unusual construct and its sense is quite ambiguous. Scoresheets must be straightforward. Of course if task changes, this needs also to change.

scoresheets/ObjectConfiguration.tex
```diff
@@ -2,23 +2,23 @@
...
 	\scoreitem[10]{5}{Description of each object is accurate}
 
 	\scoreheading{Move to Part 2} % 30
 	\scoreitem{30}{Follow official to the human room}
 
 	\scoreheading{Correct Human Error} % 130
 	\scoreitem{10}{Focus on the correct space}
-	\scoreitem[4]{13}{Identifies each of the p errors}
+	\scoreitem[4]{13}{Identifies each of the errors}
```
kyordhel: Remove each, you are scoring once per error, so it should be something like _Clearly state the nature of an error_. Plus, I feel 4x13 leading to a weird number.

tests/ObjectConfiguration.tex
```diff
@@ -1,6 +1,6 @@
 \section{Replicating a Configuration of Objects}
 
-The robot has to understand relationships between objects, instruct a human to replicate that configuration of objects, and fix errors that the human makes.
+The robot has to understand relationships between objects, instruct a human operator to replicate that configuration of objects, and fix all errors made by the operator.
```
kyordhel: Understand is a strong word here. I don't understand why the pots in my mother balcony are in certain order, but I am able to remember and replicate its configuration. However, I do understand why fork and knife are located the way they are in a table setup so I can explain it.

tests/ObjectConfiguration.tex
```diff
@@ -21,7 +21,7 @@ \subsection{Task Part 1}
 
 	\item \textbf{Analyzing:} Robot views the object configuration in front of it.
 
-	\item \textbf{Description:} Robot describes the configuration to the human collaborator.  The robot can use relational languae or any other means of description.
+	\item \textbf{Description:} Robot describes the configuration to the human collaborator.  The robot can use relational language or any other means of description.
```
kyordhel: Double space after the dot.
I haven't clear what is "relational language", meaning that I, as a human, couldn't execute the task without further explanation.

kyordhel:

PR #377, opened 2017-12-18
Adding Balloon Sorting test

AMR-:
Balloons are in room 1. Pepper must gather the balloons from room 1 and put them in room 2, while following instructions. A human participant might instruct it saying, “bring at least 5 red balloons to the living room” or “leave 2 blue balloons in the other room.” This is a competitive, timed task.

### Details
There are two rooms in this task. They may be labeled, such as the “bedroom” and the “living room”. At the beginning of the task there are 10 balloons total in one of the rooms. There is a ceiling fence in both rooms so that the balloons do not rise above 8 feet high. Balloons are yellow, red, green, and blue, in any number of each color. Each balloon has a string with a loop on the bottom at the height of Pepper’s head that can serve as a handle that Pepper is capable of grabbing.

A tournament official gives Pepper instructions regarding bringing balloons to the living room. These instructions could be piecemeal (“Please bring 3 red balloons. Now, please bring 2 blue balloons.”) or more expansive (“Please bring all balloons to the living room.”). Instructions could change over time, as a human changes their mind. For example, “I want 4 red balloons, not 3.” Pepper could be instructed to bring balloons back from room 2 to room 1, too, as in, “Remove the yellow balloons from the living room!” If Pepper receives “con- flicting” instructions, it should follow the most recent instruction (as well as previous non-conflicting instructions). Humans collectively will issue no more than 5 instructions total over the course of the task.

### Scoring
Pepper will have 10 minutes to complete the task. The score will take into account what instructions Pepper was able to complete within the allotted time (judged by number of balloons moved out of total balloons to be moved). Balloons moved in error will count against the score.

### DIFF
--- a/Rulebook.tex
+++ b/Rulebook.tex
@@ -186,6 +186,9 @@ Of course, team A \& B can both perform the test in their own turn.
 \input{tests/EEGPSR}
 
 \newpage
+\input{tests/BalloonSorting}
+
+\newpage
 \input{tests/OpenChallenge}
 
 \newpage
--- a/score_sheets.tex
+++ b/score_sheets.tex
@@ -161,6 +161,12 @@
 \input{scoresheets/EEGPSR.tex}
 \end{scoresheet}
 
+% Balloon Sorting
+\renewcommand{\currentTest}{Ballon Sorting}
+\begin{scoresheet}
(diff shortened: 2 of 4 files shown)


kyordhel:

Issue #386, opened 2018-02-03
[SSPL] Definition External Computing Device | Form Factor & Mobility


warp1337:
I just skimmed through the rulebook and some other related issues, e.g., #366 #279 #243 

So ... the rulebook says:

> [...] the  robot  has  to  maintain  full  autonomy  when  using external  computing resources, meaning there may not be a human giving the robot any kind of instructions via external computing resources.  It is up to the team to prove to the Technical Committee (TC)  that  there  was  no  cheating  introduced  via  the external  computing  resources [...]
> 

Concerning this statement:

1) Last year, at Nagoya, I noticed teams setting up high performance desktop PCs (the ones you see in gaming commercials, e.g, 2x NVIDIA TitanX, Intel Xeon 4 Ghz, etc...) in front of the arena where the external computing devices had to be placed. I guess that's okay, so far, since we don't have any rule saying you can't bring your own high performance computer cluster. My 'problem' with that is, those machines were blocking the preparation area/table where other teams were placing their external devices too, i.e., usually laptops. The table was totally crowded and the space taken by cables, screens (!), keyboards, and mice of the desktop PCs was just ridiculous. 

**Thus, my proposal for the rulebook is that external computing devices have to be mobile, e.g, have to be setup before the task and removed after the task is done.** 

2) Based on 1) the threat of 'remote controlling the robot' is inherently mitigated since external computing devices have to be setup before and removed after the task. In Nagoya we also had the "hands-off the device rule" as soon as the ref starts the task --- which can be enforced and supervised by the refs.

LoyVanBeek:
Sound very reasonable. Perhaps we can add some separate rack or table people put these devices in/on.

warp1337:
Thx @LoyVanBeek, but this would actually encourage bringing high performance PCs doesn't it (that are additionally connected to the network all the time)?

LoyVanBeek:
Maybe with a power switch per team, that gets switched on only a few minutes before a test and off directly after :-). 

But having a designated place for such external devices in a place easy to keep an eye on could help. Have ~2 of those spots, one for the active team and one for the team setting up.

warp1337:
So, how do we write this down in the rulebook?

kyordhel:
Some considerations on my side:
- OC-wise I'm a little bit uncomfortable with teams moving hardware between tests. In all Stage 1 there is no Setup time, and often we end up slightly behind schedule.
[...]

Issue #243, opened 2017-03-01
External Computing


moriarty:
This issues relates to: #128 and #173 which have been closed and answered. And the rules are clear enough for now in 3.6 of the RuleBook, but they could use some further clarifications. One reason I say for now, is that I haven't seen any teams take advantage of them. Has someone been keeping track?

I think it's great opportunity, and while competing last year our team considered it... Not to use an external cloud based option, but to have our own developer box on the network.

I also agree with the point someone brought up in the previous issue about fairness. Especially in the new Standard Platform leagues. The variety of manipulators seen in the OPL 

The main issue, was according to Section 3.6 item 5 "Availability" require us to make this publicly available, for use by robots of other teams, well before and after the competition.

It's a fine rule, and it would be great to see, but it's just hard to implement on the team side.
Last year, I did spend some time looking into it, but in the end we upgrade one of our onboard computers and didn't think it was worth it...

Considerations we had:
1.  Onsite internet connection has been bad in previous competitions. In 2 Robocup competitions, I have returned to the hotel to download some packages... To get around this, we travel with our own Linux, ROS and git mirror. But for this reason we only seriously considered having an extra machine in the arena, connected to the network, not renting or using the university resources.
2. Cloud Computer resources are expensive, and most Robocup budgets stretched... Most universities have some access to compute clusters, however, these resources are often impossible to allow other non registered users to access them. They're usually shared resources, and hard to schedule on demand jobs, or switch tasks. Having these resources available, means there is little incentive spend our budgets to use publicly available resources like AWS. Even if we could have gotten access to our So what does it mean- publicly available for use by the other teams?
3. If it needs to be available to other teams, and it's a custom solution, then we also need to support the other teams? what if it's not stable?
4. API vs. Computing Resources. Running a service continuously and publicly and making the API available, vs leaving a machine available where any other team can spin up a node (docker or openstack, vmware etc- but that requires restricting the teams during test runs) And how to limit which team access it during a test in the case of a shared resource.
[...]

PR #388, opened 2018-02-05
Add subsection for external computing devices

warp1337:
This is my first draft as discussed in #386.

### DIFF
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -34,8 +34,8 @@ Robots are allowed to use some form of external computing, for example in the fo
   		(see~\ref{rule:robot_external_computing}), must be placed at a designated location close to the arena.
   		The location features a table (or otherwise suitable furnishing) and an ethernet switch (wired)
         that connects to the arena WIFI. In the remainder of this document, this location is called
-  		\textbf{E}xternal \textbf{C}omputing\textbf{R}esource \textbf{A}rea (ECRA).
-  \item \textbf{Procedure:} before a teams' test run, external comp. devices msut be setup in time.
+  		\textbf{E}xternal \textbf{C}omputing \textbf{R}esource \textbf{A}rea (ECRA).
+  \item \textbf{Procedure:} before a teams' test run, external comp. devices must be setup in time.
   		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
         teams adequate space and time to setup their devices. No team should ever occupy the ECRA continuously.
   		As soon as the referee indicates the teams' test slot starts, team members are strictly forbidden to touch
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -26,6 +26,34 @@ Robots are allowed to use some form of external computing, for example in the fo
 
 \textbf{Remark:} Teams are allowed to use their own software in the external computing devices (not only cloud services). This software must be publicly available to other teams for scientific purposes (evaluation, test, and benchmarking), as well as for TC for inspection. Although open-sourcing the software is not mandatory, this practice is advised and encouraged by the league.
 
+\subsection{External Computing Devices on site for SSPL (Hardware)}
+\label{rule:robot_external_computing_devices_sspl}
+
+\begin{itemize}
+  \item \textbf{Location:} all external computing devices, approved by the \iaterm{Technical Committee}{TC}
+  		(see~\ref{rule:robot_external_computing}), must be placed at a designated location close to the arena.
+  		The location features a table (or otherwise suitable furnishing) and an ethernet switch (wired)
+        that connects to the arena WIFI. In the remainder of this document, this location is called
+  		\textbf{E}xternal \textbf{C}omputing\textbf{R}esource \textbf{A}rea (ECRA).
+  \item \textbf{Procedure:} before a teams' test run, external comp. devices msut be setup in time.
+  		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
+        teams adequate space and time to setup their devices. No team should ever occupy the ECRA continuously.
+  		As soon as the referee indicates the teams' test slot starts, team members are strictly forbidden to touch
+        their external comp. devices until the current test run is over. Breaking this ``hands-off'' rule
+        is penalized with 0 (zero) points for the corresponding test. Using external comp. devices during the test
+        from any other location than the ECRA will be penalized with \textbf{at least 0 (zero) points for the test}.
+  \item \textbf{Limitations:} external computing devices currently have no limitation concerning computation
+  		power and form factor. However, in order to maintain enough space and a	clear arrangement (e.g. wrt
+        inspectability for referees), stand-alone screens are \textbf{not allowed}. Mice and keyboards connected
+        to external computing devices are allowed.
+  \item \textbf{Recommendations:} show up in time and setup up your devices early enough. If your team is up last,
+        do not setup first. Consider the ECRA a FIFO queue. There will be enough space at the ECRA to establish a
+        buffer (of teams), so if you show up in time and everyone removes their equipment the transition will be smooth.
+\end{itemize}
+
+The referees, Technical Committee, and Organizing Committee members may run random checks anytime during the competition
+to check if a team occupies the ECRA continuously or devices outside the ECRA are used to control the robot.
+
 
 \subsection{Official Standard Laptop for DSPL}
 \label{rule:osl_dspl}
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -26,33 +26,36 @@ Robots are allowed to use some form of external computing, for example in the fo
 
 \textbf{Remark:} Teams are allowed to use their own software in the external computing devices (not only cloud services). This software must be publicly available to other teams for scientific purposes (evaluation, test, and benchmarking), as well as for TC for inspection. Although open-sourcing the software is not mandatory, this practice is advised and encouraged by the league.
 
-\subsection{External Computing Devices on site for SSPL (Hardware)}
-\label{rule:robot_external_computing_devices_sspl}
+\subsection{On-Site External Computing Devices}
+\label{rule:robot_external_computing_devices}
 
 \begin{itemize}
-  \item \textbf{Location:} all external computing devices, approved by the \iaterm{Technical Committee}{TC}
-  		(see~\ref{rule:robot_external_computing}), must be placed at a designated location close to the arena.
-  		The location features a table (or otherwise suitable furnishing) and an ethernet switch (wired)
-        that connects to the arena WIFI. In the remainder of this document, this location is called
-  		\textbf{E}xternal \textbf{C}omputing \textbf{R}esource \textbf{A}rea (ECRA).
-  \item \textbf{Procedure:} before a teams' test run, external comp. devices must be setup in time.
+  \item \textbf{Location:} All external computing devices, approved by the \iaterm{Technical Committee}{TC}
+  		(see~\refsec{rule:robot_external_computing}), must be placed at the designated location, e.g., a table close
+        to the arena, referred hereinafter as the \iaterm{\textbf{E}xternal \textbf{C}omputing \textbf{R}esource
+        \textbf{A}rea} {ECRA}. The OC will announce the ECRA during setup days. A switch connected to the arena
+        wireless network will be available to teams.
+  \item \textbf{Procedure:} Before a teams' test run, external devices must be setup in time.
   		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
-        teams adequate space and time to setup their devices. No team should ever occupy the ECRA continuously.
-  		As soon as the referee indicates the teams' test slot starts, team members are strictly forbidden to touch
-        their external comp. devices until the current test run is over. Breaking this ``hands-off'' rule
-        is penalized with 0 (zero) points for the corresponding test. Using external comp. devices during the test
-        from any other location than the ECRA will be penalized with \textbf{at least 0 (zero) points for the test}.
-  \item \textbf{Limitations:} external computing devices currently have no limitation concerning computation
+        teams adequate space and time to setup their devices. As soon as the referee indicates the teams' test slot
+        starts, team members are strictly forbidden to touch their external devices until the current test run is over.
+        Breaking this ``hands-off'' rule is penalized with -25 points for the corresponding test run. Also, not removing
+        equipment within a reasonable time frame is penalized with -25 points for the tests run.
+  \item \textbf{Limitations:} External computing devices currently have no limitation concerning computation
   		power and form factor. However, in order to maintain enough space and a	clear arrangement (e.g. wrt
         inspectability for referees), stand-alone screens are \textbf{not allowed}. Mice and keyboards connected
         to external computing devices are allowed.
-  \item \textbf{Recommendations:} show up in time and setup up your devices early enough. If your team is up last,
-        do not setup first. Consider the ECRA a FIFO queue. There will be enough space at the ECRA to establish a
-        buffer (of teams), so if you show up in time and everyone removes their equipment the transition will be smooth.
+  \item \textbf{Remark:} Please keep the fair play rule~\ref{rule:fairplay} in mind. The referees, technical committee,
+        and organizing committee members may run random checks anytime during the competition in order to check if a team
+        occupies the ECRA continuously and whether devices outside the ECRA are used to, e.g., remote control a robot.
+        If one of the aforementioned situations is discovered, disqualification from at least the current test --- or
+        even the whole competition is considered an option. % warp1337: we might need another strategy before and after
+        the tests, when the arena is open to everyone
+  % kyordhel says: not neccessary ;)
+  % \item \textbf{Recommendations:} Show up in time and setup up your devices early enough. If your team is up last,
+  %      do not setup first. Consider the ECRA a FIFO queue. There will be enough space at the ECRA to establish a
+  %      buffer (of teams), so if you show up in time and everyone removes their equipment the transition will be smooth.
 \end{itemize}
-
-The referees, Technical Committee, and Organizing Committee members may run random checks anytime during the competition
-to check if a team occupies the ECRA continuously or devices outside the ECRA are used to control the robot.
 
 
 \subsection{Official Standard Laptop for DSPL}
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -39,24 +39,19 @@ Robots are allowed to use some form of external computing, for example in the fo
   		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
         teams adequate space and time to setup their devices. As soon as the referee indicates the teams' test slot
         starts, team members are strictly forbidden to touch their external devices until the current test run is over.
-        Breaking this ``hands-off'' rule is penalized with -25 points for the corresponding test run. Also, not removing
+        Breaking this \quotes{hands-off} rule is penalized with -25 points for the corresponding test run. Also, not removing
         equipment within a reasonable time frame is penalized with -25 points for the tests run.
   \item \textbf{Limitations:} External computing devices currently have no limitation concerning computation
   		power and form factor. However, in order to maintain enough space and a	clear arrangement (e.g. wrt
         inspectability for referees), stand-alone screens are \textbf{not allowed}. Mice and keyboards connected
         to external computing devices are allowed.
-  \item \textbf{Remark:} Please keep the fair play rule~\ref{rule:fairplay} in mind. The referees, technical committee,
+\end{itemize}
+
+\textbf{Remark:} Please keep the fair play rule~\ref{rule:fairplay} in mind. The referees, technical committee,
         and organizing committee members may run random checks anytime during the competition in order to check if a team
         occupies the ECRA continuously and whether devices outside the ECRA are used to, e.g., remote control a robot.
         If one of the aforementioned situations is discovered, disqualification from at least the current test --- or
-        even the whole competition is considered an option. % warp1337: we might need another strategy before and after
-        the tests, when the arena is open to everyone
-  % kyordhel says: not neccessary ;)
-  % \item \textbf{Recommendations:} Show up in time and setup up your devices early enough. If your team is up last,
-  %      do not setup first. Consider the ECRA a FIFO queue. There will be enough space at the ECRA to establish a
-  %      buffer (of teams), so if you show up in time and everyone removes their equipment the transition will be smooth.
-\end{itemize}
-
+        even the whole competition is considered an option.
 
 \subsection{Official Standard Laptop for DSPL}
 \label{rule:osl_dspl}
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -29,29 +29,51 @@ Robots are allowed to use some form of external computing, for example in the fo
 \subsection{On-Site External Computing Devices}
 \label{rule:robot_external_computing_devices}
 
+% Second iteration
 \begin{itemize}
-  \item \textbf{Location:} All external computing devices, approved by the \iaterm{Technical Committee}{TC}
-  		(see~\refsec{rule:robot_external_computing}), must be placed at the designated location, e.g., a table close
-        to the arena, referred hereinafter as the \iaterm{\textbf{E}xternal \textbf{C}omputing \textbf{R}esource
-        \textbf{A}rea} {ECRA}. The OC will announce the ECRA during setup days. A switch connected to the arena
-        wireless network will be available to teams.
-  \item \textbf{Procedure:} Before a teams' test run, external devices must be setup in time.
+  \item \textbf{Location:} All \iaterm{External Computing Devices}{ECDs} must be approved by the \iaterm{Technical Committee}{TC}
+  		(see~\refsec{rule:robot_external_computing}) and presented during the robot inspection. Furthermore, ECDs
+  		that exceed the form factor of a laptop (e.g. a desktop PC), referred hereinafter as servers, must be placed and
+  		setup in a designated area that is announced by the \iaterm{Technical Committee}{TC} during setup days.
+  		A server must not have a screen, mouse, keyboard, bluetooth or any other peripheral device attached to it.
+  		Servers cannot be touched by anyone (e.g. for plugging/un-plugging cables) during the competition phase.
+  \item \textbf{Procedure:} Before a teams' test run, ECDs (except servers) must be setup in time in the
+		\iaterm{\textbf{E}xternal \textbf{C}omputing \textbf{R}esource \textbf{A}rea} {ECRA}. The OC will announce the
+        ECRA during setup days. A switch connected to the arena wireless network will be available to teams in the ECRA.
   		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
         teams adequate space and time to setup their devices. As soon as the referee indicates the teams' test slot
-        starts, team members are strictly forbidden to touch their external devices until the current test run is over.
+        starts, team members are strictly forbidden to touch their external devices until their current test run is over.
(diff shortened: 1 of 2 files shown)


general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
 
 \textbf{Remark:} Teams are allowed to use their own software in the external computing devices (not only cloud services). This software must be publicly available to other teams for scientific purposes (evaluation, test, and benchmarking), as well as for TC for inspection. Although open-sourcing the software is not mandatory, this practice is advised and encouraged by the league.
 
+\subsection{External Computing Devices on site for SSPL (Hardware)}
```
kyordhel: Let's generalize to all leagues. This subsection is then renamed to:
*On-site External Computing Devices*

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+\label{rule:robot_external_computing_devices_sspl}
+
+\begin{itemize}
+  \item \textbf{Location:} all external computing devices, approved by the \iaterm{Technical Committee}{TC}
```
kyordhel: Style changes:
- All external (capitalize A)

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+  		(see~\ref{rule:robot_external_computing}), must be placed at a designated location close to the arena.
+  		The location features a table (or otherwise suitable furnishing) and an ethernet switch (wired)
+        that connects to the arena WIFI. In the remainder of this document, this location is called
+  		\textbf{E}xternal \textbf{C}omputing \textbf{R}esource \textbf{A}rea (ECRA).
```
kyordhel: ECRA? Nice! I would rephrase paragraph to:

*must be placed at **the** designated location (e.g. a table close to the arena), referred hereinafter as \iaterm{\textbf{E}xternal \textbf{C}omputing \textbf{R}esource \textbf{A}rea}{ECRA}.*

Let's leave the Switch thing as another item:

- The OC will announce the ECRA during setup days.
- A switch connected to the Arena Wireless Network will be available to teams.

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+  \item \textbf{Procedure:} before a teams' test run, external comp. devices must be setup in time.
+  		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
+        teams adequate space and time to setup their devices. No team should ever occupy the ECRA continuously.
+  		As soon as the referee indicates the teams' test slot starts, team members are strictly forbidden to touch
+        their external comp. devices until the current test run is over. Breaking this ``hands-off'' rule
+        is penalized with 0 (zero) points for the corresponding test. Using external comp. devices during the test
```
kyordhel: Penalties must be also stated in the section of special bonuses and penalties

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+        from any other location than the ECRA will be penalized with \textbf{at least 0 (zero) points for the test}.
```
kyordhel: Penalties must be also stated in the section of special bonuses and penalties

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+\subsection{External Computing Devices on site for SSPL (Hardware)}
+\label{rule:robot_external_computing_devices_sspl}
+
+\begin{itemize}
+  \item \textbf{Location:} all external computing devices, approved by the \iaterm{Technical Committee}{TC}
+  		(see~\ref{rule:robot_external_computing}), must be placed at a designated location close to the arena.
+  		The location features a table (or otherwise suitable furnishing) and an ethernet switch (wired)
+        that connects to the arena WIFI. In the remainder of this document, this location is called
+  		\textbf{E}xternal \textbf{C}omputing \textbf{R}esource \textbf{A}rea (ECRA).
+  \item \textbf{Procedure:} before a teams' test run, external comp. devices must be setup in time.
```
kyordhel: I find the paragraph a little bit long. We must also avoid abbreviations like *comp.*. My two cents:
*There may be up to \textbf{two} teams in ECRA at any time, the testing team and the upcoming one. Immediately after the test, the team must remove all equipment from ECRA.*

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+  		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
+        teams adequate space and time to setup their devices. No team should ever occupy the ECRA continuously.
+  		As soon as the referee indicates the teams' test slot starts, team members are strictly forbidden to touch
+        their external comp. devices until the current test run is over. Breaking this ``hands-off'' rule
+        is penalized with 0 (zero) points for the corresponding test. Using external comp. devices during the test
+        from any other location than the ECRA will be penalized with \textbf{at least 0 (zero) points for the test}.
+  \item \textbf{Limitations:} external computing devices currently have no limitation concerning computation
```
kyordhel: Capitalize E

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+  		power and form factor. However, in order to maintain enough space and a	clear arrangement (e.g. wrt
+        inspectability for referees), stand-alone screens are \textbf{not allowed}. Mice and keyboards connected
+        to external computing devices are allowed.
+  \item \textbf{Recommendations:} show up in time and setup up your devices early enough. If your team is up last,
```
kyordhel: This item is unnecessary.

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,34 @@ \section{External computing}\label{rule:robot_external_computing}
...
+        do not setup first. Consider the ECRA a FIFO queue. There will be enough space at the ECRA to establish a
+        buffer (of teams), so if you show up in time and everyone removes their equipment the transition will be smooth.
+\end{itemize}
+
+The referees, Technical Committee, and Organizing Committee members may run random checks anytime during the competition
```
kyordhel: Add a \testbf{Remark:} at the beginning of the paragraph. A reference to the Fair Play rule is necessary as well as specifying that the team will be disqualified (is stronger that mentioning just a zero-point penalty) for the test, or even from the entire competition.

warp1337:
Will include the proposed changes soon so that all OC/TC people can have a look at it.

general_rules/ExternalComputing.tex
warp1337: Two teams is too little IMHO, let's make it three.

warp1337:
@kyordhel Comments are incorporated. To whom it may concern ...

LoyVanBeek:
It's on the back of my head to review somewhere this week. 
@awesomebytes @airglow any comments?

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,37 @@ \section{External computing}\label{rule:robot_external_computing}
...
+  \item \textbf{Location:} All external computing devices, approved by the \iaterm{Technical Committee}{TC}
+  		(see~\refsec{rule:robot_external_computing}), must be placed at the designated location, e.g., a table close
+        to the arena, referred hereinafter as the \iaterm{\textbf{E}xternal \textbf{C}omputing \textbf{R}esource
+        \textbf{A}rea} {ECRA}. The OC will announce the ECRA during setup days. A switch connected to the arena
+        wireless network will be available to teams.
+  \item \textbf{Procedure:} Before a teams' test run, external devices must be setup in time.
+  		Immediately after a teams' test slot, all equipment must be removed from the ECRA in order to give ensuing
+        teams adequate space and time to setup their devices. As soon as the referee indicates the teams' test slot
+        starts, team members are strictly forbidden to touch their external devices until the current test run is over.
+        Breaking this ``hands-off'' rule is penalized with -25 points for the corresponding test run. Also, not removing
```
kyordhel: Replace ``hands-off'' with \quotes{hands-off} or \enquote{hands-off} (If csquote package is not there, add it).

general_rules/ExternalComputing.tex
```diff
@@ -26,6 +26,37 @@ \section{External computing}\label{rule:robot_external_computing}
...
+        equipment within a reasonable time frame is penalized with -25 points for the tests run.
+  \item \textbf{Limitations:} External computing devices currently have no limitation concerning computation
+  		power and form factor. However, in order to maintain enough space and a	clear arrangement (e.g. wrt
+        inspectability for referees), stand-alone screens are \textbf{not allowed}. Mice and keyboards connected
+        to external computing devices are allowed.
+  \item \textbf{Remark:} Please keep the fair play rule~\ref{rule:fairplay} in mind. The referees, technical committee,
```
kyordhel: Remark outside the itemize environment.

warp1337:
All done. *bump*

kyordhel:
Awesome. Thank you @warp1337.
Let's wait for approval from another TC.

moriarty:
I'm not a TC. The changes look mostly great.

And I agree, last year, was a little tough with a lot of teams basically having full setups with monitors etc... at what should have just contained the external compute devices.
And, as OC- it was hard to see who had hands on their keyboards/mice and I did need to tell teams to back away, because there was such a pile of keyboards/mice/screens.

I just wanted to ask- how long do we intend for this process to be in place?
Teams already get backed-up bringing robots in and out of the arena, setting up etc... and now they will be bringing in and removing their external compute devices?

In the long term, future... it would be best/easier if it were possible to just get a rack mount shelf, with a lock... and just leave them there for the week. Teams put them there, don't really touch them, and don't expect to use them/have much access to them as developer PCs while at the competition.

Or, perhaps get a 16 port KVM switch for each and league keyboard mouse, with all devices locked under the table or in a rack. But I would suspect teams would fight for this during non-scheduled setup/practice time.

warp1337:
bump @awesomebytes @airglow @justinhart

LoyVanBeek:
I agree with @moriarty here that a set up where there is just a server quietly spinning in a corner or rack all week would be ideal. 
If that is not possible, then indeed the rules as suggested. Maybe the preference for such 'quietly spinner servers in a corner' could be expressed.

warp1337:
Thx @LoyVanBeek for the review. Nevertheless, I dont like the idea if teams bringing a rack and having it connected to arena wifi the entrie week 24/7. Just my 2 cent.

LoyVanBeek:
Good point. 

I think there should be a centralized place/rack people can put their 1U server in, but then again not many people will have those. But indeed, using WiFi all the time could be problematic.

justinhart:
A 1U server could represent a significant investment over infrastructure that a lab already owns. It's true that teams are setting up computers, but for the most part, they're setting up systems that look like they've been ripped off of the desk in their lab. As such, it seems like we'd just be wedging out labs who already have gear.

Why not just say that all labs need to have their computers set up on the desk at the start of the round? I don't think that any of these computers are connecting via wi-fi. Just require them to hook up to the ethernet.

warp1337:
@justinhart That's basically what I intended to express when I wrote the subsection. Maybe the current version is too fuzzy? Can someone please comment on the actual PR?

airglow:
Short summary:
* We need to place a table close to the entrance but still visible from the inside of the arena.
* The table will be divided into two, let's say by red tape. One side current team, one side next team.
* Switches connected to the Arena Wifi. 
* We need someone who is observing this area regularly.

I would add an introducing sentence for this section, what the motivation is behind (Reduce the setup time, giving equal space for preparation of external hardware, keep the setup area clean). From the issue it's directly clear from the subsection not.

In general I see all these separate sections for each sub-league critical. Doesn't it also make sense to include it to the other Leagues? However this then may will affect the command stations that some teams are using for starting and inspecting during the tasks.

warp1337:
@airglow Thanks for the comment. I will add a short introduction/explanation

@LoyVanBeek @kyordhel @justinhart 

What do you think about applying these rules to every league? Maybe we start by only applying them to SSPL/DSPL this year and see how this works?

kyordhel:
I say: keep the rulebook as simple and plain as possible. We have a 100+ pages document nobody ever reads full. Rules need to be simpler.
I also say: for sake of simplicity, let's have the same rules for all 3 leagues unless strictly necessary.

general_rules/ExternalComputing.tex
justinhart: I'm not sure that I like the idea of having the compute devices moving while people are in line. That could really slow up the line going through. Think of how long it could potentially take to start an external device, start all of the software, and so forth. I'd rather teams leave their gear hooked up and simply not communicate with the device and/or start their software until they're about to compete, but I think that moving the physical devices around could be chaotic.

general_rules/ExternalComputing.tex
justinhart: Shouldn't the availability of software to other teams be after the competition? Otherwise, a team could potentially copy another team's software.

justinhart:
I agree with @kyordhel. If there are dozens of league-specific variants of the rules, it could become difficult to make fair decisions when teams come into conflict in their interpretations of those rules.

warp1337:
@kyordhel @justinhart Alright. Agree.

LoyVanBeek:
You have my blessing, for what it matters. @kyordhel your review still says "requested changes", is the current PR OK now?

kyordhel:
@LoyVanBeek I would need to check everything again this weekend. However, whether I approve changes or not is somehow irrelevant due to:
1. As @justinhart already pointed out, moving external devices can become a logistical nightmare and reduce team's performance during the test.
2. I'm the only one officially reviewing the PR while other TC members have comments but I'm not sure if they grant a go or a no-go.

So far, to me is OK (thanks for the blessing, Father), but I'm not clicking that approve button until I see another approval.

general_rules/ExternalComputing.tex
kyordhel: No.

To breach the gap between Cloud Computing and External Computing, software resources (not source code) must be made available to all. If you have an outstanding ASR I should be able to try it also as is in your servers. We want to promote sharing and knowledge exchange in the league.

Any comments @LoyVanBeek? @balkce? @swachsmu? @dirkholz?

general_rules/ExternalComputing.tex
kyordhel: Me neither, but having the devices of all 12 teams can also be problematic. Do you have in mind any alternative solution?

warp1337:
@kyordhel Thx again for commenting, but again, concerning 1. in your last comment ... **NOT** moving equipment is an even bigger nightmare. IMHO, having a structured approach that causes trouble in the first iteration is better than having no approach at all that will make all hell break lose.

kyordhel:
Don't get me wrong @warp1337, I agree that having the equipment there (outside a locked rack) might be more problematic than moving equipment competition-wise. My point was if the rest of the TC/OC agrees on transferring the nightmare to the logistic. **I'm not taking any responsibility regarding solutions for this issue**. This is solely TC to decide. My reasons:

1. As OC Chair, I barely have people to assist in the competition. Coordinate teams moving external devices is something the OC can't afford.
2. As referee and ex-TC I understand the problem. Cheaters gonna cheat and it looks terrible to have a bunch of hackers tweaking outside the team area.

In conclusion, I will follow the rules, but I'm not hanging myself with that rope.

warp1337:
@kyordhel Yes, I understand that, totally. So, what are the other OC/TC members saying? If required, we can also have a skype call on this to speed up the process. For me, this topic is extremely important, almost as much as the network topic ;)

awesomebytes:
Hello, excuse my lateness in the discussion. By reading the rules it makes me think the ideal setup for the external computer device is a beefy laptop. Our current equipment is an Alienware Tower, and I guess others will have something similar in their labs. We aren't in condition to change this hardware.

I mention it because what I understand from `all equipment must be removed from the ECRA` is that we'll need to move the tower in this case, which implies disconnecting it from power (and previously, not having it connected to power). If I'm misunderstanding, sorry!

I find this problematic as computers refusing to boot properly is not such a rare thing (specially with nvidia cards in them), which could affect the performance of teams. If there is a short time slot to setup the machine, and we need it to boot (and without a screen, as this rules say `stand-alone screens are not allowed`) we may not know something went wrong, or just take too long to notice. Also the startup time of whatever software we need to run affects this.

I would be fine with having enough power sockets to keep the machines plugged, and just disconnect the only ethernet cable available to connect to the arena wifi. Or something similar to that.

warp1337:
@awesomebytes Thank you for the reply, really! Finally, someone actually participating in the league replied, yeah!

Now we are at the _core_ of the issue that I want to address and we have evidence that I was not sooo wrong with opening this discussion. 

I understand, you have a high-end desktop machine, which is hard to setup and subsequently to remove on site, especially without a screen so on and so forth. Got it! 

I also get that you most probably won't buy a high end laptop instead, since you already have a machine for your robot. Fair enough.

In our case, we use a high-end laptop. To be precise, we are using this one: "Dell Inspiron 15 7000 Gaming-Notebook". So, no worries from our side...

Let's boil this down ... again: we will be > 10 Pepper teams at Montreal. If everyone brings their desktop machines, including a screen and a keyboard, we will ~probably~ end up in the same situation as we've seen in Nagoya --- total chaos at the ECRA and:

**1)** Local bandwidth will be quickly exhausted because everyone leaves their comp. machines running and streaming all the time.
    Even if you have the good will and plan on shutting it down when not needed, people will forget that when it's crunch time, trust me on that one.
**2)** There simply won't be enough space.
**3)** All referees will see from the ECRA is a 'wall of screens'

I ~expect~ hope people will make a decision at some point (soon-ish) regarding this PR/Issue. 

I totally see the point of bringing a high end desktop, simply more power. If there is no limitation in the rulebook, we will probably also switch to a desktop machine (thinking competition-wise). 

However, I really don't like the idea of making this a 'who has the most TitanX GPUs in their machine' battle -- especially because it comes with all the downsides I mentioned in 1) 2) and 3). Finally, why buying a ~15K robot if you need to additionally buy a 3-4K machine in order to run it (this is the SSPL, remember).

My 2 cent.

@LoyVanBeek @kyordhel  @moriarty @justinhart 

@LoyVanBeek  @kyordhel @moriarty @justinhart

kyordhel:
### Current rules
@awesomebytes please open an issue indicating where the current rules are ambiguous regarding the use of a laptop as only alternative for external computing. So fas there is no restriction on the type of external computers that can be used.

### Pull request
Replying to @awesomebytes and @warp1337. I'm OK with an hybrid approach. Consider the following:
- **Laptops:** To be set up one slot before the test and removed after the test.
- **Desktops and servers:** To be set up early in the morning and shut down at evening (we can-t be sure power won-t be cut). The rest of the time the equipment must remain chained to the table with no keyboard/mouse/monitor plugged in.
    - In case of bandwidth problems, policy is: cable is connected before the test and disconnected afterwards.

As referee I can say: If I see anyone nearby those servers during a test, that team is immediately disqualified in the most bad-ass possible way.

LoyVanBeek:
100% agree with the change that @kyordhel proposes. 

I feel the concern about "who can cram in the most GPUs in a box", but the same would apply when using a cloud service. This is a box of Pandora opened when external computing was allowed. Restricting on the amount of T/GFlops or something is  not easy either. Perhaps limit to 1 GPU, but that's all besides this issue.

kyordhel:
@LoyVanBeek I think no restriction shall be imposed in the external computing devices, specially since:

> _The RoboCup@Home tournaments are organized in independent test sets, which are used to benchmark the robots' abilities and performance in a realistic non-standardized home environment. _ [1]

So, as part of this _realism_, the arena network should be much like the one given by standard domestic internet provider. If your external devices require high-end infrastructure, you are doing something wrong and definitively is not @Home's problem.

Yet again, there is another rule specifying that robots must remain safely functional when no wireless is available.

[1] Wisspeintner, T., Van Der Zant, T., Iocchi, L. and Schiffer, S., 2009. RoboCup@ Home: Scientific competition and benchmarking for domestic service robots. Interaction Studies, 10(3), pp.392-42

warp1337:
>Laptops: To be set up one slot before the test and removed after the test.
> Desktops and servers: To be set up early in the morning and shut down at evening (we can-t be sure power won-t be cut). The rest of the time the equipment must remain chained to the table with no keyboard/mouse/monitor plugged in.
> In case of bandwidth problems, policy is: cable is connected before the test and disconnected afterwards.

Thanks. I would agree 100% to that proposal

moriarty:
I am OK with the hybrid approach from @kyordhel above.

> Now we are at the core of the issue that I want to address and we have evidence that I was not sooo wrong with opening this discussion.

> This is a box of Pandora opened when external computing was allowed.

Yes I remember reading the TDPs last year and seeing what they had listed as "External Computing" and was a little shocked... and then I went on a rant in #243.

> So, as part of this _realism_ ...

@kyordhel  was talking network bandwidth there, but I think in the future I agree with discussing some limits like @LoyVanBeek suggests for the External Compute, at least for _majority_ of the competition... Open Challenge you shouldn't be held back and can do whatever you want/can afford. But that's another issue for later.

kyordhel:
@warp1337 With "at least" we might have the situation in which all team have their equipment there since the beginning of the test. That worsens when tests are run in cycle, they might not remove them at all!

justinhart:
Is it really necessary that teams not fiddle with whatever server between rounds? I'm not sure what function it serves.

kyordhel:
@justinhart Basically to prevent last-minute tweaks and adjustments that in many cultures are not considered as cheating but might still grant a slight disadvantage to the competing team.

warp1337:
Like starting to stream the hell out of the their depth cam point clouds. As proposed in the network issue bandwidth limitation would mitigate that. 

I recently spoke to a colleague who has also long standing experience with the RoboCup ... He introduced a, at least for me, totally different point of view. I will elaborate on that later...

justinhart:
I see. Yes, that would be really uncool, but wouldn't they be able to do
this remotely?

kyordhel:
> I see. Yes, that would be really uncool, but wouldn't they be able to do this remotely?

Maybe but not easily. Let's remember the External Devices and the robot run on the Arena Network, one to which only authorized MAC Addresses can connect. This makes unlikely that any other computer can access the server while, at the same time, makes possible for both, robot and external device to run in the same ROS environment.

@warp1337 That's an important point. Although the Arena Network is separated from the testing network, let's not forget all wireless devices share the same medium. This means that any computer streaming 1080p@60fps pointclouds through any wireless network will negatively impact all wireless within 200m radius.

warp1337:
Hi everybody!

In the following I will try to summarize the current **general** situation and propose a completely new approach / introduce a completely new point of view. I will try to keep this post as short as possible.

My original idea behind the ECRA, as proposed above, was:

- limit the 'jumble' before the task, i.e., people fiddling with other peoples' cables and whatnot
- provide insight for refs of what is happening during the task wrt 'people typing in stuff on their laptops during their task' aka 'I am just inspecting log files' 
- limit overall network traffic by not having 'server-like' machines running and streaming **all** the time

I came up with these ideas after the RC in Nagoya, where we experienced an excellent network setup btw. To be honest, these ideas are based on my latest experience in the SSPL (after almost 8 years in the OPL). I recently had a good conversation with my colleague @semeyerz who lead team ToBI with me and Sven W. in 2016 (Leipzig). When I told him the current situation he was, mildly put, ... shocked.

Here is why, and this is also my current feeling about this. 

In the past, bringing a robot to the arena for a test, actually really meant bringing a **robot** --- not a robot, two laptops, one server and 12 team members setting all this up. It was just: bring your robot, press a start button, watch the robot _not_ enter the arena, repeat. Being part of the SSPL myself and having now worked with the Pepper for almost 1 year now, I 'get' that external computing is almost inevitable with that specific robot (prove me wrong!). 

Moreover, .... hand on heart ... since we allow external cloud computing --- _note:_ this connects the robot to the **internet** --- we cannot really control whether someone is remote controlling the robot or labeling data manually or not. Don't get me wrong please, I have trust in sportsmanship and fair play and I don't allege people do that in general. It's just, we simply **cannot control this anymore**, moving laptops in-between test doesn't help here.

So here's my new proposal:

- server-like machines ---> have to be setup in a designated area before the event, have to be removed after the event. There's no fiddling during the tests. In my personal opinion, as said before, I would ban server-like machines at all, but that's a) not up to me and b) a completely different story. IMHO, the same goes for external (internet) cloud services, btw.
- bring your robot ---> **two** team members are allowed in the area formerly known as the ECRA and **one** laptop-like device to set, e.g., the final initial pose for AMCL, calibrate ASR or whatnot. **No crowd** in the ECRA, no setting up entire workspaces, no nothing :)  Laptop has to be removed after the test.

There are a few 'dependencies' in order to make this work:

- fair play, respect sportsmanship, and keep an eye on your fellow team members.
- limit the bandwidth per team/vlan (a server may now stream as much data as it wants, it'll only
affect the bandwidth within the vlan)

By implementing this, we will solve the above mentioned problems without introducing complicated rules about setting up / tearing down stuff.

What do you think of all this?

general_rules/ExternalComputing.tex
justinhart: I'm not sure. The way this is stated really makes problems, though. There shouldn't be an external monitor, meaning that the external device either must be a laptop or permit someone to run SSH or a terminal emulator into the system. Plus you have boot times to think of. Consider how crazy it is when the ref is going, "<Team Name>, it's time to set up your robot and compete or forfeit your turn." With this setup, teams may still be booting up their hardware and wrestling with network problems during that time. I really think that we just need something like a long table, not inside the lane where the teams are going, and then a requirement that you unplug your network cable until it's your turn. This allows teams to get their computers going without much issue. Maybe we give them a max footprint for their hardware on the table. Everything must fit inside this square.

justinhart:
I was just commenting something similar in Github. Sorry for being a bit
behind the stick. I had some personal matters come up. I can agree to this.

warp1337:
Thx, @justinhart for the comment.

>  had some personal matters come up.

Hope you're fine?!

What about the others @airglow @awesomebytes @LoyVanBeek  @kyordhel

awesomebytes:
I agree with the last view/proposal of @warp1337.

Letting the server-like devices in the place and not touching them. Also unplugging Ethernet cables sounds like a good way to ensure the competing team has all the available bandwidth.

Having just some specific people to access a laptop for the initial  setup of a test and get hands off it sounds reasonable too.

:+1:

warp1337:
I feel like I need to clarify one further detail. My proposal does not solve the bandwidth issue. It's just a simple and clear proposal about the procedure concerning the tests. However, the bandwidth 'situation' can only be solved by the LOC, i.e., limiting the available bandwidth per team (vlan) based on the infrastructure's capabilities --- put simply: total available wifi bandwidth is 100 Mbps and number of teams is 10 ---> 100/10 per vlan.

kyordhel:
I would like to request only three things:
- Servers must be chain-locked to the table (anti-theft).
- No people around them. Ever.
  - No plugging/unplugging cables (accidental sabotage?).
  - No bluetooth mouses, keyboards, tablets, etc. nrearby.
- SPL team members volunteer as cops
  - You see anything suspicious, you report.
  - No warnings, immediate disqualification.

warp1337:
If more than one members of the committee agree to this, I am willing to rework my PR. Thx @kyordhel

kyordhel:
@warp1337 

You have @justinhart and @awesomebytes explicit approval, as well as the tacit ones from the rest of the TC. Please proceed to make the changes in the PR.

P.S. My requests are on the operative level. Like the document distributed about Network configuration, detailed information shouldn't be in the rulebook, but standalone as specific regulations for this particular event.

warp1337:
Okay. Will do. Will take some hours, thank you.

warp1337:
I updated the PR, please review (sorry took more than a few hours...)

@LoyVanBeek @kyordhel @justinhart

warp1337:
Thank you! Finally, we could all agree :)
I believe it is important to _notify_ the people about this addition. How can we do this?

kyordhel:
@warp1337 Several changes (winter) are coming and must be committed/merged before Sunday. Then we will have an official draft which will be sent to all teams by mail (OC do this).

PR #390, opened 2018-02-13
Add a changelog

LoyVanBeek:
This should be the entrypoint of current teams to start with each year

Why haven't we introduced this years ago?

### DIFF
--- a/Changelog.tex
+++ b/Changelog.tex
@@ -4,8 +4,38 @@ Changes such as spelling and grammar fixes are not included.
 \subsection*{Changes in 2018}
 
 \begin{itemize}
-	\item Introduction of changelog
-	\item \textbf{EE-GPSR} Simplify EE-GPSR: less categories, simpler scoring. Full detail in \href{https://github.com/RoboCupAtHome/RuleBook/pull/327}{PR 327}
-	\item \textbf{VizBox} Introduce VizBox to visualize challenge with \href{https://github.com/RoboCupAtHome/RuleBook/pull/367}{PR 367}
-	\item \textbf{DSPL Standard laptop} Mandatory use of official Standard Laptop for DSPL \href{https://github.com/RoboCupAtHome/RuleBook/pull/374}{PR 374}
+	\item \textbf{General rules}
+		\begin{itemize}
+			\item Introduction of changelog
+			\item \textbf{VizBox} Introduce VizBox to visualize challenge with \href{https://github.com/RoboCupAtHome/RuleBook/pull/367}{PR 367}
+			\item \textbf{DSPL Standard laptop} Mandatory use of official Standard Laptop for DSPL \href{https://github.com/RoboCupAtHome/RuleBook/pull/374}{PR 374}
+			\item \textbf{External Devices: ECRA Rule}
+				\begin{itemize}
+					\item No people around external devices
+					\item Mobile devices must be removed, others chained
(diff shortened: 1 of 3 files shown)


Changelog.tex
```diff
@@ -0,0 +1,11 @@
...
+The changelog lists the significant changes in the rules since the previous global RoboCup event. 
+Changes such as spelling and grammar fixes are not included. 
+\subsection*{Changes in 2018}
+
+\begin{itemize}
+	\item Introduction of changelog
+	\item \textbf{EE-GPSR} Simplify EE-GPSR: less categories, simpler scoring. Full detail in \href{https://github.com/RoboCupAtHome/RuleBook/pull/327}{PR 327}
+	\item \textbf{VizBox} Introduce VizBox to visualize challenge with \href{https://github.com/RoboCupAtHome/RuleBook/pull/367}{PR 367}
+	\item \textbf{DSPL Standard laptop} Mandatory use of official Standard Laptop for DSPL \href{https://github.com/RoboCupAtHome/RuleBook/pull/374}{PR 374}
+\end{itemize}
```
kyordhel: Missing section for the list of buffed tests. I've in my ToDo list:
- Change Storing Groceries
  - Only moved objects score
  - PDF report for validation purposes
- Change Set the table & tidy up
  - Remove set the table
  - Tidy up part is a startup for the dishwasher test
  - Required by RoboCup federation as part as the P&G agreement

airglow:
I'm totally supporting this. If the commit messages are useful this could auto generated. I'm also fine with just putting it on top of the README.md

LoyVanBeek:
The commit messages aren't that useful, the overarching PR's are though. 
I don't want t include like stuff typo fixes, that would be hard to auto-exclude.

LoyVanBeek:
@kyordhel I'll extend the changelog when the changes are merged. Ideally, updating the changelog is part of the PR. 

I'd like to have this in the actual rulebook itself, so the rulebook itself is the single place to go for existing teams as well. 
Currently it's quite early in the rulebook, perhaps it should be an appendix. Having in online only is not enough I think.

kyordhel:
@LoyVanBeek Appendix? Maybe... I still don't like the idea, but if most people agree, is OK.

kyordhel:
@LoyVanBeek the PR have been merged, so I think we should update this section and merge.
Can we do this today? Here the changelog:

### Finals
- Bronze Finals: 4th and 3rd match up for the 3rd place trophy

### Stage 2
- Open Challenge:
    - Score normalized to Stage 2 highest one. 
    - Evaluation metrics updated.
- Set the Table: Replaced with P&G Challenge
- P&G Challenge: Clean up a table by placing tableware in dishwasher.
- EEGPSR: Reduced number of categories from 8 to 3
    - Three at once, object-related, and people-related
    - One of each, normal, incomplete and erroneous commands (given randomly)

### Stage 1
- Storing Groceries:
    - Per moved object scoring
    - Score for opening door increased
    - PDF report optional
- Help me carry: Unless crushing, pushing the small object won-t finish the test

### General Rules
- DSPL backpack
    - DSPL teams can use official laptop in provided backpack
- External Devices: ECRA Rule
    - No people around external devices
    - Mobile must be removed, others chained
    - No keyboards, mouses, or screens attached
- Removed bonus for data recording 
- Teams encouraged to provide annotated recording of spoken interactions

PR #392, opened 2018-03-05
Rework and balance Storing Groceries Test

kyordhel:
The test as described promoted no manipulation since most points were granted for object recognition and not for solving a task (#353). To solve this issue, now points are granted per moved object and the PDF report is only for validation purposes. In addition, the number of points granted per object increases with every move, promoting speed. The description of the task was changed accordingly. Scoring system redesigned focusing on actually solving the task and addressing #358.

Changes:
- Setup:
  - Test can take place inside or outside the arena (explicit)
  - Buffed other setup items and removed redundant references
- Task:
  - Stacking objects is allowed (*new*).
  - Finding the table, inspecting the cupboard, and opening the door can be done in any order.
- Additional Rules and remarks
  - Removed PDF report.
  - Added timing section.
- Moved PDF report instructions to its own section.
  - Report only considered if the robot moved at least on object.
  - Removed final state (same as all no initial states).
- PDF report is OPTIONAL.
- Scoresheet
  - Simplified timing
  - Reworked
  - Incremental scoring per object placed

### DIFF
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -4,7 +4,7 @@ The robot must place the first object within the first 2 minutes (+1 minute if t
 
 \begin{scorelist}
 	\scoreheading{Opening the door}
-	\scoreitem{20}{Autonomouslu opening the door}
+	\scoreitem{20}{Autonomously opening the door}
 	% Total: 20
 
 	\scoreheading{Arranging objects}
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -1,44 +1,30 @@
 
-The maximum time for this test is 1+2+2 minutes.
-The first minute is for the robot to open the Cupboard's door on its own.
-If the robot is not able to open the door within that minute, it will be opened by the referee. 
-In case the robot opens the door within the minute, the robot has a small time advantage. 
-Then, the robot has 2 minutes to recognize objects and manipulate items.
-If an item is manipulated within that time, 2 more minutes are added to the test time. 
-If not, the test ends.
+The maximum time for this test is 5 minutes.
+The robot must place the first object within the first 2 minutes (+1 minute if the robot opens the door).
 
 \begin{scorelist}
 	\scoreheading{Opening the door}
-	\scoreitem[1]{20}{Opening the door without human help}
+	\scoreitem{20}{Autonomouslu opening the door}
 	% Total: 20
-	\scoreheading{Moving objects}
-	\scoreitem[5]{10}{Successfully grasping an object (5 cm for more than 10 seconds)}
-	% Total: 50
 
-	\scoreitem[5]{10}{Successfully placing an object near to another of the same class}
-	\scoreitem[5]{ 5}{Successfully placing an object anywhere else}
-	% Total: 50
+	\scoreheading{Arranging objects}
+	\scoreitem{20}{Successfully placing the \textbf{1st} object next to another of its same class}
+	\scoreitem{30}{Successfully placing a \textbf{2nd} object next to another of its same class}
+	\scoreitem{40}{Successfully placing a \textbf{3rd} object next to another of its same class}
+	\scoreitem{50}{Successfully placing a \textbf{4th} object next to another of its same class}
+	\scoreitem{60}{Successfully placing a \textbf{5th} object next to another of its same class}
+	% Total 150
 
-	\scoreheading{Recognizing known or alike objects}
-	\scoreitem[10]{5}{\small Every correctly recognized known or alike object in the report file} 
-	% Total: 50
-	
-	\scoreheading{Recognizing unknown objects}
-	\scoreitem[10]{5}{Correctly label unknown object as \quotes{Unknown}}
-	\scoreitem[5]{20}{Label a \textit{pair} of unknown objects with the same label (e.g label both papayas both as \quotes{class A})}
-	\scoreitem[10]{15}{\footnotesize Label unknown object as correct category (e.g. label a papaya as fruit)}%
-	\scoreitem[10]{15}{\footnotesize Label unknown object as correct class (e.g. label a papaya as papaya)}%
-	% Total: 150
-	
-	\scoreheading{Incorrect recognitions}
-	\scoreitem[10]{-5}{False positive label}
-	
-% 	\scoreheading{Total task}
-% 	\scoreitem[5]{40}{Place known object near known object of same class}
-% 	\scoreitem[5]{50}{Place unknown object near unknown object of same class}
+	\scoreheading{Wrong placements}
+	\scoreitem[5]{10}{Successfully placing an object on the cupboard}
+	% Total 0
 
-	
-	\setTotalScore{320}
+	% \scoreheading{Other placements}
+	% \scoreitem[5]{50}{Each correctly arranged object after the 5th one}
+	% Total 250
+
+
+	\setTotalScore{220}
 \end{scorelist}
 
 
--- a/tests/StoringGroceries.tex
+++ b/tests/StoringGroceries.tex
@@ -15,14 +15,14 @@ This test focuses on the detection and recognition of objects and their features
 \begin{minipage}{0.70\textwidth}
 	\subsection{Setup}
 	\begin{enumerate}
-		\item \textbf{Location:} This test can take place either inside or outside the arena. The testing area must have a bookcase or cupboard, and a newarby table. The maximum distance between the Table and the Cupboard is 2 meters.
+		\item \textbf{Location:} This test can take place either inside or outside the arena. The testing area must have a bookcase or cupboard, and a nearby table. The maximum distance between the Table and the Cupboard is 2 meters.
 		\item \textbf{Start position:} The robot starts between the cupboard and the table in a random orientation, but facing towards the Cupboard.
-		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.30m and 1.80m from the ground and contains several objects grouped by category or likeliness (See \ref{rule:scenario_objects}). The cupboard has at least one free space for starting a new set.
+		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.0m and 1.80m from the ground and contains several objects grouped by category or likeliness (See \ref{rule:scenario_objects}). The cupboard has at least one free space for starting a new set.
 		\begin{itemize}
 		 	\item \textbf{Door:} The cupboard has a single door, which is closed initially.
 		 	This door encloses some of the objects, covering up to one half of the cupboard (e.g. the left or bottom half), as indicated by the hatched area in Figure \ref{fig:storing_groceries_shelf}.
 		\end{itemize}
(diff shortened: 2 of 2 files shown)


kyordhel:
ToDo:
- Object distribution (on table)
  - 3 Known objects
  - 3 Alike objects
  - 2 Unknown objects
  - 2 Special objects (cloth, tray, container)
- Object recognition report template
  - Refer address of repository with template
- Add handle specification (if any).

airglow:
I would change the timing. I guess when I would evaluate the task during the competition it's complicated to always get this 1+2+2 min correct. I would suggest a 5 min time limit, and if the robot doesn't do anything in 30 sec or a minute it's over. Much simpler and we motivate teams to do manipulation. Even if they take 1.10 min for opening the door I wouldn't like to stop the task when they start sorting the stuff from the table to the shelf.

scoresheets/StoringGroceries.tex
```diff
@@ -1,44 +1,30 @@
...
+	% Total 150
+
+	\scoreheading{Wrong placements}
+	\scoreitem[5]{10}{Successfully placing an object on the cupboard}
+	% Total 0
+
+	% \scoreheading{Other placements}
+	% \scoreitem[5]{50}{Each correctly arranged object after the 5th one}
+	% Total 250
+
```
airglow: The score sheet is now much simpler! I like it that way. Only the timing should also simplified. Should we really care when a robot is manipulating slower. I mean this already reflects in the scoring. Faster manipulation -> higher score. If the robot is stuck we should abort after 30sec or a minute.

tests/StoringGroceries.tex
```diff
@@ -7,18 +7,22 @@ \subsection{Goal}
...
 \begin{minipage}{0.70\textwidth}
 	\subsection{Setup}
 	\begin{enumerate}
-		\item \textbf{Location:} One of the bookcases or cupboards in the apartment is used for this test, one where a table is near or can be put. 
-		\item \textbf{Start position:} The robot will start between the cupboard and the table in a random orientation, but facing towards the Cupboard.
-		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.30m and 1.80m from the ground and contains several objects (See \ref{rule:scenario_objects}).
+		\item \textbf{Location:} This test can take place either inside or outside the arena. The testing area must have a bookcase or cupboard, and a newarby table. The maximum distance between the Table and the Cupboard is 2 meters.
+		\item \textbf{Start position:} The robot starts between the cupboard and the table in a random orientation, but facing towards the Cupboard.
+		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.30m and 1.80m from the ground and contains several objects grouped by category or likeliness (See \ref{rule:scenario_objects}). The cupboard has at least one free space for starting a new set.
```
airglow: is this 30cm limit relevant? This indirectly excludes the lowest level of most cupboards. Can we always ensure to have 5 shelves?

tests/StoringGroceries.tex
```diff
@@ -7,18 +7,22 @@ \subsection{Goal}
...
 		\begin{itemize}
 		 	\item \textbf{Door:} The cupboard has a single door, which is closed initially.
 		 	This door encloses some of the objects, covering up to one half of the cupboard (e.g. the left or bottom half), as indicated by the hatched area in Figure \ref{fig:storing_groceries_shelf}.
-		\end{itemize} 
-		\item \textbf{Table:} A table near to the Cupboard has 10 objects (See \ref{rule:scenario_objects}). If not all objects fit on the table, they will be added during the test. The maximum distance between the Table and the Cupboard is 2m.
-		\item \textbf{Objects:} Objects in the Cupboard and on the Table can be known, alike, or unknown. Also, there will be more than one object on each shelf. 
+		\end{itemize}
+		\item \textbf{Table:} The table may have up to 10 objects. If not all objects fit on the table, they will be added as the robot frees up space.
```
airglow: May change to at least 5, cause only 5 objects are awarded for storing?

tests/StoringGroceries.tex
```diff
@@ -32,55 +36,83 @@ \subsection{Focus}
...
+		\item If the Cupboard has no similar object, the robot must clearly state its decision on how to solve the problem. For instance, the robot can start a new set in a free space for either all unknown objects or all objects sharing a particular feature (color, shape, function, etc.).
+		\item Moving two objects at a time (2-handed manipulation) is allowed.
 	\end{itemize}
 
-	\textbf{Note:} Either before or after grasping an object the robot may announce the name of the object found. 
+	\textbf{Note:} Either before or after grasping an object the robot may announce the name of the object found.
 	\item \textbf{Repeat:} This repeats until the time is up or all groceries are stored.
 \end{enumerate}
 
+
```
airglow: Is this PDF stuff still relevant?

tests/StoringGroceries.tex
kyordhel: The 30 cm is relevant because I think is outside the specs of the HSR (Rules are cross-compatible with DSPL and OPL). Before DSPL it was 0-180m. 

Regarding the 5 shelves, so far there has been no problem.

tests/StoringGroceries.tex
kyordhel: There is a commented line over there for 6+ objects in the scoresheet. But let's be realistic, no robot can more more than 1 object per minute.

Why not _at least 5_ because it might be that the table can't hold all 5 at once. Indeed we can reduce the number of objects to max 5 **but** you want to be sure that robots can recognize as many objects as possible in real time.

tests/StoringGroceries.tex
kyordhel: I would remove it but we should keep the number of changes to a minimum. Plus, it might be useful for clarification if some teams comes up with some weird way of grouping objects (unlikely) e.g. by color, shape, etc. So in case the team comes arguing that the robot grouped all cans just because of a coincidence, you need something to check against.

kyordhel:
@airglow Thanks for the review. I agree in most of your changes (specially the time constraint), but the idea was to keep changes to a minimum. Let's see what other TC members have to say before applying such changes. Of special concern are the upper and lower limits of the shelf due to HSR limitations.

scoresheets/StoringGroceries.tex
```diff
@@ -1,44 +1,30 @@
...
-Then, the robot has 2 minutes to recognize objects and manipulate items.
-If an item is manipulated within that time, 2 more minutes are added to the test time. 
-If not, the test ends.
+The maximum time for this test is 5 minutes.
+The robot must place the first object within the first 2 minutes (+1 minute if the robot opens the door).
 
 \begin{scorelist}
 	\scoreheading{Opening the door}
-	\scoreitem[1]{20}{Opening the door without human help}
+	\scoreitem{20}{Autonomouslu opening the door}
```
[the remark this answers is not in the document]
kyordhel: Thanks! This change definitively goes in!

tests/StoringGroceries.tex
```diff
@@ -7,18 +7,22 @@ \subsection{Goal}
...
+% Setup
+%
+% %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \begin{minipage}{0.70\textwidth}
 	\subsection{Setup}
 	\begin{enumerate}
-		\item \textbf{Location:} One of the bookcases or cupboards in the apartment is used for this test, one where a table is near or can be put. 
-		\item \textbf{Start position:} The robot will start between the cupboard and the table in a random orientation, but facing towards the Cupboard.
-		\item \textbf{Cupboard:} The cupboard has 5 shelves between 0.30m and 1.80m from the ground and contains several objects (See \ref{rule:scenario_objects}).
+		\item \textbf{Location:} This test can take place either inside or outside the arena. The testing area must have a bookcase or cupboard, and a newarby table. The maximum distance between the Table and the Cupboard is 2 meters.
```
LoyVanBeek: Typo: "newarby" --> "mearby"

tests/StoringGroceries.tex
LoyVanBeek: HSR can lift from the ground, with gripper and vacuum cup. It has a camera in the gripper so it should be even more capable than the average OPL in that regard.

tests/StoringGroceries.tex
LoyVanBeek: 10 is more cluttered (interesting) but robots could select, out of the 10, the items they can manipulate etc. 
But indeed 'at least 5' would be OK

tests/StoringGroceries.tex
```diff
@@ -32,55 +36,83 @@ \subsection{Focus}
...
-	\item \textbf{Moving objects:} The robot chooses which object to move first from the Table to the Cupboard, allocating similar objects all together.
+	\item \textbf{Evaluating the situation:} The robot inspects its surrounding and analyzing the best course of action. In any order, the robot has to:
+	\begin{itemize}
+		\item \textit{Inspect the cupboard} (locating and categorizing existing groceries).
+		\item \textit{Open the cupboard's door.} If the robot can't open the door, it may ask the Referee to do it.
+		\item \textit{Find the table}
+		\item \textit{Inspect the table} (analyze the newly bought groceries, i.e. objects).
+	\end{itemize}
+
+	\item \textbf{Moving objects:} The moves as many objects as possible from the Table to the Cupboard (max 5), allocating similar objects all together.
```
LoyVanBeek: First "as many as possible' and then "max 5", that's a contradiction

tests/StoringGroceries.tex
LoyVanBeek: In the stress of a test, checking correct classes etc is quite busy, PDF reports are easier in that regard.

tests/StoringGroceries.tex
kyordhel: Thanks. That change definitively will go in

tests/StoringGroceries.tex
kyordhel: You and @airglow suggested that change, so goes in.

tests/StoringGroceries.tex
kyordhel: Done!

tests/StoringGroceries.tex
kyordhel: Done!

tests/StoringGroceries.tex
kyordhel: I didn't change anything here. I think @LoyVanBeek has a point, so PDF is now **optional**. Teams will score per object moved **and correctly classified**. However, robot's reasoning might not seem straightforward to us and no NLP engine will allow us to ask the robot why did it sorted out groceries as it did. Therefore, checking via the PDF report would be really useful. I think most teams will generate it.

tests/StoringGroceries.tex
kyordhel: Fixed. Now is as many as possible within the given time but only first 5 scores. 
In any case, no robot so far is that fast. But assuming we get The Robo-Flash, additional objects would go for Outstanding Performance and to please the audience.

kyordhel:
@airglow @balkce @justinhart Could any of you approve the merge?

PR #403, opened 2018-03-23
Added optional audio data recollectionfor GPSR, EEGPSR and SPR.

balkce:
Thanks to @kyordhel for removing the recollecting data from the whole rule book. However, as discussed in the meeting, it is of interest to recollect audio data for speech recognition. Since GPSR, EEGPSR, and SPR are the most speech-recognition-based tests, we are encouraging teams to submit their annotated data to the TC. No mentions of points is given, just "encouragement". It's very probable that teams won't do it, but it's a possibility.

### DIFF
--- a/tests/EEGPSR.tex
+++ b/tests/EEGPSR.tex
@@ -116,6 +116,8 @@ This test particularly focuses on the following aspects:
 	\item Take the command and total time per team.
 \end{itemize}
 
+\subsection {Audio Data Recollection}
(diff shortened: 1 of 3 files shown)


kyordhel:
I approve. Maybe we should specify that the audio files must be in PCM Wave 22050Hz, or 44.1kHz, or 48kHz, 16 bit either stereo or mono.

Maybe we should add a certificate for those who provide annotated files.

Issue #362, opened 2017-08-22
Standard door handle


kyordhel:
This year robot failed big time at opening door, most (if not all) didn't even try. Most recurrent complain from part of the teams is that "it was impossible to know beforehand how the door would look like or how the handle will be, with barely any time during the competition for train and test".

An intermediate milestone is then suggested. Robots are required to open the door but an official handle is described in the rulebook. Since there is no company distributing such handles worldwide (i.e. no standard), we should also provide the schematics (based on LOC selection) of the handle in case teams want to build or 3D print one for training.

Example: IKEA HACKÅS or IKEA MÖLLARP
![image](https://user-images.githubusercontent.com/8583949/29568268-1cd992b6-8750-11e7-95b5-27f7f7ca37b5.png)

maximest-pierre:
We (me and the rest of the LOC) are going to have a look at it and come with a response as soon as possible. Do you have a specific deadline? I believe one from ikea would be the best solution.

kyordhel:
@maximest-pierre just the guy I was looking for.
One big favour. Could you please send me via email (oc[at]robocupathome.org) the full name of all LOC for Montreal 2018, including their email address, and affiliation? This is to keep in touch, add you to the mailing list of the OC, and add your names to the rulebook and official website.

lmmartine:
I am agreeing to give a model of the handle, but I am going for a model it more realistic.
Example : https://www.thingiverse.com/thing:1504403
 
![image](https://user-images.githubusercontent.com/4855655/30161768-d8cbde18-93c9-11e7-8c06-6dd1bdfef0f4.png)

kyordhel:
@maximest-pierre Do we have standardized handles or not?

maximest-pierre:
No, we are still waiting on some of our suppliers. I will poke them later today. If they don't answer, I am going to Ikea and I am choosing the first one that I see that fits.

kyordhel:
@maximest-pierre is a little bit more complicated than that. We must provide schematics for the replication of the handle to be fair with those teams that might not have a distributor nearby. We have qualified teams from countries in which there are no IKEAs yet.

If we can't get schematics, I'm afraid it would be a no-go for all teams (unless one team volunteers itself to provide a schematic or model for 3D printing).

Specifications must be:
- Dimensions
- Materials
- Colors

kyordhel:
It seems to me this won't happen in 2018
Clossing issue now.

PR #405, opened 2018-03-23
Storing Groceries: Opening door score 50pts

kyordhel:
Increase the number of points for door opening from 20 to 50 in Storing Groceries

So far no team is trying to open the door. Therefore, increasing scoring for it may sense.

We must keep in mind that **no more than 20 points** should be granted if we provide an **official door handle** (see #362).


balkce:
I thought we settled on 35 points.

kyordhel:
That comes in the next patch

PR #435, opened 2018-05-10
Roughly predefine table location & type

### LINKED ISSUE #419: Predefine table position in Storing Groceries
In the current Storing Groceries task, the position of the table is defined as only "within 2 meters of the starting position". 

Given the focus of this task (manipulation and object **recognition**),  I think the table position is too vague. (The focus is not to find furniture)

In the Stage 2 Restaurant task, the table position is even more defined (left or right). 
Given the already limited time in this challenge, I'd say: predefine the table position to left or right of the robot. 
Maybe even go as far as to define which table will be used.

LoyVanBeek:
Fix #419

### DIFF
--- a/tests/StoringGroceries.tex
+++ b/tests/StoringGroceries.tex
@@ -47,7 +47,6 @@ This test focuses on the detection and recognition of objects and their features
 	\begin{itemize}
 		\item \textit{Inspect the cupboard} (locating and categorizing existing groceries).
 		\item \textit{Open the cupboard's door.} If the robot can't open the door, it may ask the Referee to do it.
-		\item \textit{Find the table} Only a rough location is defined beforehand
 		\item \textit{Inspect the table} (analyze the newly bought groceries, i.e. objects).
 	\end{itemize}
 
@@ -114,6 +113,7 @@ A significant amount of objects can be unknown to the robot (See \ref{rule:scena
 	3 alike objects,
 	2 unknown objects, and
 	2 special objects (bowl, cloth, dish, etc.).
+	\item \textbf{Table} The table's rough location will be announced beforehand, having it's position either left, right, or behind the robot.
 	% \item \textbf{Timing:} The robot has to successfully place the first object within the first two minutes, otherwise the test is ended. If the robot opens the cupboard door by on its own, one additional minute is added to the 2-minutes limit. The maximum time for this test is 5 minutes.
 \end{enumerate}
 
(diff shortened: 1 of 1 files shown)


tests/StoringGroceries.tex
```diff
@@ -129,6 +129,10 @@ \subsection{OC instructions}
 \textbf{2 hours before the test}
 \begin{itemize}
     \item Announce the startup location for robots.
+    % This test is about manipulation and object recognition, NOT about finding furniture
+    % Finding a table is done in the Restaurant task
+    \item Announce which table will be used in the test
+    \item Announce a rough location for the table, i.e. left or right of the robot's startup location
 \end{itemize}
 
 \subsection{Referee instructions}
```
kyordhel: I would add here that the referee can move the table right before the test to make sure the robot hasn't to move much.
- Adjust the table position to the robot's working space.

tests/StoringGroceries.tex
```diff
@@ -47,7 +47,7 @@ \subsection{Task}
 	\begin{itemize}
 		\item \textit{Inspect the cupboard} (locating and categorizing existing groceries).
 		\item \textit{Open the cupboard's door.} If the robot can't open the door, it may ask the Referee to do it.
-		\item \textit{Find the table}
+		\item \textit{Find the table} Only a rough location is defined beforehand
```
kyordhel: This is unnecessary. Instead let's add **Table** under Additional Rules and Remarks.

_The table location will be roughly announced beforehand, having it's position either left, right, or behind the robot. The distance between the cupboard and the table will be adjusted to prevent robot from displacing (i.e. only turning)._

tests/StoringGroceries.tex
```diff
@@ -129,6 +129,10 @@ \subsection{OC instructions}
 \textbf{2 hours before the test}
 \begin{itemize}
     \item Announce the startup location for robots.
+    % This test is about manipulation and object recognition, NOT about finding furniture
+    % Finding a table is done in the Restaurant task
+    \item Announce which table will be used in the test
+    \item Announce a rough location for the table, i.e. left or right of the robot's startup location
```
kyordhel: Would change it to:
Announce where the table will be placed (e.g. left or right).

tests/StoringGroceries.tex
LoyVanBeek: Adjusting the table position is also not needed I think. If the arena puts a table close enough to a closest in the default arena state, I'd definitely would like to use that table.

Issue #431, opened 2018-05-03
[General Rules] Change score rectification procedure


johaq:
Section 2.1 states:

> Once  the  test  scoresheet  has  been  signed  or  the  scores  has  been  published,  the  TC
decision is irrevocable.

This section should be removed since it is not enforced. 
Alternatively we should talk about the procedure of signing and publishing. Fabrice told me in other leagues (work I think) at a test every team leader is present and has to sign off on the score. That is of course a lot of effort but can effectively prevent a lot of discussion. Personally I do not think this is feasible for @home but maybe we should discuss it.

airglow:
I'm strongly against removing this section. At German Open it was a trade-off between publishing preliminary results right after the tests, which I also think is very important. I would stick to the current procedure but before handing them discuss the scoring between TC's and EC's. This was a strong error that we made at German Open, which should not happen again. 
If you really want to complain about that now we can revert the score to the state before the correction which does not change the end result at all.

johaq:
I am also for keeping the section. But I think we should discuss the procedure.

**My proposal**: Invite the leader or a representative of every team to be present at every test (of their own team as well as others). Any issue with how the test is conducted or scored has to be addressed immediately. Discussions later in the day or even the day after are not allowed.

I think this would guarantee that each team is more involved in the rule book and avoid endless discussions afterwards. It also promotes exchange between the teams.

kyordhel:
First, I would like to clarify that scores in RoboCup@Home are NEVER
published untill al scoresheets are signed.

Removing this section is a NO GO from the OC point of view. Morevover, I
invite you people to join OC to experience chasing Team Leaders while at
the same time handle the presure of all teams + venue requesting for scores.
Current procedure works so I invoke the Rule #1 of engineering: does it
work? Yes? Then don't change it.

johaq:
I do not think that the current procedure works very well. You said it yourself: chasing team leaders and dealing with other teams is a nightmare. Hence my proposal which I think would fix or at least alleviate these issues.

kyordhel:
Maybe, but so far when teams want the scores, TL goes th toe OC table to ask and signs all pending scoresheets.

I've no idea what happened in German Open, but in the international we have had problems with unhappy teams a couple of times and is tiring and time consuming, yet solvable. What you propose will lead to several teams complaining about tests that are already gone from referees' mind prior to advancing to Stage 2/Finals. That can potentially break up all logistics.

johaq:
[...]

PR #438, opened 2018-05-11
Ammendment to score rectification procedure

johaq:
This addresses a proposal made in #431

### DIFF
--- a/general_rules/Procedure.tex
+++ b/general_rules/Procedure.tex
@@ -208,6 +208,17 @@ Hand gestures may be used to control the robot in the following way:
 	\end{enumerate}
 \end{enumerate}
 
+\subsection{Signing and Protesting of Scores}
+\label{rule:signing}
+\begin{enumerate}
+	\item Immediately after a test one member of the team (ideally the one already inside the arena) has to either confirm the scoring by the referees or protest a particular aspect about the scoring or how the test was conducted.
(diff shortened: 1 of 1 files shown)


balkce:
This has always been a point of contention between OC/TC/EC members, since some believe the committees have the last say on scoring (thus, no protests/challenges should be allowed) while others tend the other side of that spectrum and provide score review at any point of the competition.

The additions provided here seem adequate and more or less compatible to what we already do in practice: teams can either sign off on their score or can challenge the score via video/audio recorded evidence. What needs to be decided upon is the "when". My take is that I don't want teams to be challenging scores during a test (the judges are busy scoring other teams). Thus, I would like to make it clear that a challenge will be reviewed only after all teams have finished carrying out the test and, like it already states, before the next test starts.

In addition, I suggest to change each use of "protest" to "challenge", since the former seems more linked to screaming.

semeyerz:
I would suggest that the teams protest right after the test. This is then the only point of time where you are allowed to declare protest. But only declaration is done right after the test and written down.

When all tests are done the referees discuss the declared protests. But no new protests can be made at his point of time.

kyordhel:
*Closing reason:* Not approved (no TC reviews) before Final version deadline.

This is not a reject. Please reopen PR merging over 2019/master

PR #439, opened 2018-05-12
Remove PDF report from Storing Groceries as it is irrelevant

### LINKED ISSUE #424: PDF Recognition Report for storing groceries not needed anymore
Section 5.5.5.  PDF Recognition Report can be removed as it does not award any points.

semeyerz:
fixes #424

### DIFF
--- a/tests/StoringGroceries.tex
+++ b/tests/StoringGroceries.tex
@@ -67,34 +67,6 @@ This test focuses on the detection and recognition of objects and their features
 \end{enumerate}
 
 
-% %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-%
-% PDF Recognition Report
-%
-% %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\subsection{PDF Recognition Report}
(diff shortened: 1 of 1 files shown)


kyordhel:
I'm against removing the PDF reports since it can be used to explain robot's behavior to TC even though it's information is not directly scored. It could have been removed when the test was restructured but TC decided to keep it *just in case*.

I would require the review and approval of at least 2 TC members before merging.

semeyerz:
In my opinion the rulebook is already long enough and everything that is irrelevant should be removed asap.

And this test is pretty easy to score.
Placed object close to category. Done.

I do not understand where this report should be useful.
And what advantage comes from describing it in such detail?
Teams can provide what ever evidence they like (videos, logs, recordings, reports), ...
But why specify this only in this test?

kyordhel:
Because this test was scored using a PDF report before (minor changes only). As me myself, I hate PDF reports and I would love to shrink it. However, as Referee I must stick to what was agreed and wait for my peer's decision.

LoyVanBeek:
Based on my experience, barely any team ever does something optional, so no need for the report to be described in the test. 
Would it make sense as an appendix?

On a side-note:
The PDF makes refereeing a bit more relaxed in my opinion, as you have a bit less stress to keep track of what the robot was doing in a test but can review it later. Not sure if the added stress of gathering reports cancels that out, to be honest.

PR #442, opened 2018-05-12
Allow external computation

### LINKED ISSUE #421: Situation on external computing unclear
3.6.  External computing states:

Availability:
The resources must be publicly available, for use by robots of other teams,
well before and after the competition.

Which contradicts the fact, that some teams run computation on ECDs that they do not make publicly available.

The Remark of the section repeats the statement that it has to be available for other teams.

semeyerz:
fixes #421

### DIFF
--- a/general_rules/ExternalComputing.tex
+++ b/general_rules/ExternalComputing.tex
@@ -1,14 +1,10 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\section{External computing}\label{rule:robot_external_computing}
-Robots are allowed to use some form of external computing, for example in the form of so-called \quotes{Cloud services} and/or \quotes{Internet API's} etc. 
+\section{External Computing}\label{rule:robot_external_computing}
+Robots are allowed to use external computing. 
 \begin{enumerate}
 	\item \textbf{Definition:} Computing resources that are not physical part of the robot are \iterm{external computing resources}. 
 	\item \textbf{Inspection:} In general, external computers are not allowed unless explained to and allowed by the \iaterm{Technical Committee}{TC}.
-	  A team must announce to the TC at least 1 month in advance the external computing resources they want to use, for what purpose and how to reach the resources (e.g. specify the URL or IP address and port). Inspected software must meet the following \textbf{requirements:}
-	  \begin{itemize}
-	  	\item The software must be open source (BSD/GPL/etc), or
-        \item Detailed information about the propietary product must be provided (e.g. vendor, patent number, licencing, pricing, etc.), as well as publishing the interface for scientific use.
-	  \end{itemize}
+	  A team must announce to the TC during the inspection the external computing resources they want to use, for what purpose and how to reach the resources (e.g. specify the URL or IP address and port).
 	All relevant information must be specified in the team description paper.
 	\item \textbf{Connection:} The robot may connect to \iterm{external computing resources} via a network connection, e.g. the Internet. 
 	  The competition organisation cannot make any guarantees concerning availability, connectivity and performance of the connection. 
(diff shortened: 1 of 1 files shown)


general_rules/ExternalComputing.tex
```diff
@@ -17,15 +13,8 @@ \section{External computing}\label{rule:robot_external_computing}
 	\item \textbf{Autonomy:} The robot has to maintain full autonomy when using \iterm{external computing resources}, 
 	  meaning there may not be a human giving the robot any kind of instructions via \iterm{external computing resources}.
 	  It is up to the team to prove to the \iaterm{Technical Committee}{TC} that there was no cheating introduced via the \iterm{external computing resources}. 
-	  For example, the use of Amazon Mechanical Turk to classify and recognize objects during a competition will be considered cheating, since effectively a human will do the classification.
```
LoyVanBeek: The tele-operation bit is still relevant. It's hard to check that but that is certainly not supposed to happen. Should be obvious to anyone in the competition though...

semeyerz: Yes autonomy is still important. I only removed the example as examples are always incomplete and are redundant to the general rules.

LoyVanBeek: Just removing the specific example is fine, maybe I didn't read the diff correctly.

kyordhel:
I understand that the rulebook requires to be shrink, but at the same time, rewritten sections must be crystal clear.

This PR leaves the section under-specified. A new team will not know which information to include in the TDP. Plus, allowing external computing and allowing the use of cloud services are two entirely different topics, and the difference is not clear here.

kyordhel:
K. Let's set stuff into context.

1. If we are shrinking the rulebook (and by that I mean really rewriting the entire rulebook to a compact version of itself), let's do it for the next year.
  - If that's the case, I would ask all modifications fall not to master but to a 2019 branch to be merged after the competition.
  - Otherwise I think changes would be way to big for just a rule review (final draft was already published).
  - In fact, maybe we should discuss several sections to know how is more convenient to restructure the Rulebook.
2. Missing information,
  - What is required by the TC in order to authorize external/cloud resources.
  - If, as you said, is already specified in another section, there should be a cross reference.
3. Historically,
  - External Computing refers to computers accessible to the robot via the Arena Network without accessing the Internet.
  - Cloud Computing refers to all services accessible to the robot via Internet.

Now, being Extremely picky with the PR,
1. PR subject literally is: Allow external computation
  - This is confusing. Wasn't allowed before?
2. PR description refers only to an issue, with no details on why the PR is introduced
  - For sake of clarity, it should summarize what is changed and why.
  - For sake of clarity, why elements are removed should be clearly stated, and, if there was redundancy, where they can be found.
  - For sake of clarity, all related issues should be referred.

The former mostly because most TC members don't know the Rulebook by heart, and is always nice to them to set stuff into context.

semeyerz:
> 1. If we are shrinking the rulebook [ ... ]
sorry but I do not see why this is relevant for this PR.

> 2. Missing Information, What is required by the TC in order to authorize external/cloud resources.
I do not know. Was not clear before. In my opinion for external cloud stuff they should check licenses and so on but this would be another PR. This one is only about fixing inconsistencies.

> 2. If, as you said, is already specified in another section, there should be a cross reference.
There is already a cross reference from TDP to this section

> 3. Historically,External Computing refers to computers accessible to the robot via the Arena Network without accessing the Internet.
I can not remember that any team used local external computing in the last years.
I think it was not even allowed. Than there came this cloud rule that allowed "cloud services"

> 2nd 1.    PR subject literally is: Allow external computation   This is confusing. Wasn't allowed before?
Yes. Before only cloud services were allowed that are accessible by everyone.
A rule/section was added(#388) were to place local computing and so on but this new text was comflicting with the existing section on cloud services.
This PR cleans this up.

> 2nd 2. PR description refers only to an issue, with no details on why the PR is introduced

This is the default way. Raising an issue and then writing a pull request that fixes the Issue.

justinhart:
We should leave the bit about Mechanical Turk in there. Amazon isn't going to significantly profit from being mentioned as a specific ban in our rulebook. While I can appreciate the push to make the rulebook more concise, this also introduces a lot of issues regarding cross-referencing and what could be essentially seen as changes to the rules to make them more permissive. If we want to cut the rulebook for length, why not defer to 2019? If a team hasn't read the rulebook by now, I just don't know what to do with that.

semeyerz:
@justinhart if we leave the Rulebook like it is right now there is not only some "junk" in it but also conflicting rules regarding external computation.

kyordhel:
**Closing reason:** Not approved (no TC reviews) before Final version deadline.

This is not a reject. Please reopen PR over 2019/master

PR #447, opened 2018-05-16
Explicitely specifiy "announce bar location" in OC instructions

### LINKED ISSUE #436: Q: Where is the Bar located in Cocktail Party?
In the Cocktail Party task, are we told the location of the bar during setup days?

If not, what aspects of the bar allow us to identify it?  Will it be the only surface in the room with drinks atop it?

kyordhel:
Formerly, the test only instructed the O.C. to

> Specify and announce the location where the drinks are served

which although obvious, still raised concerns for not explicitly specify the drinks were served at the bar. This is now fixed.

This PR addresses and fixes #436

### DIFF
--- a/tests/CocktailParty.tex
+++ b/tests/CocktailParty.tex
@@ -89,7 +89,7 @@ The referees need to
 2h before test:
 \begin{itemize}
 	\item Specify and announce the rooms where the test takes place.
-	\item Specify and announce the location where the drinks are served.
+	\item Specify and announce the location where the drinks are served (i.e. bar location).
 \end{itemize}
 
 \newpage


semeyerz:
I think the problem in #436 was that there is no description for teams/robots.
Meaning that a team is not able to fully understand the task except it reads the instructions for the OC which is not obvious.
Teams should be able to understand what the robot should do and can expect without reading the OC and TC instructions.

kyordhel:
> We could rename those chapters to "Definitions/parameters for the test (to be announced by OC/TC)" for clarity.

Yes, @LoyVanBeek, but then we would need to change all tests in the rulebook accordingly and that's more suitable for 2019. As a matter of fact, the Bar thingy was never specified in any of the iterations of Cocktail party (2010 on) and never was a problem with it.

airglow:
In the past the bar location was equally to the default locations for drinks I guess, that's why it was not specified. Now this could be different, because the bar is not in the party room which may includes the default location for drinks. At GermanOpen we announced it (as bar), and in Nagoya (as drink location) as well so it makes sense to add it to the rulebook.

PR #451, opened 2018-05-17
Remove robot design award

airglow:
AFAIK there was no call for a robot design award this year, thus the section should be removed


kyordhel:
@airglow It was agreed to use a *rebase* policy in the rulebook (i.e. don't merge master). Do you think it would be possible for you to submit another PR without merges?

Regarding the removal of an award, I think we need to wait for approval from the Execs.

PR #487, opened 2018-11-04
[Appendix] Remove test-specific Appendices

moriarty:
This just deletes them... And hopefully the new tests can be described in such a way that they don't need to be re-added.


justinhart:
Agreed. This works for me.

PR #498, opened 2018-11-17
Fix typos

awesomebytes:
Meanwhile giving it a read I fixed a couple of typos and a mention to Open Challenge.

(There is another mention in finals but given I don't know the decision for the finals, I didn't touch it).


justinhart:
This all makes sense to me.

Issue #493, opened 2018-11-05
Take out garbage [restructured][draft]


moriarty:
## Main Goal

There is no garbage inside of the apartment

## Setup

There is garbage inside of the apartment. For example: One paper basket near a desk, and one recycling bin and one compost bin near the sink.

johaq:
Some proposed changes (Pull request #519):
1. The multiple collection zones seem unnecessary to me and overly complicated. Also, task is already hard without outside navigation. So I removed the outside one in favor of one zone near the entrance which also should be easily understandable by audiences.
2. Changed the bonus goals to emphasize the manipulation aspect. One rewards the robot for removing the lid from the bin and a very large bonus for carrying both bags at once to also have some encouragement for bi manual manipulation.
3. I think the bins should just be part of the normal arena setup. It's easier for referees because no additional setup required.
4. Added Deus Ex Machina rule to request a human to hand over the bag resulting in a 80% score reduction.

kyordhel:
Thanks @johaq for the PR. I still have to review it, but from this comment I got some counter-arguments
1. This test was a perfect excuse to test unsupervised outdoor navigation. As you say, it might make more sense to have the containers next to the entrance, so let's think about any other test that allows us to test navigation outside.
2. No lids please. Any bucket with a bag can serve as trash container. We are not sure we can get an standard trashbin with a lid.
    - If you insist in having a bin with a lid, please look for an IKEA one add it to the Objects section under General Rules
3. Add it to the Objects section under General Rules
4. Let's not work with percentages but net points (e.g. 400pts).

johaq:
1. Don't we have outdoor nav in restaurant and potentially GPSR? 
2. Was thinking sth like this: https://www.ikea.com/gb/en/products/small-storage-organisers/waste-sorting-bins/knodd-bin-with-lid-galvanised-art-10218937/
3. Will do
4. So far every DEM rule I have seen uses percentages

kyordhel:
1. GPSR faith is uncertain. Restaurant is Stage 2. We need to test outdoor navigation in Stage 1.
    1.1. I'm not saying We must keep it in this test, just that we might need put it somewhere else.
2. I love it but teams will hate you. Those containers are almost invisible to the Kinect/Prime Sense
![image](https://user-images.githubusercontent.com/8583949/51386166-d3a22380-1b21-11e9-9a81-d957d74dce16.png)
3. :+1: 
[...]

PR #500, opened 2018-11-26
Add initial draft of "Take Out the Garbage"

nickswalker:
Addresses #493

### DIFF
--- a/general_rules/Scenario.tex
+++ b/general_rules/Scenario.tex
@@ -57,8 +57,6 @@ The minimal configuration consists of:
 	\item a small table,
 	\item a small dinner table with two chairs,
 	\item a trash bin,
-	\item a recycling bin,
-	\item a compost bin,
 	\item an open cupboard or small table with a television and remote control,
 	\item a cupboard with drawers, and
 	\item a bookcase or shelf with doors and some books inside
--- a/general_rules/Scenario.tex
+++ b/general_rules/Scenario.tex
@@ -56,6 +56,9 @@ The minimal configuration consists of:
 	\item a couch,
 	\item a small table,
 	\item a small dinner table with two chairs,
+	\item a trash bin,
+	\item a recycling bin,
+	\item a compost bin,
 	\item an open cupboard or small table with a television and remote control,
 	\item a cupboard with drawers, and
 	\item a bookcase or shelf with doors and some books inside
--- a/tasks/TakeOutGarbage.tex
+++ b/tasks/TakeOutGarbage.tex
@@ -1,32 +1,34 @@
 \section{Take Out the Garbage [House Holder]}
-The robot has to take out the trash, recycling and compost.
+The robot has to take out the trash bags from the two bins in the apartment.
 
 % \subsection{Focus}
 % This test focuses on object perception, manipulation, and planning.
 
 \subsection{Main Goal}
-All garbage in the apartment has has been deposited into the correct exterior bin.
+All garbage bins in the apartment are empty and the garbage has been moved to the specified collection zone.
 
-\noindent\textbf{Reward:} 450pts (150pts per type of garbage)
+\noindent\textbf{Reward:} 600pts (300pts per bag)
 
 \subsection{Bonus rewards}
 \begin{enumerate}[nosep]
-	\item Opening the apartment door (100pts)
-	\item Closing the apartment door (100pts)
+	\item Moving a bag to the exterior collection zone (400pts, max one)
 \end{enumerate}
 
 \subsection{Setup}
 \begin{itemize}[nosep]
-	\item \textbf{Location:} This test takes place in the arena and its immediate exterior.
-	\item \textbf{Interior Bins:} There are three bins in the apartment for trash, recycling and compost.
-	\item \textbf{Exterior Bins:} There are three larger bins outside the apartment. They may have lids covering their contents.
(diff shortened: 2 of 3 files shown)


tasks/TakeOutGarbage.tex
```diff
@@ -0,0 +1,61 @@
...
+	\item \textbf{Location:} This test takes place in the arena and its immediate exterior.
+	\item \textbf{Interior Bins:} There are three bins in the apartment for trash, recycling and compost.
+	\item \textbf{Exterior Bins:} There are three larger bins outside the apartment. They may have lids covering their contents.
+\end{itemize}
+
+
+\subsection{Additional rules and remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Door:} The team decides whether the apartment door starts open or closed.
+	\item \textbf{Exterior bin lids:} If the robot is unable or has failed to open an exterior bin lid, it must clearly say so. The robot may then leave the bag or interior bin next to the exterior bin for a reduction of 30\% in scoring for that type of garbage.
```
maximest-pierre: I feel that if the robot are unable or failed to open the exterior bin lid the robot should ask the operator to open the lid. They should still have a penalty but less severe than 30%.

general_rules/Scenario.tex
```diff
@@ -56,6 +56,9 @@ \subsection{Furniture}
 	\item a couch,
 	\item a small table,
 	\item a small dinner table with two chairs,
+	\item a trash bin,
+	\item a recycling bin,
+	\item a compost bin,
```
kyordhel: This is a No-Go from my perspective. We are requesting two relatively big and expensive objects per arena to the LOC for a single, non-mandatory test. I'd just drop them and add only the trash bin. Maybe two small trash bins are OK.

The robot can take the kitchen and bathroom trash outside, to the appropriate designated location.

tasks/TakeOutGarbage.tex
```diff
@@ -0,0 +1,61 @@
+\section{Take Out the Garbage [House Holder]}
+The robot has to take out the trash, recycling and compost.
+
+% \subsection{Focus}
+% This test focuses on object perception, manipulation, and planning.
+
+\subsection{Main Goal}
+All garbage in the apartment has has been deposited into the correct exterior bin.
+
+\noindent\textbf{Reward:} 450pts (150pts per type of garbage)
```
kyordhel: Not sure if 150pts is enough reward for taking out a plastic bag from a bucket on the floor. It has never been done before.

Just 150pts for taking out, opening a lid, and placing inside is completely out of the question. Not even I am that cheap with scoring.

tasks/TakeOutGarbage.tex
```diff
@@ -0,0 +1,61 @@
...
+
+\subsection{Bonus rewards}
+\begin{enumerate}[nosep]
+	\item Opening the apartment door (100pts)
+	\item Closing the apartment door (100pts)
+\end{enumerate}
```
kyordhel: Threre's something missing here. The test should round up to 1000pts including all bonuses

maximest-pierre:
Can we change it so there is trash around the apartment and the robot must be able to sort and drop it in the right container? This seems more entertaining than seeing a robot move a bags across the arena and as a bonus, we could make the robot bring the bins outside of the arena.

kyordhel:
> Can we change it so there is trash around the apartment and the robot must be able to sort and drop it in the right container?

We already have that task. Check the Clean-Up test in https://github.com/RoboCupAtHome/RuleBook/blob/2019/restructure/tasks/CleanUp.tex

nickswalker:
@kyordhel I think the changes address the bulk of your concerns. This version would allow teams to score the majority of points for taking the pre-sealed trash bags from two buckets in the arena to a collection zone somewhere inside the arena.

As I can see so far, the purpose of the bonus points distinction is to reward teams that accomplish challenging and desirable-to-end-user extensions of the task. From this simplified version of the task, my personal most desired capability would be for the robot to go the last mile and put the trash outside the door so I don't have to manipulate the trash myself. Also important, but potentially significantly more challenging, would be replacing the bag (fine grained manipulation of a deformable object that probably can't be accomplished without bimanipulation). I see no reason not to include both of these as options for ambitious teams. I haven't thought through from where the robot would obtain the bag or other details, so I haven't included that in these changes.

general_rules/Scenario.tex
nickswalker: Agree. Moved to a minimum of a single bin which may be small (though a second reduces the shuffling that the ref has to do).

tasks/TakeOutGarbage.tex
nickswalker: Bumped to 300pts for taking a single bag out and putting it on a designated spot on the floor somewhere in the arena.

tasks/TakeOutGarbage.tex
nickswalker: Ah, I read 3.6.3 "The maximum total score per task in Stage I is 1000 points." as, for each task, the scoring of that task is no more than 1000 points. The correct reading is for each task, the maximum attainable score is equal to 1000pts.

tasks/TakeOutGarbage.tex
nickswalker: I agree that this is too much to ask for too little return, so the latest revision drops the lids altogether.

justinhart:
This looks like it works. We can always edit the tasks later to tweak, but it belongs in the competition and the current format is about what was asked for.

PR #501, opened 2018-12-18
Replace "House Holder" with "Housekeeper"

nickswalker:
Householder -- a person who occupies a house or tenement alone or as the head of a household
Housekeeper -- one employed to manage the domestic duties involved in maintaining a house


maximest-pierre:
thank you for the patch :smile_cat:

PR #503, opened 2019-01-10
fix times of afternoon test slot

MatthijsBurgh:
The proposed times are just a guess. I don't know if you had you something different in mind.

### DIFF
--- a/general_rules/Organization.tex
+++ b/general_rules/Organization.tex
@@ -97,7 +97,7 @@ In case of having no considerable score deviation between a team advancing to th
 	\multicolumn{1}{ c }{}
 		& \multicolumn{4}{ c }{\wcell{0.5\baselineskip}{\color{gray}Lunch}}\\\cline{2-5}
 
-	\cell{Block 2\\\footnotesize(9:00 - 12:00)}
+	\cell{Block 2\\\footnotesize(14:00 - 17:00)}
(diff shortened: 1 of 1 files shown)


general_rules/Organization.tex
```diff
@@ -97,7 +97,7 @@ \subsection{Schedule}
 	\multicolumn{1}{ c }{}
 		& \multicolumn{4}{ c }{\wcell{0.5\baselineskip}{\color{gray}Lunch}}\\\cline{2-5}
 
-	\cell{Block 2\\\footnotesize(9:00 - 12:00)}
+	\cell{Block 2\\\footnotesize(14:00 - 17:00)}
```
maximest-pierre: So we are putting 2 hours for lunch and repairs and preparation

MatthijsBurgh: 1 hour is definitely to short IMO, 1,5 maybe ok, maybe to short. 2 hours definetly good.

maximest-pierre: I would push the upper limit from 17:00 to 18:00. Just to make up for lost time or a late start in the morning.

kyordhel: I'd go for a 3 hour gap to give enough time to teams to prepare and eat. Also, this adds a nice gap in which we can fit an Open Challenge for those interested.

PR #507, opened 2019-01-14
Add Find My Mates task

awesomebytes:


maximest-pierre:
The only thing I notice is that the explanation is not that clear of what the end goal is. It took me 5 reading of this task before having an understanding of what is the robot suppose to accomplish. A small modifications would be adding an example to make the whole thing clearer and more concise of what the robot must accomplish. Other than that the task is actually going to be hard and fun for teams :+1: .

awesomebytes:
@syedaraza Can you incorporate an example please?

We tried to keep it very short (and actually avoid examples to not 'force' people to solve the task in a specific way) but it does make sense if its not clear enough, sure!

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
+\section{Find My Mates [Party Host]}
+
+The robot starts in a room and waits for someone to approach it. The guest knows the name of other guests at the party but doesn’t know how these people look like and where the people are. The robot will gather this information for the guest.
+
+
+\subsection{Main Goal}
+The robot provides the location and unique description of 3 guests of the party as required by another guest.
+
+\noindent\textbf{Reward:} 750 points (250 points per guest).
```
kyordhel: Check if max can be 500 with option to raise it via bonuses. I would reduce it to 150pts per guest, providing only location.

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
+\section{Find My Mates [Party Host]}
+
+The robot starts in a room and waits for someone to approach it. The guest knows the name of other guests at the party but doesn’t know how these people look like and where the people are. The robot will gather this information for the guest.
```
kyordhel: I'd change it to this:

> The robots start in a predefined room waiting for an operator. The operator has only the list of guests but does not know them. The robot shall describe all the people in the list to the operator.

However, let's see if we can shrink it to a one-liner

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+
+
+\subsection{Main Goal}
+The robot provides the location and unique description of 3 guests of the party as required by another guest.
```
kyordhel: This is more clear 
> The robot provides the location [and a comprehensive, unique description] of at least 3 party guests to the operator.

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+	\item \textbf{People}: 5 people distributed around the house.
+\end{itemize}
+
+\subsection{Additional rules and remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per guest.
+
+	\item \textbf{Calling the operator for the final answer:} 50\% reduction if the robot does not go back autonomously to the operator to report the information.
+	
+	\item \textbf{Penalty:} If the robot talks to an empty room or does not clearly face a person or crowd when talking (e.g. speaks to a wall), -75 points of penalty.
```
kyordhel: Let's remove this

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+	\item \textbf{Location}: The robot is in a predefined room at least one person that will approach it.
+
+	\item \textbf{People}: 5 people distributed around the house.
+\end{itemize}
+
+\subsection{Additional rules and remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per guest.
+
+	\item \textbf{Calling the operator for the final answer:} 50\% reduction if the robot does not go back autonomously to the operator to report the information.
```
kyordhel: Ambiguous

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+
+\subsection{Bonus rewards}
+\begin{enumerate}[nosep]
+	\item Robot providing unique description of a person without having asked such information (75 points)
+\end{enumerate}
+
+
+\subsection{Setup}
+\begin{itemize}
+	\item \textbf{Location}: The robot is in a predefined room at least one person that will approach it.
```
kyordhel: Change to: ```This test takes place inside the arena in a predefined room (e.g. living room)```

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+
+	\item \textbf{People}: 5 people distributed around the house.
```
kyordhel: Let's constrain it to just the living room

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+\end{itemize}
+
+\subsection{Additional rules and remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per guest.
```
kyordhel: This must be extended as follows:
- -75pts if a person has to wave of call the robot in order to be found
- -75pts if a person has to tell the robot where is she sitting/standing
- -150pts if a person has to approach to the robot (e.g. walk and stand in front of it)

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+\end{enumerate}
+
+
+\subsection{Referee instructions}
+
+The referees need to
+\begin{itemize}
+	\item Make sure there every person has a name.
+	\item Make sure people stay roughly in the same place.
+	\item Distribute the people in the rooms.
```
kyordhel: Yeah... Nope.

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+\end{itemize}
+
+\subsection{OC instructions}
+
+2h before test:
+\begin{itemize}
+	\item Announce the location where the robot will be instructed.
```
kyordhel: Missing:
- Assign predefined names to the guests
- Announce predefined room in which test takes place

tasks/FindMyMates.tex
```diff
@@ -0,0 +1,52 @@
...
+
+
+\subsection{Main Goal}
+The robot provides the location and unique description of 3 guests of the party as required by another guest.
+
+\noindent\textbf{Reward:} 750 points (250 points per guest).
+
+\subsection{Bonus rewards}
+\begin{enumerate}[nosep]
+	\item Robot providing unique description of a person without having asked such information (75 points)
```
kyordhel: This reward is way too small for the complexity. Is easier to find a person than to describe her.

Consider the following:
- 50pts navigating back to the operator and provide requested information.
- 50pts for retrieving the list of guests (automatically scored if good ASR).
- 150pts for providing a a unique and comprehensive description of a guest (e.g. gender, clothing, skin, or hair color, pose, etc).

tasks/FindMyMates.tex
awesomebytes: In the meeting we said 750 points for the main goals and 250 for the bonuses for all tests...

tasks/FindMyMates.tex
awesomebytes: That sounds less mechanical, I like it.

tasks/FindMyMates.tex
awesomebytes: I'm alright with your proposed version. But the sentence "at least 3 party guests" makes me wonder if we can just give the bonus points for extra people given their location and unique description (after given the main goal ones, ofc).

tasks/FindMyMates.tex
awesomebytes: Ok.

tasks/FindMyMates.tex
awesomebytes: What?

tasks/FindMyMates.tex
awesomebytes: Aren't we providing points just for the Main goal achieved? This sounds to me like going back to scoring lots of little things.

awesomebytes:
> This test requires to be reworked. In fact, it can be split in two tests.
> 
> ### Option 1 (Who is who)
> As it is, is quite ambiguous whether the goal is locating people in the apartment or describe people. These are two different features quite challenging on their own, so I would separate. All proposed changes address _people description_ and not far-distance people finding. Hence the test is constrained to one room.
> 
> The robot would score 150pts for finding 3 people in a room (something robots haven't really mastered yet). There are also 50 points for delivering the report so, if completed, the task rounds up to 500pts. Optionally, for reaching the 1000pts there is a speech recognition component and a [semantic] person-identification component. Further, the robot might skip finding people at all (or bypass it by waving) and focus in feature identification and recognition to provide a description.
> 
> ### Option 2 (Who's there?)
> The test description and main goal must emphasize that the robot must find people in the apartment. Operator might say something like:
> 
> _tell me who else is in the apartment_
> 
> The robot then shall look for people over there (let's say there are 4: 2 standing, 2 sitting, 1 lying). In this case, I would grant 200 points per person found + 50 points per retrieved name. Optionally, the robot can find a third person that would grant 300pts + 100pts for retrieving the name and 100 extra points for returning with the operator and report all the names. The third person found grants more points due to speed (is very hard for a robot to find 3 people in the apartment in 5 minutes), so most teams will go just for finding the first two people and perhaps delivering the list.

I don't agree on splitting the test. Given one can use as much HRI as you want, the test should be very doable. If a team wants to implement the test by mostly using HRI but have score reduction on it. Cool. Maybe that can be an initial implementation and then keep building on it up until the point it is doable without the score reducing HRI.

tasks/FindMyMates.tex
kyordhel: There might be a fourth if there is enough time for outstanding performance.

tasks/FindMyMates.tex
kyordhel: For consistency depends on the focus. If you're finding people in the house, point 3 stays. Otherwise rules out.

tasks/FindMyMates.tex
kyordhel: For consistency with other tests, we want a test to be
- Main goal: ~500pts
- Bonus goals: ~500pts
- Maximum total score: **exactly 1000pts**

Main goal, as I understand the test, is to find a person. Additionally, the robot may get the person's description.
- If the robot fails to find a person, may ask the person to wave (score reduction)
- If the robot fails to find a person, may ask the person to stand in front (no points)
- Whether the robot found the person or not, the robot may ask the person for a description
- If the robot is fast enough, it can navigate back and deliver the person's location and description for points

Otherwise you'll have teams frustrated because the robot failed to either find the person, or get a description of her. If, on top, you look in the entire apartment, is too much for a Stage 1 test.

kyordhel:
Push the task to Stage 2 and I'm ok with it as is. For a Stage 1 test, I think you're asking too much from the robots, Specially Pepper.

@balkce @justinhart @johaq What do you think? Find 3 people in the entire arena and then provide location and description of each in less than 5 minutes.

balkce:
If the issue is the amount of persons, finding 2 out of 5 people seems doable in five minutes.

I propose to grant scoring per person found, with the main goal being finding the number of persons you think is doable (again, 2 seems fine), with bonuses given if the robot finds more. This way the team can decide to either find a person and come back to describe them, or find several and come back to describe them all.

tasks/FindMyMates.tex
syedaraza: For the main description we can have something like,

__Find My Mates [Party Host]__
The robot fetches the information of the party guests for the operator who knows only the names of the guests.

tasks/FindMyMates.tex
syedaraza: And the main goal can be more concise like this,

__Main Goal__
The robot reports to the operator the location and description of at least two party guests.

tasks/FindMyMates.tex
syedaraza: The test should have three more rules in addition to Deus ex Machina,

__Reporting without returning to the operator:__ 50\% reduction in points if reporting the retrieved information without returning to the operator.

This rule is a substitute to providing bonus points for navigating to the operator. If the robot is running out of time, it can tell the information without returning to the operator. This way the robot still gets the points for retrieving the info but with reduction. Also, it can get some points for information gathering if the navigation fails while returning to the operator.
	
__Description of a Person:__ The robot should provide at least three descriptions of a person which may include information about gender, clothing, skin, face, hair, pose, etc.

__Description of Location:__ Not sure how to describe the location if we suppose all the guests are in the same room. It needs to be discussed.

tasks/FindMyMates.tex
syedaraza: 'Calling the operator' can be replaced by 'reporting without returning to the operator' as defined above.

tasks/FindMyMates.tex
syedaraza: If all the guests are in the same room then how to describe the location of each guest? Should we drop the description of location?

tasks/FindMyMates.tex
syedaraza: And it would be convenient to limit the guests to 4. It will help in summing up the total score to 1000 pts (if we distribute 250 pts per guest).

tasks/FindMyMates.tex
syedaraza: Agree. But the last point should also mean that there will be no deduction if the robot requests to stand straight in front of it after finding a person.

syedaraza: For the sake of detecting the descriptors of a person

tasks/FindMyMates.tex
syedaraza: The second point can be like,

* Make sure people stay at the designated place.

Then the third point becomes redundant.

tasks/FindMyMates.tex
syedaraza: Should also have,

* Assign predefined location to each person.

If we want the robot to report the location too.

Also, OC should announce the list of names. This way the robot will know the list of names before the test starts which will save time to achieve the main goal.

tasks/FindMyMates.tex
syedaraza: A simple and clear rewarding scheme can be like, 

__Rewards__ _500 points (100 pts for location and 150 pts for description per guest)._

__bonus rewards__ _For retrieving information about each additional guest there will be 100 points for location and 150 pts for description. A total of 500 bonus pts._

However, if we drop the reporting of location then will it be justified to give away 250 pts just for the description?

tasks/FindMyMates.tex
kyordhel: That's definitively a No-Go from all perspectives. Wherever the robot is looking for people, the people shall not have assigned a predefined location. Even less two hours before the test. We don't know which volunteers we can rely on and for how long. Further, people must be swapped to prevent cheating.

Regarding the list of names, that's published during setup days. It's in General Rules

tasks/FindMyMates.tex
kyordhel: This is getting way too complex.

- **Reporting:** Only scored if the robot goes back and delivers the report in time. Otherwise it makes no sense.
- **Description of a person:** No *at least*. If there is a single bald person and the robot is smart enough to say: John is the bald guy in the living room, one descriptor suffices. If, on the other hand, I've 2 tall, Caucasian, bearded, orange-and-black dressed guys, the robot may also need to report that Max is wearing slippers and Loy has baby face.
- **Description location:** Sammy is the pale guy sitting in the coach. Anna is the bearded dude standing next to the lamp. I see no problem there.

tasks/FindMyMates.tex
kyordhel: We need to be careful with the main goal. If we include the report, points can only be scored if the robot delivers the actual report.

tasks/FindMyMates.tex
kyordhel: I rather focus in giving points based on how hard is to accomplish a task, then round to sum.

Yet again I think that a reward of 100 points for finding a person in the whole apartment is scarce. I'd say the same for a semantically meaningful description of the people (i.e. listing descriptors is **not** a valid approach).

johaq:
I'm also ok with it since it is a hard challenge pushing teams to get better to fully solve this task but it also allows to be partially solved and successful if a team is not quite there yet i.e. only find and describe one person as @balkce mentioned.

tasks/FindMyMates.tex
syedaraza: 100 points are for finding the guests in the same room.

syedaraza: Assuming that finding the third person will be very challenging in 5 mins, we can give bonus of 300 pts for description and 200 pts for location for reporting the third guest.

Issue #492, opened 2018-11-05
Clean Up [restructured][draft]


moriarty:
## Main Goal

There are no items, clutter or garbage where it doesn't belong

## Setup

There are items where they don't belong.

For example: Toys on the floor, Homework on the diner table, clothing on the bedroom floor, take-out boxes on the kitchen counter.

kyordhel:
@johaq Some comments:
1. *Clean up* test is tradition since 2009! We shouldn't change it!
2. We have the "leaving the arena" bonus to promote a quick execution (extra points if you're fast) and to avoid people entering the arena to pick up the robot, which is logistically a nightmare.
3. Just make sure:
  - Main goal: ~500pts
  - Bonus goals: ~500pts
  - Sum: *exactly 1000pts*
4. Agreed. Consider those numbers are to keep fairness and to help OC setting up the task, so maybe setting a range would be better.
5. Decoration is out of robot's reach. Unknown objects are _garbage_. Robot must be smart enough to deal with it. It wasn't a problem in 2009, shouldn't be a problem today.
6. Thanks

johaq:
1. Fair enough. I'll change it back.
2. Ok. I'd say bonus for quick execution is you get to more objects. I can add it back but I would give it less points. If we add it back I'd also think we should clarify that the points are only available if the robot successfully carried away x objects. Otherwise we get teams going inside the arena, stating I cannot grasp, leaving the arena, and then wanting the bonus points.
3. I was not aware that was the standard.
4. I like the flexibility so the referees can decide at the competition how they want to handle things and maybe up the ante over the years without needing to change numbers in the rulebook.
5. Ok, I think the description should then clearly state that these objects are garbage i.e. candy wrappers.

I'll update my pull request accordingly tomorrow.

johaq:
4. "Hey this year this task is new, lets go for 4 known and only one unknown for every team." - "Last year went well, lets put more unknowns and alike." 
I think referees can just quickly discuss this at each competition to have it fair and consistent for each team.

johaq:
I updated my pull request. Thanks for the feedback!

kyordhel:
Bump. Check #517

PR #517, opened 2019-01-16
Clean Up

johaq:
#492


johaq:
Seems like that file needs a complete update as it still contains old tests. I'll update it in a separate pull request so we can add to it when new score sheets become available since as far as I can see some tasks still need theirs.

kyordhel:
Ok. Just be sure to add the test and generate the scoresheets file to check how it looks. Sometimes items need rewording due to aesthetic and space constrains.

scoresheets/CleanUp.tex
```diff
@@ -0,0 +1,25 @@
...
+The maximum time for this test is 5 minutes.
+
+\begin{scorelist}
+	\scoreheading{Main Goal}
+	\scoreitem[5]{100}{Place an object at the appropriate location}
+	\penaltyitem[5]{-30}{Receiving human help (telling where an object can be found)}
+	\penaltyitem[5]{-40}{Receiving human help (pointing out object to be moved)}
+	\penaltyitem[5]{-20}{Receiving human help (Guiding the robot near an object to be moved)}
+	\penaltyitem[5]{-30}{Receiving human help (telling robot which category an object is or where to place it)}
+	\penaltyitem[5]{-60}{Receiving human help (physically interacting with object)}
```
kyordhel: I'm dubious about this last point. If a human operator moves the object and places it in the right spot, it might well cause a 100pts reduction (net to zero)

tasks/CleanUp.tex
```diff
@@ -1,67 +1,71 @@
...
-The robot has to clean the room from the house it has been told to clean up.
+The robot has to tidy up a room inside the arena.
 
 % \subsection{Focus}
 % This test focuses on object perception, manipulation, and planning.
 
 \subsection{Main Goal}
-Upon entrance, the robot requests the operator which room shall be cleaned.
-All misplaced known objects found in this room must be taken to their predefined locations and unknown objects thrown to the trash bin.
+Inside one room in the arena are some misplaced objects. The robot has to find each one and bring them to their predefined locations. Unknown objects have to be thrown into the bin. Near the arena entrance will be an operator who can tell the robot which room is messy and will open the entrance door if the robot requests it.
```
kyordhel: - There are some misplaced objects in one of the arena's rooms.
- Unknown objects are thrown to the bin

@justinhart  maybe you can give later a buff

scoresheets/CleanUp.tex
```diff
@@ -0,0 +1,25 @@
...
+
+	\scoreheading{Bonus rewards}
+	\scoreitem{200}{Opening entrance door without human help}
+	\scoreitem{150}{Moving a \emph{tiny} object}
+	\scoreitem{150}{Moving a \emph{heavy} object}
```
moriarty: whoever is running the test will also need to check that not only the number of objects in total is the same for all teams, but also that the number of `tiny` and `heavy` objects are the same for all teams.

scoresheets/CleanUp.tex
moriarty: Yes, I think the distribution penalties is odd.

If I point from across my living-room from the sofa, towards my desk, and say `tidy up that crap` and the robot does it and I haven't gotten off the sofa, then I think it's okay. Same with me saying `go clean up the kids toys from the spare bedroom` and again, I haven't gotten off the sofa, then I again think it's okay.

But they're all high, which means the teams are really pushed to solving the whole task as is so I think that is good.

kyordhel:
@johaq We discussed yesterday the object numbering in the meeting. We came out with the following decision:

- To keep consistency between local (open) tournaments and the Worldcup, we need to specify a range of objects (e.g. 5-10).
- It's up to robots to interact with available objects or not.
- The number of objects available can never be smaller than the number of objects to be scored.
- The number of objects to be scored is already in the scoresheet and so tests are balanced. There might be variation in the number of present objects but all teams in all leagues shall be evaluated the same. Hence this number remains fixed.

Same as with Storing Groceries, there will be more objects to clean than points, so the robot chooses what to pick and what not to. 

Please do the pertinent corrections so we can merge.

scoresheets/CleanUp.tex
kyordhel: Let's leave the scoring at criteria of the referee. If a human solved the task, no points. If the human mostly solved the task but the robot somehow helped, it will score.

tasks/CleanUp.tex
```diff
@@ -1,67 +1,71 @@
...
 	\begin{itemize}[nosep]
-		\item\textit{Known objects}: Any two regular and two alike objects.
-		\item\textit{Unknown objects}: One unknown object at grasping distance (i.e. no decorations).
+		\item\textit{Known objects}: Regular and alike objects.
+		\item\textit{Unknown objects}: Garbage lying around at grasping distance.
 	\end{itemize}
 \end{itemize}
 
-\subsection{Additional rules and remarks}
+\subsection{Additional Rules and Remarks}
```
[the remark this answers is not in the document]
kyordhel: Yes. They are. Is harsh, but you're still scoring.
The idea is foster autonomy.

johaq:
Added back object ranges in description.

PR #519, opened 2019-01-18
Take out Garbage

johaq:
#493

### DIFF
--- a/tasks/TakeOutGarbage.tex
+++ b/tasks/TakeOutGarbage.tex
@@ -1,4 +1,4 @@
-\section{Take Out the Garbage}
+\section{Take Out the Garbage [Housekeeper]}
 The robot has to take out the trash bags from the two bins in the apartment.
 
 % \subsection{Focus}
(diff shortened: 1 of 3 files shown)


tasks/TakeOutGarbage.tex
```diff
@@ -1,43 +1,49 @@
-\section{Take Out the Garbage [House Holder]}
```
kyordhel: Why are we removing the Scenario?

johaq: Yeah I was at first thinking this is a leftover of when we had this every task as GPSR idea in mind. Are these scenarios mentioned in the general rules? I think they should be if we have them in the title.

kyordhel: The scenarios are mentioned in §3.6.1 and §3.6.2. Manipulation tasks can only be chosen within *House Holder/Keeper* scenario, while HRI tasks belong to *Party Host*. Test should specify to which scenario they belong.

johaq: Yeah definitely. I'll add it back in my PRs

Issue #490, opened 2018-11-05
Storing Groceries [restructured][draft]


moriarty:
## Main Goal

All groceries put away.

## Setup

Groceries have been delivered to the front door via some fancy online grocery service.

nickswalker:
> **HINT**: The robot can ask the referee where to place the carried object (relative positions or pointing are both allowed).

Instead of framing this as a hint, it may match more with the rest of the tasks to include this as a Deus ex Machina in the remarks. That would also be the place to highlight whether we want to/how much to expressly penalize human help with the manipulation.

johaq:
My proposed changes (pull request #520):
1. I removed a lot specific mentions of numbers. Again, I feel they are unnecessary to the execution of the task, makes refereeing harder and less flexible, and encourages teams to hard code information.
2. As @nickswalker mentioned I moved the hint to the DEM section as it fits better with the other tasks.

I also would like to discuss removing the part about the operation area being clear. I understand it has been added recently by request of some teams so I did not just want to remove it without discussing it.
I think it contradicts parts of the general rules (robot always being safe) and I do not feel it puts a lot of extra burden on teams. Robots without sufficient sensors can always announce going backwards and be safe that way.

kyordhel:
Bump. Check #520

johaq:
Regarding #545. One way I see to make this task significantly easier to setup would be to remove the part about grouping in categories. I would be fine with just seeing a robot get objects into the shelf

kyordhel:
@johaq The point of this test is categorization. Pick and place is more or less a must.

kyordhel:
As I see it, it is quite simple. You set the cupboard once (grouping apples with apples and cokes with cokes is trivial to us) and let the robots do their best. Every time a robot finishes a test you take the objects it moved back to the table.

Shouldn't take more than 10 seconds, but I agree description must be shorten.

johaq:
The way I see it that is already in the rules. It states nowhere you need to reset the cupboard everytime.

I added back the 5-10 object range as requested and removed the clear area rule since no one seemed to object to removeing it.

kyordhel:
@RoboCupAtHome/technical-commmittee Can we consider the Storing Groceries task as complete?

Can we close the issue?

johaq:
I think so, yes

PR #520, opened 2019-01-18
Storing Groceries

johaq:
#490

### DIFF
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -3,7 +3,7 @@ The maximum time for this test is 5 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[5]{100}{Move an objects next to their peers in the shelf}
+	\scoreitem[5]{100}{Move an object next to their peers in the shelf}
 	\penaltyitem[5]{-30}{Receiving human help (point at target location)}
 	\penaltyitem[5]{-100}{Receiving human help (move object)}
 
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -3,7 +3,7 @@ The maximum time for this test is 5 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem{500}{Move 5 objects next to their peers in the shelf}
+	\scoreitem[5]{100}{Move an objects next to their peers in the shelf}
 	\penaltyitem[5]{-30}{Receiving human help (point at target location)}
 	\penaltyitem[5]{-100}{Receiving human help (move object)}
 
--- a/tasks/StoringGroceries.tex
+++ b/tasks/StoringGroceries.tex
@@ -1,15 +1,15 @@
-\section{Storing Groceries [Housekeeper]}
-The robot stores groceries into the shelf next to the objects of the same kind that are already there (e.g. for instance by placing apples near pears and bananas).
+\section{Storing Groceries}
+The robot stores groceries into a pantry shelf while paying attention to sorting objects in their appropriate place, i.e. storing an apple next to other fruits.
 
 % \subsection{Focus}
 % This test focuses on the detection and recognition of objects and their features, as well as object manipulation.
 
 \subsection{Main Goal}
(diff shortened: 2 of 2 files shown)


tasks/StoringGroceries.tex
```diff
@@ -48,16 +45,12 @@ \subsection{Setup}
...
 
-	\item \textbf{Object distribution:} The 10 objects to be moved are evenly distributed in random fashion including
-	4 known objects,
-	4 alike objects, and
-	2 unknown objects.
-	Among these, the robot will always find
-	a heavy object,
-	a tiny object, and 
-	an amorphous object.
+	\item \textbf{Clear area:} The robot may assume that the working area is clear, (i.e. can move slightly backwards for its task).
```
kyordhel: I strongly disagree with this change. It is necessary to keep fairness (even distribution among teams) and for logistical reasons. Further, the heavy and tiny objects are necessary due to bonus points (see [scoresheet](https://github.com/RoboCupAtHome/RuleBook/blob/2019/restructure/scoresheets/StoringGroceries.tex)).

johaq: I agree that distribution has to be equal for all teams but I think that is something for the referees to decide on location and is not something that should or needs to be in the rulebook. Why not allow the referees the flexibility to decide: "Hey at this competition no team is really strong in manipulation let's use easier objects for everyone." And also if teams get stronger we do not need to update single numbers from year to year and keep more consistency in the rulebook over time

kyordhel: Because scoring needs to be consistent among tests. This tests makes pick and place an object automatically worth ~100pts, so it can be taken as measure for assigning scores to other tests. Further, we are implying that each manipulation should not take more than one minute.

Besides, we need to be consistent in the three leagues (or at least between OPL and DSPL). If you leave the decision to the referees, the difficulty will vary between leagues and you have no consistent progression over the years.

Teams know:
- There are [5, 10] objects to deal with
- We must move 5
- 3 are easy to recognize (and grasp?)
- The rest are a PITA. (← all decision making enters here)

So yes, a state machine can be used, but a more robust decision making will produce better results to maximize scoring. IMHO, that setup is way too complex for a state machine.

johaq: I don't see how this affects consistency among other tests.
Consistency among leagues I understand but don't the referees switch leagues anyway during a competition and talks are with referees of all leagues.
Consistency over the years I agree is lost a little here but definitely not entirely.
I prefer teams know:
-There are several objects on a table
-We get points for moving up to five of them
-Our robot should know which of the objects are easy to grasp for it

scoresheets/StoringGroceries.tex
```diff
@@ -3,7 +3,7 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem{500}{Move 5 objects next to their peers in the shelf}
+	\scoreitem[5]{100}{Move an objects next to their peers in the shelf}
```
kyordhel: Careful!

```\scoreitem{500}{Move 5``` means that team won't score unless 5 objects are moved.
```\scoreitem[5]{100}{Move an``` means one object is enough to start scoring bonus points.

I agree on the 2nd one. HOWEVER: *move an objects*?

johaq: Yeah should be "an object"
I think second option is more in line with scoring in clean up task which seems similar in some ways.

kyordhel: :+1

kyordhel:
I'm OK with this so far, but lets hear other TC members opinion on the removal of the object distribution  before merging.

@RoboCupAtHome/technical-commmittee @moriarty @justinhart @awesomebytes your comments, please

kyordhel:
@johaq We discussed yesterday the object numbering in the meeting. We came out with the following decision:

- To keep consistency between local (open) tournaments and the Worldcup, we need to specify a range of objects (e.g. 5-10).
- It's up to robots to interact with available objects or not. 
- The number of objects available can never be smaller than the number of objects to be scored.
- The number of objects to be scored is already in the scoresheet and so tests are balanced. There might be variation in the number of present objects but all teams in all leagues shall be evaluated the same. Hence this number remains fixed.

Please do the pertinent corrections so we can merge.

Issue #495, opened 2018-11-05
Receptionist [restructure][draft]


moriarty:
## Main Goal

All guests have been guided to the living room and their outerwear has been taken care of.

## Setup

Guests are outside of the entrance, they arrive individually or in groups, some might have outerwear which shouldn't be worn inside the house.

## OC Preparation

There should be some storage near the entrance: coat rack, umbrella bin, shoe rack _etc._

RemiFabre:
In the current version of the draft it says : 
"Sitting the female guest next to a male guest (100pts)"
=> To avoid randomly giving points to robots that managed this without doing it on purpose, robots should clearly state who is close to the pointed sitting location. Something like "You can sit here, next to Eric". Since this is harder, maybe we could increase the bonus points for doing it?

kyordhel:
Nope. I don’t think so.

This feature might be removed to prevent enrage politically correctness inquisition, but if it stays:

1. It is there to make clear to the referees that the robot correctly guessed the gender of a person. 
2. We must keep the test as simple as possible.

justinhart:
I still think that this test doesn't really address HRI and really needs
some re-working to be a concrete HRI thing.

kyordhel:
It might be the case @justinhart, but thinking it needs reworking won't get the reworking done. We need proposals ASAP.

You can nominate someone if you wish.

justinhart:
Yeah, but I still don't really see these as HRI tasks. Finding the guest is
sort of a face recognition thing, right? That seems like computer vision,
and requesting the operator could be a simple as blurting out "I need an
operator," right?

How about something like this. If we modified the "Farewell" task like
follows, this would incorporate some of what I view as HRI elements. Maybe
something similar could be injected into Receptionist.

1) Guests must wave, look-at, or otherwise non-verbally indicate to the
robot that they are ready to leave.

2) If the guest engages in conversation with another guest, the robot must
wait for the conversation to end. If it lasts for more than 1 minute, the
robot should ask if the guest is ready to go and go help another guest if
they are not. If not, the robot should resume helping the guest to leave.

syedaraza:
I agree there is more room to improve HRI.

I think it should be the robot's choice how to find someone for the help in opening the door. For example, the robot can go to the living room and say if someone can help me in opening the door then please raise your hand or the robot goes to each guest one by one ask if they can help in opening the door. I don't think facial recognition would be a must in this case.

[...]

PR #523, opened 2019-01-27
Added a rule in the Receptionist task for opening the door.

syedaraza:
Regarding the comment here https://github.com/RoboCupAtHome/RuleBook/issues/495#issuecomment-457883229


awesomebytes:
I agree with @kyordhel

Issue #496, opened 2018-11-05
Farewell [restructure][draft]


moriarty:
## Main Goal

All guests have left the party with their belongings.

## Setup

There are guests attending the party and which previously gave their belongings to the robot.
See #495. But for the purpose of this task, it is not required that the robot memorize which belongings belong to each guest, they should be able to solve the task using natural language to describe the item.

kyordhel:
@moriarty Giving object description to retrieve belongings sounds way too hard to me (at least for Stage I). Main idea was just retrieve the coat (by color), but we want to skip manipulation in all *Party Host* scenario because *House Holder* is basically manipulation only.

I aso had something else in mind and so is reflected in the commit 736b26e.

awesomebytes:
First, let me copy-paste here a version of the test:

```

5.3 Farewell [Party Host]

It’s raining outside. The robot takes tired guests to their cab/uber. Ladies go first.

5.3.1 Main Goal
Some guests are tired, so they call the robot to retrieve their coat. It’s raining outside and there is only one umbrella, so the robot takes the guests one by one to their cab and returns with the umbrella.
Reward: 600pts (300pts per guest).

5.3.2 Bonus rewards
Delivering the right coat (150pts each, max 300pts)
Identifying the female calling person and escorting her first (100pts)

5.3.3 Setup
Location: The test takes place inside and outside the arena. Guests are in the living
room.
Start Location: The robot starts inside the arena at a predefined location near the
entrance door.
Guests: Initially there are five people in the living room, three sitting and two standing. At least two guests are female. All guests have a name and coat color assigned.
Calling guests: Only two guests are willing to leave. Both guests, male and female, will call the robot by waving or shouting at the same time.

5.3.4 Additional rules and remarks
Partial scoring: The main task allows partial scoring. Robot score per guest upon re-entering the arena.
Deus ex Machina: Score reduction for requesting human assistance is applied per guest as follows.
Indicating leaving guest: Telling the robot which guest is leaving (e.g. verbally or pointing at her) causes a reduction of 10% in scoring for that particular guest.
Guiding to the cab location: Guiding the robot where the Cab Driver is standing causes a reduction of 40% in scoring for that particular guest.
Guiding back to the house: Guiding the robot back to the house causes a reduction of 40% in scoring for that particular guest.
Coat handover: Handing over a coat to the robot (i.e. no hook off) causes a reduction of 50% of the scoring bonus.
[...]

PR #524, opened 2019-01-27
New scenario for the Farewell task.

syedaraza:
An alternative scenario with simplified rules and more doable tasks. In the context of comments here https://github.com/RoboCupAtHome/RuleBook/issues/496#issuecomment-457905048

### DIFF
--- a/tasks/Farewell.tex
+++ b/tasks/Farewell.tex
@@ -1,41 +1,38 @@
 \section{Farewell [Party Host]}
-It's raining outside. The robot takes tired guests to their cab/über. Ladies go first.
-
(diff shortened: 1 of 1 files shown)


kyordhel:
@syedaraza The changes you introduced are too big, hence producing a brand new test.

I'm closing this PR to keep the *Farewell* task as is, but at the same time I would like to ask you to craft a new task with these ideas, and send a PR for it.

Let's remember that since 2019 teams will chose which tasks they will solve, so instead of modifying a task to make it SSPL compliant, we better design a task for the league (other leagues are also welcome to choose it). The more tasks we have, the better.

Issue #356, opened 2017-08-01
New Object Recognition Test: What is this?


swachsmu:
If we concentrate in Storing Groceries on the manipulation part, there would be space for another test concentrating on visual recognition and human-robot interaction: "What is this?"

The general idea is that a human is walking through the apartment and is pointing to various things (annotated places, furniture, objects, doors, etc.). The robot needs to recognize the pointing gesture, should point at the same thing, and says what it is. The task includes the robot needs to position hinself so that it sees and is able to evaluation the pointing. It needs to follow the human through the apartment.

The Pepper would be an ideal platform for repeating the pointing. The recognition of gestures seems to be there based on e.g. open pose. If the audience hears what the robot is saying, it would be an easily understandable task, which could be scaled from simple to very complex.

balkce:
And if there is no pointing capabilities, the person could always show the
robot the object by putting it up in his/her hand, which has its own set of
issues (blurriness from movement).

LoyVanBeek:
Could also be part of an extension of Speech&Person recognition i'm thinking about (would need to be renamed to 'Question Answering' though): just ask the robot questions about the current environment, eg. "What color is the thing on the shelf", "Name this thing" (while pointing) and the robot goes to find out.

kyordhel:
@airglow, @awesomebytes, @HideakiNagano21, @justinhart, @mnegretev, and rest of TC: your thoughts, comments and proposals, please. We also need a volunteer to script this.

kyordhel:
No major changes in the rulebook for 2019. Rescheduling for 2019.

kyordhel:
@moriarty @justinhart This sounds somehow familiar. Shall we design a variant of a task or this issue is just closed?

johaq:
Closed. Some aspects incorporated into Hand me That test.

PR #526, opened 2019-01-27
Add the Find my Disk task

kyordhel:
This task is full HRI through dialog-based object recognition

## Find My Disk [Housekeeper] (Stage 2)
The robot helps a blind person to locate a LP, compact disk, or audio-cassette (hereinafter a _disk_) in a shelf. The operator will ask the robot to describe what it sees in either the shelf or the media cover, but won't allow the robot to touch her treasures.

### Focus
Human-Robot dialogs based on object recognition and perception.

### Main Goal
The robot shall provide accurate visual descriptions of a shelf and as many objects as it gets presented until the desired object is found.

### Highlights
- Full HRI
- No manipulation (none at all)
- Operator is blindfolded
- Provides labeled data recording for HRI and NLP


tasks/FindMyDisk.tex
```diff
@@ -0,0 +1,88 @@
+\section{Find My Disk [Housekeeper]}
+The robot helps a blind person to locate a LP, compact disk, or audio-cassette (hereinafter a \textit{disk}) in a shelf. The operator will ask the robot to describe what it sees in either the shelf or the media cover, but won't allow the robot to touch her treasures.
```
moriarty: How will the disks be on the shelf? face out or spine out?

Or a mix of both and the robot needs to ask the `blind person` to flip to the next `disk`?


1. https://www.amazon.com/Hudson-Hi-Fi-Mount-Record-Display/dp/B071WKY96X
2. https://www.amazon.com/Way-Basics-2-Shelf-Storage-Espresso/dp/B075M53XKB
3. https://www.amazon.com/Vinyl-Record-Storage-Rack-Eco-Friendly/dp/B074JG9X1M/

tasks/FindMyDisk.tex
```diff
@@ -0,0 +1,88 @@
...
+% %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%
+% Setup
+%
+% %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\subsection{Setup}
+\begin{enumerate}
+	\item \textbf{Location:} The test takes place near a bookcase in the living room.
+
+	\item \textbf{Bookcase:} The bookcase contains at least 10 media boxes placed one next to another with no gaps between them. The bookcase door (if any) is open.
```
moriarty: See comment above, spines facing outward vs covers facing outward.

kyordhel:
My initial idea was with **books**, but this is kind of stupid since a blind person can't read unless it's Braille and Braille books have tactile titles.

So I changed it to disks. Think about a collection of DVDs (most likely we will use DVD instead). The robot sees something like this:

image

I'm fine with the robot failing to recognize one DVD from another, but it can guide the person and request the first one (the person can touch and feel the first), then see the cover and describe. I think 10 disks is a reasonable amount.

On the amazon gadgets you suggest. Well, I've nothing against that. The house of a blind person must be somehow blind-accessible, and we can include perhaps one special box so the robot can flip disks and see each one's cover. Whatever we choose must be natural.

benatuts:
This sounds fun but I've probably misunderstood this task because I'm not sure clear about its intent.

In particular, I just tried this with the product image search in Amazon's shopping app as well as Google Lens. It can find an Amazon product page from cover art. So, can the Amazon shopping app with accessibility features turned on get full marks in this challenge? Even without image search, can one get full marks on this challenge by just doing nothing other than "mindless" optical character recognition (OCR) on anything put in front of the robot's camera?

Perhaps it could be made more aligned with robotics by slightly reversing the challenge: the vision impaired person can show the robot objects; the robot will ask questions (setting annotations/tags for the object), and then the robot will watch as the person shelves the object (perhaps out of sight) somewhere in the house. The robot can now assist the vision impaired person to locate objects again.

e.g.,
User: "This object is great for when I'm happy"
User puts object away while robot follows and watches
.... time passes ....
User: "I'm feeling happy"
Robot: "Look inside the dining room cabinet"
User pulls object out of the cabinet
User: "Is this it?"
Robot: "Yes"

kyordhel:
@benatuts 

The task you're proposing sounds interesting, but its focus differs widely from the one proposed here. Therefore, before proceeding to explain this one, I would like to ask you to open a PR with the task you're proposing, so it can be analyzed by our Technical Committee.

The depicted test make sense only to a person with [blindness](https://medlineplus.gov/ency/article/003040.htm). The test is simple:

*I can't see so I ask my robot to tell me how the cover of a disk looks like, so I can find the one I want.*

The **focus** is on computer vision + natural language dialogs. This would be one of the few tests in the history of @Home that focuses more on language generation than in speech recognition and NLP. 

About how you solve it? Well, that's irrelevant here. We propose challenges within the state-of-the-art for our experts to solve. If they succeed using Amazon's or Google's tech, good for them, but I'm 100% sure my grandma's collection is not in the Amazon DB.

benatuts:
Hi @kyordel,

In that case, I think it needs more clarity about what counts as an explanation.

Some thoughts, from my point of view, offered in the spirit of positive support:

- The test appears to not involve the robot at all: there is no need for the system to understand context or location or the human agent. My reading of the test is that a mobile phone on a stick could pass this test.

- The benchmark for what counts as a sufficient explanation is not clear. Would OCR and then text to speech be enough... "That says 'The Beatles'".... "That says 'Hot Hits of the 90s'". If so, then I would suspect that this could be the easiest challenge, or perhaps one that is more dependent on the quality of the robot's cameras rather than anything else.

- It sounds like an image captioning challenge requiring very limited "integration" with other systems on a robot. There are already other captioning competitions, so how does this differ from the existing competitions other than the image source being a robot camera?
https://ai.google.com/research/ConceptualCaptions
https://www.kaggle.com/c/datalabcup-image-caption#description


Again, it may be that I don't understand the challenge and that some extra explanation in the task would more clearly differentiate this task and clarify the role of the robot.

Here's another idea for modifying the task while still retaining a focus on computer vision and natural language dialog:

What if instead of describing CD covers, the challenge is to describe the room? A vision impaired person has entered the house and the robot must guide them to the sofa (or some other object) using words alone. This requires the robot to describe a scene from the user's perspective ("there is a wall in front of you", "mind the table to your left" "take one step right before walking through the door"), and adapt if the user does something different.


Hope that helps,

-Ben

kyordhel:
@benatuts Thanks for the feedback and the time reviewing this PR. I appreciate the observations, but my comprehension is very limited. Unless you have punctual changes or fixes, I'm afraid I can't do any better.

On your main three remarks.
- Yes, a smartphone could solve the test. Nonetheless I don't see how a blind person could use a smartphone nowadays. Perhaps I like of imagination.
- Well yes... and no. Many disks or movies have texts that are impossible to OCR, and object/cover recognition will fail if you don't have a database with all titles. The idea is to have a robot describing stuff it has never seen before. It might be silly and stupid, yes, but we people do it quite often without realizing.
- Indeed. You can replace the robot with a pole, a camera, and speakers. The idea is to introduce dialogues in @Home using very basic computer vision. I don't care much about integration in Stage I. That you can solve it using existing technologies is pretty much irrelevant to me, for I want that robots can use the feature with anything, not only disks.

You know? I used to help my grand father to find the right disk by describing the cover.  I wasn't able to read titles in English or Italian by that time (let alone Cyrillic) and I didn't knew the people, but could describe: *four guys crossing the street, one is barefoot; blonde girl in red dress with a microphone; purple cover with a gray gramophone*. Later on I learned he had all disks ordered so my help wasn't necessary, but he pretended to to share stories about the disks with me. What he taught me is how to efficiently describe the unknown, and that's something that has been with me ever since and helped me along in all my travels. 

Whatever, nobody cares. Romantic nonsense aside. Your idea on room description sounds interesting and somehow overlaps with #356 and might have an impact on restaurant. Since it deviates from the purpose of this PR, I would suggest you to design that test and submit a PR for review.

balkce:
I like the language-generation feel of the test. I also like that the objects are quite specific: books on tape or CD/DVD's with music or movies. This will make its setup quite easy.

As for @benatuts idea, it's sounds really cool, but the intent with that idea seems to be more in the vision/recognition side of things. While this test focuses more on the interaction side (that's why the visual recognition is simple). I suggest @benatuts provide a PR for another test which focuses in environment description. Idea (although it may be a bit complex): a visual-impaired person has rented their apartment to an AirBnb guest, and the robot helps the person find what has changed.

kyordhel:
@balkce @moriarty I'd like to merge this one. Please check if it meets your expectations

PR #527, opened 2019-01-27
Add (Refactored) Restaurant test

### LINKED ISSUE #497: Restaurant [restructure][draft]
## Main Goal

All patrons of a restaurant have been attended to. They have been offered menus, their orders have been taken and their food has been served to them.
Robots are operating in a previously unknown (not mapped!) environment.

## Setup

Ideally this test happens in a real world environment. Such as the Tim Hortons outside of the venue in 2018.

## TC/OC preparation

1. Have someone who speaks the local language and is polite to do the negotiation with the location.
2. DO NOT LET THE TEAMS KNOW WHERE IT WILL HAPPEN.

## TEAMS should note:

1. We don't guarantee WiFi, nor internet access (during any part of the competition).
2. Robots should be able to move to a new location easily and we will be as reasonable as possible but you must keep in mind your robots battery life.

MatthijsBurgh:
I have chose to make the following procedure for taking an order the default:
- Barman instructs robot to take order at table X
- Barman will guide robot to table X

This allows for the following bonuses:
- Robot detects quest requesting service by gesture or calling
- Robot arrives at guest without needing guidance.

No guidance as default is not possible, because if the robot is instructed to take an order at table X, the robot doesn't know where it is. Therefore it will always need guidance. Guidance is not needed by detection by the robot, because the robot can detect the position of the guest. But the robot may still ask for guidance it that case.

So if no guidance should be the default, then detecting the guest should also be default. But in that case. The robot always needs to take two penalties, one for being instructed to take an order at table X and needing guidance  to table X. The double penalty is not preferred in my opinion and therefore I chose the first option.
Fixes #497


tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
+\section{Restaurant}
+The robots are tested in a real environment such as a real restaurant or a shopping mall. The test should be executed without a map at the start.
+There are \emph{two} robots helping clients in the restaurant at the same time.
```
moriarty: Same or different teams?

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+% \subsection{Focus}
+% This test focuses on online mapping, safe navigation in previously unknown environments, gesture detection, human-robot interaction, and manipulation in a real environment.
+
+\subsection{Main Goal}
+Taking two orders from the quests and serving the quests with the requested orders. Before taking the order, the robot can be guided by a \textit{Professional Waiter} (member of the TC). During delivering phase, the robot should finds its own way.
```
moriarty:
```diff
- quest
+ guest
```

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+\noindent\textbf{Reward:} 1000pts, 350 regular per order + 150 bonus per order
```
moriarty: I thought these were standardized at some lower number, unless it was said Restaurant was an exception.

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+\subsection{Bonus rewards}
+\begin{enumerate}[nosep]
+	\item Detection of quest requesting service by gesture or calling (75pts)
```
moriarty: guest... I think you can safely `%s:quest/guest/g` in vim :wink:

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+    \item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per order as follows.
+        \begin{itemize}[nosep]
+          \item Needing guidance during the deliver phase causes a reduction of 40\% in scoring.
+          \item Needing assistance for picking up an item/the tray causes a reduction of 10\% in scoring.
+          \item Have a human place the item in a basket causes a reduction of 20\% in scoring.
+          \item Have the quest take the item out of the basket causes a reduction of 20\% in scoring.
+          \item Asking the barman at which side the \textit{Kitchen-bar} is causes a reduction of 5\% in the total scoring.
+        \end{itemize}
+
+	\item \textbf{Start:} There is no fixed start signal in this test, it starts when both robots are ready (though within a reasonable time).
```
moriarty: No fixed start what so ever? usually we have a queue of Teams and there will be some sort of way to get them to the starting location and have them start

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+	\item \textbf{Order:} The way the user provides information to the robot is up to the robot's team. A natural interaction is preferred.
+
+	\item \textbf{Location:} This test can be arranged in any real restaurant or shopping mall. If this is not possible, the test can be conducted in an arbitrary room containing the appropriate locations.
+	  The only requirement is that this room is not part of the arena and that the teams do not know the room beforehand.
+	  The exact location, including the object and delivery locations, will be defined by the technical committee on site (and in corporation with the local organization).
+	  In addition, to avoid unnecessary time investment for navigation, the distances between tables and the \enquote{Kitchen Bar} will be minimal.
```
moriarty: Suggest adding: `nearby electrical outlets, and wi-fi access should not be expected`

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+	\item \textbf{Disturbances from outside:} If a person from the audience (severely) interferes with the robot in a way that makes it impossible to solve the task, the teams may repeat the test immediately.
```
moriarty: This is hard for us to define `severely`.

tasks/Restaurant.tex
moriarty: Also: something specifying that `the the location will likely be space confined. All RoboCup@Home participants will be expected to follow the instructions of the onsite OC, and may not be able to watch their or other teams performances.`

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+\subsection{OC instructions}
+During Setup days:
+\begin{itemize}[nosep]
+	\item Check with local (security) management if the possible location can be used for the restaurant test.
+\end{itemize}
+
+1 hour before the test:
+\begin{itemize}[nosep]
+	\item Gather all teams and robots to move to the location of the test.
```
moriarty: move teams to some nearby queuing area and instruct the teams how/when to move to the actual restaurant

tasks/Restaurant.tex
```diff
@@ -0,0 +1,90 @@
...
+
+The referee needs to
+\begin{itemize}
+	\item Prepare orders for each client in advance, so that there can be no confusion. These orders must also be available at the kitchen.
+\end{itemize}
+
+\subsection{OC instructions}
+During Setup days:
+\begin{itemize}[nosep]
+	\item Check with local (security) management if the possible location can be used for the restaurant test.
```
moriarty: Also find a queuing area. In Montreal, it was our queue of teams which caused more problems than the test itself.

tasks/Restaurant.tex
MatthijsBurgh: I am not sure what the standard is at the moment. I have seen some different numbers. We can talk about it in the meeting tomorrow.

tasks/Restaurant.tex
MatthijsBurgh: I would say that is preparation. No fixed start like opening a door. (This is just copied from the old Rulebook)

tasks/Restaurant.tex
kyordhel: Severely is at criterion of the TC, and means that the robot cannot continue anymore (e.g. a kid runs and hugs the robot, we had that in the past, Eindhoven I think). Its very rare that this happens.

kyordhel:
@MatthijsBurgh I still need to go through the test carefully, but based on what you explain in the PR comment, I'd like to discuss the following scenario.

*Setup: There are at least 3 group of people with assigned tables and at least two have an order to place.*

*Standard calibration procedure: The robot receives the start signal and is instructed by the Barman where the bar is (i.e. beverages served), where customers are (shouldn't be necessary), and where it should wait. The robot starts at the waiting position looking towards the tables.*

So far, this our setup the robot, so no penalties, bonuses, nothing. When the test starts the robot may:
- a) detect a person waving, go there and take the order
- b) start patrolling and look for people, when a person is found it may ask if they will to order something
- ask the barman assistance:
    - c) Ask from which table take an order
    - d) Ask to be guided there
    - Something else I didn't foresee

From my perspective, (a) should give a bonus, while (b) is just an expected autonomous behavior. On the other hand, (c) and (d) are *Deus ex Machina* and come with a small penalty. The robot shouldn't have a problem taking the order and going back autonomously to the bar, but still it may request the barman's assistance in exchange for further penalty. During the process the map is generated, so the robot should be able to place the order and deliver it. If not, well, more *Deus ex Machina*.

The rest is bonuses.

tasks/Restaurant.tex
moriarty: Yes last meeting my call was dropping in and out, so I didn’t hear the end of that discussion, but after reading through the rules it doesn’t seem standardized yet.

kyordhel: Legacy from 2009-2014 was
- **Stage I** 1000pts
- **Stage II** 2500pts

Difficulty is proportional. We expect teams to score avg 1/3~1/2 available points.

tasks/Restaurant.tex
johaq: Wifi cant be expected is covered in the general rules. I think this is fine as is and agree with Matthijs that we should not overcomplicate the rules in the test description.

johaq:
This looks good to me. I like the change that the robot does not have to ask the barman whether it can take an order anymore and can patrol on its own. Always seemed like an awkward interaction to me. But it still leaves the question what to do when both robots detect a waving person and both arrive at the table at about the same time and want to take the order. Is there a commando to send one of the robots back?

tasks/Restaurant.tex
```diff
@@ -0,0 +1,88 @@
...
+
+	\item \textbf{Order:} The way the user provides information to the robot is up to the robot's team. A natural interaction is preferred.
+
+	\item \textbf{Location:} This test can be arranged in any real restaurant or shopping mall. If this is not possible, the test can be conducted in an arbitrary room containing the appropriate locations. The only requirement is that this room is not part of the arena and that the teams do not know the room beforehand.
+	  The exact location, including the object and delivery locations, will be defined by the technical committee on site (and in corporation with the local organization). In addition, to avoid unnecessary time investment for navigation, the distances between tables and the \enquote{Kitchen Bar} will be minimal. Because the location can be anywhere, nearby electrical outlets and Wi-Fi access can't be guaranteed.
+
+	\item \textbf{Disturbances from outside:} If a person from the audience (severely) interferes with the robot in a way that makes it impossible to solve the task, the teams may repeat the test immediately.
```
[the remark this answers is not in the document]
kyordhel: Ask the barman?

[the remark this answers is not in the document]
johaq: Difficult when both robots are patrolling the restaurant. Should they go back and ask? That would again seem pretty weird and unnatural to audiences. Which waiter goes back to the bar and asks if they are allowed to take an order.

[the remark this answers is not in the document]
kyordhel: I think the patrolling idea was discarded in the past meeting. Basically the robot ask the barman what to do and if it can, the barman takes over the test (guide robot to table, take order for robot, guide robot back, dispatch order, deliver order). The more work the barman does, the higher the chances your robot scores zero and gets fired.

tasks/Restaurant.tex
MatthijsBurgh: Eindhoven loves robots 😍

tasks/Restaurant.tex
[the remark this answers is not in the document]
MatthijsBurgh: Yep pratrolling is discarded.

johaq:
Removed that two robots are in the task at the same time. 
Was always more difficult for referees and does not really fit with new slot structure.

Otherwise I changed the points a bit and reformulated some things.

kyordhel:
image

Issue #508, opened 2019-01-14
Finals [restructure][draft] 


awesomebytes:
As discussed in the meeting from 14/01/19 we need to agree on what to do with the finals.

The two proposals that came up are:

* Do a EEGPSR scenario.
* Keep the format of previous finals.

This is related to [About removing the finals and open challenge](https://github.com/RoboCupAtHome/RuleBook/issues/484)

balkce:
Several ideas have been discussed for this.

I ask of you to discuss the following ideas. By Friday, February 1st, we need come to a decision. I'll write it up in the rule book that day.

1. Leave the finals as they are right now. Pros: since the Open Challenge have been removed, this is the only test that officially endorses an open demonstration. Cons: the finals have been an issue since I can remember because it is difficult for the audience (and even the jury) to ascertain which is the better robot.

2. Rework the finals such that they have a "theme" in which the two teams need to show off something in terms of that "theme". The theme should be small and easily understandable. For example: "navigation", "object manipulation", etc. The idea is to have three "themes" that are decided upon (when, how, who decides this are still up for debate). Robot A enters the arena shows off a task related to theme 1, Robots B enters and shows off a task related to theme 1, Robot A shows off related to theme 2, etc.

   a. The teams choose the themes. Pros: they can show off their research. Cons: conflicts can arise if a non-gentleman team chooses a theme just to disrupt the other team. It is also difficult logistically, since the teams won't be able to prepare their finals beforehand.
   b. The TC chooses the themes. Pros: lets the TC dictate what they want the league to be each year, and lets the team prepare their finals beforehand. Cons: won't let the teams show off their own research.
   c. The TC chooses two themes, leaving the third to be an open demonstration. Pros: a bit of both worlds. Cons: we are still inheriting the issues of the old version of the finals, although at a lesser degree.

The are more variations to this, and these two ideas surely aren't the only ones that can be discussed.

When commenting please consider that the Finals are more or less the face of the league to the outside world (it is the most watched test of our competition), and that this year we are aiming for test clarity and successful robots.

balkce:
Since nobody has commented so far, I'm letting you know that I'm tending to go with option 2c, with the themes for OPL and DSPL being navigation/object manipulation, and for SSPL being navigation/person interaction.

I'm willing to be convinced otherwise.

I'll begin doing these changes tomorrow (Feb. 1) at 12 PM Mexico City local time.

MatthijsBurgh:
[...]

PR #537, opened 2019-02-01
Changed Finals to a themed demonstration.

balkce:
- Change the Finals to a theme presentation paradigm, discussed in #508 .
- Removed the bronze competition. Now the 3rd place is the team with the highest point that did not made it to the finals.
- Removed team-based jury, and left internal jury only as EC.
- Left scoring the same (internal: 25%, external: 25%, Stage I/II: 50%)


balkce:
Done.

Some thoughts:

- As for limiting the microphone time: I rather not put any numbers. If they spend too much time with the microphone with the robot doing nothing, it will impact their scoring (I believe that is enough incentive). We can also tell the judges that they are free to comment if they don't see anything happening. What I don't want to happen is to produce confusion and make the teams believe that they are not allowed to talk: I'm fine with the team narrating what's happening, as long as the robot is doing something.
- I don't understand what you mean when you say that you're "missing the formula to calculate scoring". Do you mean the percentages in the subsection "Scoring"?

kyordhel:
@balkce Never mind, my bad. I wrongly thought outliers were also removed in Finals as they are in the Open Challenge.

This is ready to merge.

PR #538, opened 2019-02-02
Incentivize quality over quantity

benatuts:
This is one possible way to incentivize teams producing high-quality, reliable and excellent solutions rather than a scatter-gun approach of trying to solve every problem.

Teams will want to earn bonus rewards on a single problem, rather than doing the bare minimum across many many tasks.

This is related to the discussion in https://github.com/RoboCupAtHome/RuleBook/issues/508#issuecomment-459926511

### DIFF
--- a/general_rules/Organization.tex
+++ b/general_rules/Organization.tex
@@ -122,7 +122,8 @@ In case of having no considerable score deviation between a team advancing to th
 \label{rule:score_system}
 Each task has a main objective and a set of scoring bonuses.
 To score in a test, a team must successfully accomplish the main objective of the task; bonuses are not considered otherwise.
-Overall scoring is calculated as the sum of the maximum score obtained in each ability.
+
+The overall score is calculated as the sum of the maximum scores obtained in each thematic scenario of each stage. For example, if a team scores 500 in \textit{Clean Up [Housekeeper]}, 400 in \textit{Take Out the Garbage [Housekeeper]} and 800 in \textit{Farewell [Party Host]}, their total score is $max([500, 400]) + max([800]) = 1300$ for Stage I.
 
 The \iaterm{score system} has the following constrains
 \begin{enumerate}


kyordhel:
This change seems to make sense, and is specially helpful for well-intentioned newcomers. Nonetheless it will present a severe disadvantages when introducing experienced competitors into the equation.

1. A team will use all possible attempts to maximize score in one single task that it has practiced to master. On perfection (or close to) will attempt its second best.
2. The former will lead to experienced competitors to bet for the unbalanced easy test (always one escapes to the TC) in which a rule, score, or objective is easily exploitable.
3. As consequence, might have all teams trying to solve one single test, leaving the rest of the rulebook unexplored.
4. With Deus Ex Machina, we might have a circumstance in which a robot commands a human to solve a task and still score. Therefore, a state machine controlling speech generation would be used to easily farm points from an ill-designed test.

In the past, we tried to balance simple tests with multiple repetitions to discard luck, foster benchmarking, and test both, features and integration. That led to exhaustion of the teams and a direct request from the Federation to reduce the teams workload. In our experience:
- Remove repetitions reduces workload but introduces a luck factor that a restart can't compensate.
- Removing feature-testing (all GPSR) would lead to complex tests with more fail ratio
- Removing full integration testing makes teams fail in Stage II since systems are not properly integrated.

In consequence, we are trying in 2019 to keep both types of tests and reduce the workload by giving teams the chance of choose their own repetition/diversity ratio. This is not perfect.

justinhart:
I don't really like this change. It essentially allows teams to pick a subset of rounds to compete in and then gives the win to whoever does the best in a single round. Honestly, if you have a single round that you're doing all that well in to begin with, then you're probably ahead of the field anyway.

balkce:
I understand the idea behind this modification: to reward robots being able to do something well and in a repeatable manner, instead of rewarding robots that do lots of things in a mediocre manner. However, I don't believe that this modification achieves this idea, since it is being applied in Stage II which is a whole other ball game.

If anything, I think that a modification like this would better serve its purpose in Stage 1, since it is in this stage where the evaluation of specific modalities (navigation, manipulation, etc.) is carried out. In fact, that is one of the purposes of Stage I: simple tests to evaluate one or two modalities. You could almost say that the idea behind your modification is already more or less carried out in Stage 1, since there are different tests that can be solved by using the same modality, and the team is rewarded if they solve all of those tests.

The purpose of Stage II, on the other hand, is to integrate all of those modalities to solve a much more complex task. If a team did badly in Stage II, but did well in Stage I, then it means that is now time to start working in improving other modalities and in integrating them with the modalities that already work well to solve those much harder tests.

I understand where you are coming from. How about modifying the way the whole of Stage I is scored as to better fulfill the idea you're proposing?

balkce:
I believe that this PR and #539 are quite related.

I would suggest removing the use of "maximum" and start using "average of". The idea of repetitions (and what I would like as a consumer) is to not only reward robots that do something well, but also those that do so _repeatedly_.

Going for maximum, makes a lucky round be considered representative of the robots performance, where it could be just a blip. Additionally, it does not reward a team that consistently shows that their robot does something well (which is related to the point brought up in #539 ).

What do you think of instead of maximum, use the average of the test carried out in each thematic scenario? Meaning:

The overall score is calculated as the sum of the averages of the scores obtained in each thematic scenario of each stage. For example, if a team scores 500 in _Clean Up [Housekeeper]_, 400 in _Take Out the Garbage [Housekeeper]_, 800 in _Farewell [Party Host]_, and 100 in _Receptionist [Party Host]_, their total score is **avg([500, 400]) + avg([800,100]) = 950** for Stage I.

Just a thought. It may be that it is very similar to just adding them all up.

kyordhel:
Dismissed  by the TC

PR #539, opened 2019-02-02
Reduce breadth of rulebook

benatuts:
This is one possible way to reduce the breadth of the competition and ensure that teams are set up to succeed and look good.

A rule of this kind might be removed in 2020 or a future year when the league has reached the point that most teams are able to reliably achieve over 50% on at least half the Stage I tests.

This is related to the discussion in https://github.com/RoboCupAtHome/RuleBook/issues/508#issuecomment-459926511

Even if this particular proposal is not popular, I do want to raise attention to the need for setting teams up to succeed and look good to create a healthy, competitive and welcoming community, rather than every test needing to be a stretch goal.

A look at last year's results for standard platform leagues suggests that even achieving basic stage I competencies remains a challenge. It shouldn't be the case that hard-working and talented researchers from universities across the world are getting low or zero scores in many categories. This is perhaps caused by hardware limitations in the SPLs. e.g., Pepper's sensors are limited so that effective sensing in "real world" scenarios remains a difficult challenge.

### DIFF
--- a/general_rules/Organization.tex
+++ b/general_rules/Organization.tex
@@ -29,6 +29,11 @@ The \iaterm{Housekeeper} scenario features tasks related to cleaning, organizing
 
 In case of having no considerable score deviation between a team advancing to the next stage and a team dropping out, the TC may announce additional teams advancing to the next stage.
 
+\subsection{Stage II for SSPL and DSPL}
+
+SSPL and DSPL teams may attempt Stage I tests in Stage II, in addition to any of the Stage II tests.
+
+As Stage I tasks have a maximum reward of 1000 points, if those tasks are used in Stage II, their rewards (and penalties) are doubled.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Schedule}


benatuts:
> Why not open it to OPL? I don't think any SPL is inferior so requires special baby-sitting.

Because I checked the Stage I scores for all three leagues and it was clear that Stage I still presents challenges for the majority of teams in the SPLs but not for OPL. It may be due to the leagues being new and still maturing or it may be due to the more challenging design space that SPLs must work in (i.e., they can't use better sensors). This has nothing to do with inferiority.


Ultimately, I believe that the breadth of rulebook is the underlying problem causing teams to be spread too thin, resulting in miserable faces, some Stage I tests where no team gets even of the 13% of the possible points, and finals prepared at the last-minute. To reduce the breadth, you need to actually be _willing_ to reduce the breadth, and resist the temptation to offer giant carrots which have good intentions but only incentivise teams to spread themselves thin.

Consider [2018 SSPL stage I results](https://github.com/RoboCupAtHome/Montreal2018/blob/master/Scores/SSPL%20-%20Stage%201.pdf), and keep this standard in mind when evaluating how things might play out under the new rulebook (where there are almost many challenges and even higher hurdles for success).

Suppose 2018 had been a Stage-I-only competition for SPLs. It may very well have been a more exciting, polished competition where teams would be showing off successes rather than constant failure. I believe that kind of dramatic scope reduction and/or rethinking of assumptions is needed here.

general_rules/Organization.tex
```diff
@@ -29,6 +29,11 @@ \subsection{Stage system}\label{rule:stages}
 
 In case of having no considerable score deviation between a team advancing to the next stage and a team dropping out, the TC may announce additional teams advancing to the next stage.
 
+\subsection{Stage II for SSPL and DSPL}
+
+SSPL and DSPL teams may attempt Stage I tests in Stage II, in addition to any of the Stage II tests.
+
+As Stage I tasks have a maximum reward of 1000 points, if those tasks are used in Stage II, their rewards (and penalties) are doubled.
```
maximest-pierre: You should only be able to get half of the possible points if you do a stage 1 task in stage 2. I would personally add OPL in there just to be fair for every league.

moriarty:
I disagree with the PR, and agree with Mauricio that the risk here is stalling the league.

Teams should do Stage 1 tasks in stage 1, stage 2 tasks in stage 2.

justinhart:
I don't like this suggestion, agreeing with both Alex and Mauricio, however, I do think that what should basically happen, maybe after the release of the 2019 rulebook, is that we fully develop a roadmap for what we want to accomplish as a league and base all tasking on that.

I really think that some of the stuff that is being suggested has low research value and that it's difficult to keep organizations engaged from that standpoint. If we can show that these tasks all tie back to fundamental research, then the rulebook kind of writes itself.

We build a roadmap of where we want to go, both hardware and software-wise as a league. We pick things that are a challenge because of the research value, and we move forward. We share and publish research results. Every task should essentially be the title of the paper that we publish as a league at the end of it. Moreover, teams should focus on whether or not they're able to get papers out of their participation in RoboCup@Home, not whether they already know how to make their robot succeed at the task.

general_rules/Organization.tex
moriarty: This math just doesn't work. If Stage 1 tasks are used in Stage 2, they should be given 0 points because they're not solving Stage 2 tasks, or they should get some minuscule fraction of points, and some multiplicative factor on the penalties.

If not ZERO, then I would suggest:

Stage I can be used in Stage II for 1/10th of their original point value, and 10x the penalty value...

Basically: you better do it perfectly or just go and solve a Stage II task.

kyordhel:
Dismissed by lack of consensus from part of the TC

PR #541, opened 2019-02-05
Added a scoresheet for the receptionist task.

syedaraza:
A scoresheet is added based on the receptionist task defined in tasks/Receptionist.tex


syedaraza:
Changes as per the comments here https://github.com/RoboCupAtHome/RuleBook/issues/495#issuecomment-461379302

PR #544, opened 2019-02-05
Stickler for Rules Task

johaq:
Proposal to add Party Pooper (my working title) task to Party Host scenario in Stage 2.
This test is designed to test a robots ability for activity recognition.

TODO: 
-Come up with more rules
-Bonus Goals


tasks/PartyPooper.tex
```diff
@@ -0,0 +1,46 @@
...
+
+The referee needs to
+\begin{itemize}
+	\item Instruct party guests on which rules to break.
+\end{itemize}
+
+\subsection{OC instructions}
+During Setup days:
+\begin{itemize}
+	\item Announce which room uests are not allowed to enter.
```
balkce: Typo: "Announce which room GUESTS are not allowed to enter."

balkce:
The test is really cool. The only issue I have with it is that the robot would need to do object recognition (for the garbage), person localization (for the private room), and even a mix of both (drink and shoes).

I would remove the bit about the garbage, so that is focused on person activity.

Also, I would definitely change the title something else less scatological.

balkce:
I apologise. I thought this was a Stage 1 test. In this case, leave the garage but in.

Just change the title.

Rulebook.tex
```diff
@@ -84,7 +84,7 @@ \chapter{Tests in Stage II}
 \input{tasks/SetTable}
 
 \newpage
-\input{tasks/PartyPooper}
+\input{tasks/SticklerForRules}
```
kyordhel: What about *House Cop* or just *Stickler*?

scoresheets/SticklerRules.tex
```diff
@@ -0,0 +1,12 @@
+
+The maximum time for this test is 10 minutes.
+
+\begin{scorelist}
+	\scoreheading{Main Goal}
+	\scoreitem[5]{100}{Recognize guests breaking rules and ask them to stop}
```
kyordhel: Same as in the test file, my suggestion for the scoresheet is:

*Identify and detain all guests breaking the [house] rules*

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
+\section{Stickler for the Rules [Party Host]}
+The robot has to enforce house rules set by the homeowner.
```
kyordhel: *Enforce* I love that verb

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+
+
+\subsection{Main Goal}
+The robot has to recognize when party guests are breaking rules and politely ask them to stop.
```
kyordhel: This is somehow ambiguous to me. Traditionally in @Home recognize means _(Person, Name)_ pairing. My suggestion is:

_Identify and detain all guests breaking the [house] rules_

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
+\section{Stickler for the Rules [Party Host]}
```
kyordhel: What about *House Cop* or just *Stickler*? Just a suggestion.

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+The robot has to enforce house rules set by the homeowner.
+
+
+\subsection{Main Goal}
+The robot has to recognize when party guests are breaking rules and politely ask them to stop.
+
+\noindent\textbf{Reward:} 500pts (100pts per recognized rule breaking).
```
kyordhel: Spelling: rule **breaker**

I think finding 5 misbehaving people for 100pts each is too much. I'd place the same 5 but only two are making a mess, so 250 each. Also you have 3 being nice, so there is space for false positives the robot must avoid

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+
+\subsection{Setup}
+\begin{itemize}[nosep]
+	\item \textbf{Location:} This test takes place in the arena.
+	\item \textbf{Guests:} There are several party guests inside the arena.
```
kyordhel: More description. If you have 5 people breaking the rules, you need at least 6 people in the arena.

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+
+
+\subsection{Main Goal}
+The robot has to recognize when party guests are breaking rules and politely ask them to stop.
+
+\noindent\textbf{Reward:} 500pts (100pts per recognized rule breaking).
+
+\subsection{Setup}
+\begin{itemize}[nosep]
+	\item \textbf{Location:} This test takes place in the arena.
```
kyordhel: Location must make clear if the test takes place in the *whole* arena or just in a predefined room

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+	\item \textbf{Guests:} There are several party guests inside the arena.
+\end{itemize}
+
+
+\subsection{Additional rules and remarks}
+\textbf{House Rules:}
+\begin{enumerate}[nosep]
+	\item No shoes allowed inside the house. All guests have to take off their shoes near the entrance.
+	\item No guests are allowed to enter one of the rooms in the house.
+	\item Guests are not allowed to drop garbage on the floor.
```
kyordhel: drop → throw | leave

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+	\item All guests have to have a drink in hand at all times.
```
kyordhel: have to have → must have

You drunk people!

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+
+The referee needs to
+\begin{itemize}
+	\item Instruct party guests on which rules to break.
+\end{itemize}
+
+\subsection{OC instructions}
+During Setup days:
+\begin{itemize}
+	\item Announce which room guests are not allowed to enter.
```
kyordhel: Announce **in** which room **the**

tasks/SticklerForRules.tex
```diff
@@ -0,0 +1,46 @@
...
+	\item Guests are not allowed to drop garbage on the floor.
+	\item All guests have to have a drink in hand at all times.
+\end{enumerate}
+
+\subsection{Referee instructions}
+
+The referee needs to
+\begin{itemize}
+	\item Instruct party guests on which rules to break.
+\end{itemize}
```
kyordhel: You're missing:
- Assign/provide drinks to each guest (that can be done by OC)
- If the robot needs to recognize people, names must be assigned

tasks/SticklerForRules.tex
johaq: I think rule breaking is fine here, both work. 
Addressed the points in commit.

tasks/SticklerForRules.tex
johaq: no recognizing people needed I think

johaq:
Addressed some (I hope most) of the points.

johaq:
What do you mean by compact version? Do you think the house rules should be separate in the general rules (like garbage bag description etc.) or just more concisely expressed?

Don't we have partial scoring in some of the other tasks (pretty much every task in stage 1 right now)?

kyordhel:
1. Compact version = one-liners. No example, no explanation.
2. Yes, we do. When we have partial scoring we must add the additional rule (to prevent further complaints).

johaq:
Right now it is one sentence for the rule and one for how to enforce it. Should I remove the how to enforce it part to let teams come up with their own ideas?

PR #547, opened 2019-02-06
General Rules Fine Tuning

johaq:
I went over the general rules. These are mostly grammar fixes, formulating things differently and removing duplicate or inconsistent information.

Other addressed issues:
- We are missing institutions for: Angelica Lim, Allison Henry and Agnes Delaborde. Would be nice to add or to know if they do not want an associated instituion
- Rulebook states team website is REQUIRED. Is that really the case? Last time I checked I did not find websites for every team. Maybe should be changed to encouraged or positive for qualification process.
- I changed the skill awards to be in line with the new rulebook concept
- Removed location classes. I think I've never seen sofa be assigned to the "seat" class or it be relevant in any task. But I might be wrong here.
- Removed teams must announce external computing devices one month in advance since it was never enforced
- Section 3.6 in the current form requires every task to be 5 minutes (or shorter). This is in conflict to 3.7.14 where it states Stage 2 tasks are 10 minutes long. Need a decision and make this consistent.


general_rules/Organization.tex
```diff
@@ -110,34 +110,33 @@ \subsection{Schedule}
 	\end{tabular}
 
 	\caption{Example schedule.
-		Each team has assigned at least two test slots in every block.
-		At least two blocks are scheduled per day with an assigned theme.
-		A team can choose a different task in each test, meaning at least 4 different tests per stage.
+		Each team has two test slots in every block.
+		Two blocks are scheduled per day with an assigned theme.
+		A team can choose a different task in each test, meaning at most 4 different tests per stage.
```
balkce: Is this correct ("at most 4")? I would think that, given the repetition of tests in each block, a team could carry out a different test in each repetition if they wanted to.

kyordhel: No, this is wrong.

2 blocks per scenario (2) with **at least** 2 attempts each means **at least** 4 tests per day, and at least 8 tests in Stage I.
Stage II might be less, as usual

johaq: Well I took it as this description specifically referring to the example given in the graphic which would mean at most 4 tests. The at least thing is I think already mentioned in the more general description before a specific example is made

johaq:
Bump. I think I addressed the request and this could be merged

kyordhel:
@balkce You requested changes. Please give it a look.

CompetitionConcepts.tex
```diff
@@ -18,23 +18,23 @@ \section{Autonomy \& Mobility}
 
 Furthermore, the specific tasks must not be solved using \emph{open loop control}.
 
-\section{Aiming for applications}
+\section{Aiming for Applications}
 \label{concept:aiming_for_applications}
-To foster advance in technology and to keep the competition interesting, the scenario and the tests will steadily increase in complexity. While in the beginning necessary abilities are being tested, tests will focus more and more on real applications with a rising level of uncertainty. Useful, robust, general, cost effective, and applicable solutions are rewarded in RoboCup@Home.
+To foster advances in technology and to keep the competition interesting, the scenario and the tests will steadily increase in complexity. While in the beginning necessary abilities are being tested, tests will focus more and more on real world applications with a rising level of uncertainty. Useful, robust, general, cost effective, and applicable solutions are rewarded in RoboCup@Home.
```
kyordhel: Not sure about this change, I'd add a native speaker to triple check

CompetitionConcepts.tex
```diff
@@ -18,23 +18,23 @@ \section{Autonomy \& Mobility}
...
 
-\section{\iterm{Social relevance}}
+\section{Social Relevance}
 \label{concept:social_relevance}
-The competition and the included tests should produce socially relevant results. The aim is to convince the public about the usefulness of autonomous robotic applications. This should be done by showing applications where robots directly help or assist humans in everyday life situations. Examples are: Personal robot assistant, guide robot for the blind, robot care for elderly people, etc. Such socially relevant results are rewarded in RoboCup@Home.
+The competition and the included tests should produce socially relevant results. The aim is to convince the public of the usefulness of autonomous robotic applications. This should be done by showing applications where robots directly help or assist humans in everyday life situations. Examples are: Personal robot assistant, guide robot for the blind, robot care for elderly people, etc. Such socially relevant results are rewarded in RoboCup@Home.
```
kyordhel: Again, requires native speaker

Setup.tex
```diff
@@ -44,7 +44,7 @@ \section{General Setup}
 
 \section{Welcome Reception}
 \label{sec:welcome_reception}
-Traditionally --since Eindhoven 2013-- the RoboCup@Home holds an own \iterm{welcome reception} in addition to the official opening ceremony. During the welcome reception, a \iterm{poster session} is held in which teams present their research foci and latest results (see~\refsec{sec:poster_teaser_session}).
+Traditionally --since Eindhoven 2013-- the RoboCup@Home holds their own \iterm{welcome reception} in addition to the official opening ceremony. During the welcome reception, a \iterm{poster session} is held in which teams present their research foci and latest results (see~\refsec{sec:poster_teaser_session}).
```
kyordhel: *its* would be most correct I think, but their make sense since there are several leagues. I don't know.

general_rules/ExternalDevices.tex
```diff
@@ -3,35 +3,34 @@
 % External Devices
 % 
 % %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\section{External devices}
+\section{External Devices}
 \label{rule:robot_external_devices}
-Everything which is not part of the robot is considered an \iterm{external device}.
+Everything not part of the robot is considered an \iterm{external device}.
```
kyordhel: This one I know

Either 
- Everything not being
- Everything which is not

general_rules/Procedure.tex
```diff
@@ -181,13 +181,13 @@ \subsection{Operators}
 
 \subsection{Moderator}
 \label{rule:moderator}
-The LOC is responsible of organizing test moderation in the local language.
+The LOC is responsible for organizing test moderation in the local language.
```
kyordhel: I think *of* is correct here

general_rules/Procedure.tex
```diff
@@ -181,13 +181,13 @@ \subsection{Operators}
...
 The OC may request the participating teams to provide a team member for moderation.
 Candidates have to be fluent in the moderation language (default is English).
 
 \noindent\textbf{Responsibilities:} The moderators have to:
 \begin{compactitem}
-	\item Do \textbf{NOT interfere} with the performance
+	\item \textbf{NOT interfere} with the performance
```
kyordhel: *Do NOT* is more correct. Emphasizes

johaq: I was thinking in context of the full sentence here:
"The moderators have to NOT interfere..."

For the rest, let's wait for a native speaker.

kyordhel: Yep, I got your intention, but as when you have this signs with 2 words you write:

- No Smoking
- Do NOT smoke in this area

The *do* adds three tons of stress. On the rest, I agree in having an expert in English Grammar =P, I often make stupid mistakes (e.g. with in/on/at which all map to *en* in Spanish).

CompetitionConcepts.tex
moriarty: Looks good to me.

Setup.tex
moriarty: `its` Or `their` are both better than the original “an own”.

We have several sub leagues but only one OC/TC/EC so it is difficult to say what is correct here.

The RC@Home leagues hold their own.
The RC@Home league holds its own.

general_rules/ExternalDevices.tex
moriarty: I’d accept the change as is. “Everything not part of the robot...”

general_rules/Procedure.tex
moriarty: For is correct.

general_rules/Procedure.tex
moriarty: You don’t need the “do” if you want to add stress you can replace “have to” with “must”

general_rules/Organization.tex
balkce: Seems good to me now.

CompetitionConcepts.tex
balkce: Seems good to me.

Setup.tex
balkce: I'm fine with either. Accepting change.

general_rules/Procedure.tex
balkce: The native has spoken. Resolving.

PR #548, opened 2019-02-07
Added a task called Carry my Luggage in Stage I.

syedaraza:
Created this test in response to a previous PR #524 

This is similar to the Help Me Carry task of previous years in that the robot has to follow the person outside and come back.

Any help will be appreciated in making it more clear.


kyordhel:
@syedaraza 
1. Please add your name to your GitHub account. We'd like to know who's contributing.
2. Please enhance the description (i.e. TL;DR version)

scoresheets/HelpMeCarry.tex
```diff
@@ -1,52 +1,20 @@
...
-\ifSSPL{%
-	\scoreitem{30}{[SSPL] Asking a person to step aside (\textit{smart} obstacle)}
-}%
-\ifNotSSPL{%
-	\scoreitem{30}{[DSPL \& OPL] Moving away movable object}
-}%
-
-	\setTotalScore{200}
+	\scoreheading{Main Goal}
+	\scoreitem{500}{Carry the bag to the car's location and coming back home}
```
kyordhel: Take one of the bags from the car to the livingroom

scoresheets/HelpMeCarry.tex
```diff
@@ -1,52 +1,20 @@
...
+	\penaltyitem{-100}{help in finding the bag}
+	\penaltyitem{-100}{help in carrying the bag}
+	\penaltyitem{-200}{help in tracking the guest}
+	\penaltyitem{-100}{help in going back home}
+	
+	\scoreheading{Bonus rewards}
+	\scoreitem{100}{Find the bag by understanding the pointing behaviour}
+	\scoreitem{400}{Avoid all the obstacles outdoor}
```
kyordhel: Split per obstacle

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
+\section{Help Me Carry [Party Host]}
+The robot helps a guest to carry some luggage to his/her car.
```
kyordhel: - ~a guest~ → the operator
- ~his/her car~ → the cab parked outside

Rulebook.tex
```diff
@@ -70,6 +70,8 @@ \chapter{Tests in Stage I}
 \input{tasks/ServingDrinks}
 \newpage
 \input{tasks/FindMyMates}
+\newpage
+\input{tasks/HelpMeCarry}
```
kyordhel: Rename

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
+\section{Help Me Carry [Party Host]}
```
kyordhel: # Rename!

We have been running *Help Me Carry* in a certain way for years. Changing the entire test content and keeping the name will confuse teams. Either
- Have the robot taking groceries from the car to the kitchen
- XOR Change the name so people won't get confused

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+The robot helps a guest to carry some luggage to his/her car.
+
+\subsection{Main Goal}
+Robot helps a guest to carry a bag to the car which is parked outside and comes back home.
+
+\subsection{Focus}
+It includes understanding pointing gesture, object identification and manipulation, navigation in known and unknown environments, person-following while carrying a luggage, and social navigation.
```
kyordhel: Covers everything

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+	\item Find the bag by understanding the pointing behaviour (100 pts).
+	\item Avoid all the obstacles outdoor (400 pts).
+\end{enumerate}
+
+
+\subsection{Setup}
+\begin{itemize}
+	\item \textbf{Location}: The test takes place inside and outside the arena.
+
+	\item \textbf{Start Location}: The robot starts at a predefined location in the living room.
```
kyordhel: Not living room. It will be announced by OC based on arena configuration

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+
+	\item \textbf{Guest}: There is only one guest in the house, standing in front of the robot at the start.
```
kyordhel: The operator is standing in front of the robot

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
+\section{Help Me Carry [Party Host]}
+The robot helps a guest to carry some luggage to his/her car.
+
+\subsection{Main Goal}
+Robot helps a guest to carry a bag to the car which is parked outside and comes back home.
```
kyordhel: ~a guest~ → the operator

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+
+\subsection{Setup}
+\begin{itemize}
+	\item \textbf{Location}: The test takes place inside and outside the arena.
+
+	\item \textbf{Start Location}: The robot starts at a predefined location in the living room.
+
+	\item \textbf{Guest}: There is only one guest in the house, standing in front of the robot at the start.
+	
+	\item \textbf{Bags' Location:} There are at least two bags placed between one to two meters  away from the guest's start position and both should be visible to the robot.
```
kyordhel: ~guest~ → operator

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+	
+	\item \textbf{Pointing to the Bag:} When the test starts, the guest points to his/her bag and requests the robot to carry it to the car.
```
kyordhel: - ~his/her~ → a
- ~guest~ → operator
- ~carry~ → take

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+	
+	\item \textbf{Outdoor Obstacles:} The robot will face multiple obstacles outside. The obstacles may include a person, a group of people, a stationary object or a moving object.
```
kyordhel: How many? Important to keep consistency. Avoiding a standing person is solved and shouldn't grant any points, while going through a crowd was typically awarded with 300.

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+\end{itemize}
+
+\subsection{Additional rules and remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied as follows.
+	\begin{itemize}
+		\item help in finding the bag (-100 pts).
```
kyordhel: How?

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+		\item help in carrying the bag (-100 pts).
+		\item help in tracking the guest (-200pts).
+		\item help in going back home (-100pts).
```
kyordhel: How?

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+	\end{itemize}
+
+	\item \textbf{Pointing to the Bag:} The guest keeps pointing to the bag and does not move.
```
kyordhel: ~guest~ → operator

> *The operator stands still while pointing at the bag*

tasks/HelpMeCarry.tex
```diff
@@ -0,0 +1,76 @@
...
+		\item help in carrying the bag (-100 pts).
+		\item help in tracking the guest (-200pts).
+		\item help in going back home (-100pts).
+	\end{itemize}
+
+	\item \textbf{Pointing to the Bag:} The guest keeps pointing to the bag and does not move. 
+
+	\item \textbf{Car Location:} There is no car outside. Instead, a fixed location is supposed as a car location outside the arena.
+
+	\item \textbf{Reaching the Car:} The robot can reach the car's location only by following the guest. After the robot reaches the car's location, the guest says "Stop following me".
```
kyordhel: ~guest~ → operator

~After the robot reaches the car's location, the guest says "Stop following me".~ It is up to the robot to instruct the operator. We don't hard-code follow-me since 2009

scoresheets/HelpMeCarry.tex
syedaraza: Ok then it will better to have a fixed number of obstacles, say three, with 100pts for avoiding each.

tasks/HelpMeCarry.tex
syedaraza: A personal car parked outside makes more sense because the robot does not know the car's location and it must follow the person. Otherwise, the robot can find the cab outside and go it without following the person.

tasks/HelpMeCarry.tex
syedaraza: Carry My Luggage sounds good. How about Roustabout or Drudge?

tasks/HelpMeCarry.tex
syedaraza: Then we should remove Focus section

tasks/HelpMeCarry.tex
syedaraza: I think carry sounds more meaningful. You can take something out by using a trolley or by pulling it.

tasks/HelpMeCarry.tex
syedaraza: Ok then how about avoiding three obstacles 1) a crowd obstructing path 2) a small 3d object 3) a small toy car moving randomly in robot's path?

tasks/HelpMeCarry.tex
kyordhel: Not necessarily, just shorten it.

scoresheets/HelpMeCarry.tex
kyordhel: Depends on the nature of the obstacle. As responsible it is your job to analyze the technical difficulty of an straightforward solution and assign a reward proportional to other skills in the rulebook.

The number of obstacles is irrelevant as long as its position remains unknown to the robot.

tasks/HelpMeCarry.tex
kyordhel: then *the car* would do.

tasks/HelpMeCarry.tex
kyordhel: *Carry My Luggage* sounds quite descriptive to me. Let's consider that most participants have a B2-equivalent English proficiency, so uncommon words should be avoided.

tasks/HelpMeCarry.tex
syedaraza: Then alternatively,

Reaching the Car: The robot can reach the car's location only by following the operator. The robot should instruct the operator how to start and stop following.

tasks/HelpMeCarry.tex
syedaraza: Let's remove the penalty for not finding the bag autonomously because that's a bonus task. Similarly, help in going back to home should not be penalized because as per Section 3.8.2 no bonus points are scored in case of any help. 

Besides, help in carrying the bag can be addressed using the DeM rules in Section 3.8.2 of the rulebook. So we can also drop it from the DeM list here.

For help in tracking the operator there can be many ways and listing all of them seems infeasible. I can think of, robot can request to wave, request to call its name loudly, stand in front of it, press a button on its screen, hold its hand and drag him along, push him from back in the right direction, etc. So to keep it simple we can have a single rule to assign a fixed penalty for any help in tracking the operator. And since tracking the person is the most important component of this task let's have a heavy penalty for that.

What would you say based on your previous experiences? How did people work around if they got stuck in person following tasks?

tasks/HelpMeCarry.tex
syedaraza: Ok I will try.

tasks/HelpMeCarry.tex
kyordhel: Sounds good. Five obstacles were used in the past (3 from Follow Me, 2 from Navigation), but standing people and people walking by are kind of solved.

You can come out with some sort of new obstacle we haven't addressed before (e.g. traffic light). My only requirements are:
- Unspecified order
- Brief, concise description
- Nature and scoring clearly specified for each
- Special rule for collision if necessary

tasks/HelpMeCarry.tex
kyordhel: > Let's remove the penalty for not finding the bag autonomously because that's a bonus task.

Agreed.

> help in going back to home should not be penalized [...] no bonus points are scored in case of any help.

A partial bonus can be awarded if the robot returns home with assistance (e.g. lost operator who needed to stop and wave). That's preferable over not returning home but shall score less than returning without losing track of the operator.

> carrying the bag can be addressed using the DeM [...] So we can also drop it [...]

Agreed. Nonetheless it would come handy to have the handover score reduction in the scoresheet.

> listing all of them seems infeasible

Agreed, how it is done is irrelevant IF THE OPERATOR DOES NOT TOUCH THE ROBOT. Waving, calling, and other natural interactions, are ok (20% reduction), while doing gestures (e.g.  \o/ or  Lo⅃ ) might incur in 40% reduction. Taking robot's hand and slightly pull wasn't allowed before, but now it is. This is mechanically guiding the robot, so I would apply at least 80% reduction if not 100% reduction.

As I said above, it would come handy to have the reductions in the scoresheet.

### Example
**Main goal:** Take the bag to the car (300pts)
**Bonus goal:** Come back home (200pts)
**Deus Ex Machina:** (These might be only in the scoresheet)
- Losing track of the operator during the *guiding phase* reduces the score in 50pts
- Non-natural interaction during the *guiding phase* reduces the score in 100pts
- Touching the robot during a *guiding phase* neglects the score
- Requesting indications during the *Come back home* reduces the bonus' score in 50pts

tasks/HelpMeCarry.tex
kyordhel: I don't think so. To keep consistency with previous rulebooks you don't state that here. Instead, we normally add an additional rule about instructing the robot (you can find it in any Follow-me test, Navigation test, or Help-me-carry). Important is that during instruction:
- Max 1 minute
- No marks allowed
- No touching the robot
- **TEAM DO NOT INTERACT WITH THE OPERATOR** (I can never stress this enough)

syedaraza: In the last year's HelpMeCarry description it says,

Following the operator: When the robot signals that it is ready to start, the operator starts walking –in a natural way– towards the car. Upon arrival, the operator will indicate the robot when they have reached their destination as instructed by the robot (e.g. by saying “here is the car” or “stop following me”).

So a similar rule could say,

Reaching the Car: The robot can reach the car's location only by following the operator. When the robot signals that it is ready to start, the operator starts walking –in a natural way– towards the car. Upon arrival, the operator will indicate the robot when they have reached their destination as instructed by the robot (e.g. by saying “here is the car” or “stop following me”).

tasks/HelpMeCarry.tex
syedaraza: > A partial bonus can be awarded if the robot returns home with assistance (e.g. lost operator who needed to stop and wave). That's preferable over not returning home but shall score less than returning without losing track of the operator.

Section 3.8.2 says,

Bonuses: No bonus points can be scored when the robot requests help to solve part of a task that normally would grant a bonus.

If we are going to keep this rule then any help in solving the bonus task should result in zero points for that bonus task.


> Agreed, how it is done is irrelevant IF THE OPERATOR DOES NOT TOUCH THE ROBOT. Waving, calling, and other natural interactions, are ok (20% reduction), while doing gestures (e.g. \o/ or Lo⅃ ) might incur in 40% reduction. Taking robot's hand and slightly pull wasn't allowed before, but now it is. This is mechanically guiding the robot, so I would apply at least 80% reduction if not 100% reduction.

> As I said above, it would come handy to have the reductions in the scoresheet.

Agree.

tasks/HelpMeCarry.tex
syedaraza: Retractable Posts randomly placed in robot's path [like this](https://www.retaildisplaydirect.com.au/black-retractable-bollards/?gclid=CjwKCAiAqaTjBRAdEiwAOdx9xqzXfB8QDr1BujtBaUY280sbHXF9c7zBeBubXn9yR-syBUDteAx4aRoCJXYQAvD_BwE).

Or a random area covered with Barrier [tape](https://www.bollardshop.com.au/products/crowd-control/yellow-black-barrier-tape-75mm-x-100m/?gclid=CjwKCAiAqaTjBRAdEiwAOdx9xstZkVEZU78XDjh_ZKId9Cyj640E84OHxgY5_TlXLM1ibeb6qlKyAhoCvq4QAvD_BwE) and [bollards](https://www.safetyxpress.com.au/bollards/t-top-bollards-6kg-base/?gclid=CjwKCAiAqaTjBRAdEiwAOdx9xpmrHxkUiFaXdlaZr2fmy9lLuJFDLaVaciXw1LFmW_ZQLCJ7cG36lxoCfkcQAvD_BwE).

Those will cost under $100.

tasks/HelpMeCarry.tex
syedaraza: I agree. Finding the bag by understanding pointing is not hard and should not grand bonus points.

tasks/HelpMeCarry.tex
syedaraza: > * Losing track of the operator during the _guiding phase_ reduces the score in 50pts

If the robot looses track and then finds the operator without any human help then we should not penalize it.

tasks/HelpMeCarry.tex
syedaraza: Or

Following the Operator: The robot should indicate the operator when it is ready to follow. The operator walks naturally towards the car. _After reaching the car, the operator takes the bag back from the robot and thanks the robot._

It will appear as a more natural way of stop following me. Do you think it will work?

tasks/HelpMeCarry.tex
kyordhel: None of those (yet). This is the nature of objects the robot had to avoid in the past.

image
image
image
image


We have had retractable posts but so far robots just undo their path (no optimization), so no robot has crashed yet with the ribbon, although I'm 100% sure no robot is able to see them.

### Constraints:
- A **Hard-to-see object** won't be detected by a Laser Rangefinder Scanner at a fixed height.
- A **Small object** is lying on the floor and tops 5cm tall. Normally humans don't see them, but robots must because of "reasons".
- An **interactable object** is either a person, rolling chair, door, or something the robot needs to interact with to get it/him/her out of its way.

tasks/HelpMeCarry.tex
kyordhel: > Section 3.8.2 says...

I wrote Section 3.8.2, but at the beginning of §3 it says:

> Every rule in this section can be considered to implicitly include the term *"unless stated other-
wise"*, meaning that additional or contrary rules in particular test specifications have a higher
priority than those mentioned herein in the general rules and regulations.

From my previous experience, if a bonus reward is too challenging teams won't attempt it unless there is partial score involved.

Being said that, it is your test, you decide whether to allow partial scoring on bonuses.

tasks/HelpMeCarry.tex
kyordhel: Sounds fine to me.

maximest-pierre:
I see a couple of typos but we can fix it later. I will get @natcon0368 to go over it after its merge.

syedaraza:
@kyordhel Thanks for your reviews!

PR #549, opened 2019-02-08
Proposal for the Onlooker test.

syedaraza:
This test is proposed for Stage I. It involves person detection, person tracking, human activity recognition, and motion planning for activity recognition.

It is inline with the proposed Stickler for Rules test in Stage II. Hence, it will help teams to focus on activity recognition in both the Stages.

A well defined list of activities is very crucial for this test. Help will be needed from those who have done R&D in human activity recognition.


kyordhel:
@syedaraza 
1. Please add your name to your GitHub account. We'd like to know who's contributing.
2. Please enhance the description (i.e. TL;DR version)
3. We have no tests, we have **tasks** now

scoresheets/Onlooker.tex
```diff
@@ -0,0 +1,18 @@
+
+The maximum time for this test is 5 minutes.
+
+\begin{scorelist}
+	\scoreheading{Main Goal}
+	\scoreitem[5]{500}{reports the activities performed by the main person}
```
kyordhel: Missing *Deus Ex Machina reductions*

scoresheets/Onlooker.tex
```diff
@@ -0,0 +1,18 @@
...
+
+	\scoreheading{Bonus rewards}
+	\scoreitem[5]{500}{reports the activities performed by the second person}
```
kyordhel: Missing _Deus Ex Machina reductions_

tasks/Onlooker.tex
```diff
@@ -0,0 +1,57 @@
+\section{Onlooker [Housekeeper]}
+Robot observes activities performed by an individual and reports them.
```
kyordhel: To who?

tasks/Onlooker.tex
```diff
@@ -0,0 +1,57 @@
+\section{Onlooker [Housekeeper]}
+Robot observes activities performed by an individual and reports them. 
+
+\subsection{Main Goal}
+Robot reports the sequence of activities performed by a person.
```
kyordhel: To who?

tasks/Onlooker.tex
```diff
@@ -0,0 +1,57 @@
...
+	\item \textbf{Start Location}: The robot starts at a predefined location.
+
+	\item \textbf{Persons}: There are two people. At the start, the main person to look for is standing, and the second person is sitting.
+\end{itemize}
+
+\subsection{Additional rules and remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Partial scoring:} The task allows partial scoring for each reported activity.
+	
+	\item \textbf{Activity List:} talk to someone, talk on a phone call, reading, writing, relaxing, walking, drinking, and using laptop.
```
kyordhel: - ✔️ Phone call 
- ✔️ Reading
- ⚠️ Writing: i.e. taking notes with pen and paper?
- ⚠️ Relaxing: Ambiguous, since reading a book can be relaxing, as well as listening to music, etc. I'd change it to watching TV
- ⚠️ Walking: Might seem forced without a purpose. Some people walk while on the phone.
- ⚠️ Drinking: As in actually drinking from a bottle or as in holding a glass with vodka?
- ✔️ Using laptop
- ❌ Talk to someone: Can't do if there are only two people in the room and both are in mutually exclusive activities

Some other activities to include:
- Serving drinks
- Sleeping

tasks/Onlooker.tex
syedaraza: to the operator. Thanks

tasks/Onlooker.tex
syedaraza: Yes, writing will mean taking notes with pen and paper. May be writing couple of sentences would be enough.

I agree, relaxing can be ambiguous. But as you suggested we can replace it with sleeping which would be a more clear form of relaxing. Also, it would help if we restrict sleeping to sleeping (lying down).

I agree walking may seem purposeless and can be removed from the list of activities. However, we should somehow force the people to move around. Otherwise, it would not be interesting if a person does five activities while sitting or standing on the same location. There can be a rule saying people will move around the house to do different activities. And the referee can distribute the activities around the living room. For example, the person will read and sleep on sofa, write on the table, use laptop at the desk, drink from a glass of water placed on the kitchen counter, etc.

Drinking would be more interesting if someone actually drink from a glass or bottle. Otherwise, it could be the case that the person is holding the drink for someone else.

The reason to include 'talk to someone' was to recognize a group activity. This can be the only mutual activity between the two people. And before and after 'talk to someone' each person can do individual activities. I think it will be challenging specially when two people are doing activities in the same room. 

Serving drinks would be a mutual activity. But it would not be the same for both people because one would be serving and one would be receiving. Some other examples of mutual (group) activities could be playing board game and moving table together.

scoresheets/Onlooker.tex
syedaraza: There is no much need of DeM in this task. The robot is not supposed to interrupt people in the house. Asking the person to confirm if he/she is doing ActivityX would be meaningless. Besides, asking to repeat an activity is rude and unnatural. So the robot has to observe the people in the house silently.

A DeM can be used only if the robot fails to back to the operator (the start position) to report and asks the operator to come to the robot or wave or call the robot.

scoresheets/Onlooker.tex
syedaraza: Please see the above comments.

tasks/Onlooker.tex
syedaraza: exercise (push ups or squats) can an interesting activity to report.

kyordhel: I mostly like some of your ideas, but I must admit that although they sound appealing for a lab setup of 2~3 runs, they are completely unfeasible in the competition.

I can ask  volunteer to sleep for the 3 hours a block last, but exercising, running from here to there, and other kind of physically demanding activities, I don't think so. Same for drinking. I can hold a glass of water for 3 hours, but I can't drink more than a couple of sips every 30 seconds for more than half an hour before needing urgently to make a technical step-over.

syedaraza:
@johaq I agree those are important concerns.

In the first three minutes, the audience will observe the two persons doing different activities because they would be curious to see if the robot can report those activities correctly. Besides, the robot would need to move head and change its position to have a clear observation of activities. I believe it wouldn't be possible to observe all the activities from a single angle and position. So it will be interesting for the audience to see if the robot is making the appropriate moves.

Regarding the waste of time concern, I think three minutes is a small duration. Furthermore, we will have only lightweight activities. We will not include any activity which will be physically demanding. Therefore, repeating the activities will not be a burden on the volunteers.

It is true that it will be hard for the referee to judge if something is wrong with the robot. However, if the robot is looking in the direction where the people are, then it deserves a chance. Otherwise, if the robot continuously look in the wrong direction (for example, looking at the wall or audience) then the referee can judge something is wrong.

Random guessing would be hard because there would be 6720 permutations possible if a person performs 5 activities selected from a pool of 8 activities.

tasks/Onlooker.tex
syedaraza: You are right those activities would be hard to repeat. Besides, recognizing that a person is holding a glass or bottle would be enough to ascertain that the person is drinking because there would not be any other activity which would involve holding a glass. Therefore, there is no need of actual drinking.

So the list of activities will be,

Phone call
Reading
Writing
Sleeping
Drinking
Using laptop

We should have at least two more activities (i.e., eight in total) so that random guessing of activities would be hard. However, it will be good to have more activities.

How about 'play with a ball' and 'clean a table'?

johaq:
Well the thing is, there are other leagues at robocup. If I go to a robotics competition and see only people doing smth in the arena for even just a minute I will move on to the next league's arena.

In the current form you get points per guess so guessing is 1 out of 8.

These are actually not my main concerns. Not everything has to super exciting for audiences imo. But right now this task also does not seem very interesting to implement as a robot behavior for the teams. Maybe an idea is to have the robot inquire about the activities. So if the robot sees someone reading, it drives to the person and asks what they are reading or if it is a good read etc. You'd have the robot more active and more HRI.

kyordhel:
@syedaraza Please take off your hat of scientist and wear your hat of average, not that smart person.

You, as a commoner, are in a venue full of robots. You're pushing your last selfies to your social networks, chasing for something interesting to make a pic or a video and brag with your friends. A robot cooking, storing groceries, or serving drinks will immediately catch your eye, whilst a robot *observing* people (whether it moves to another angle or not) is pretty much boring. Since you're a commoner, your attention span is no bigger than 15 seconds, and I'm being generous.

> if the robot is looking in the direction where the people are, then it deserves a chance

Ten years in this business have taught me teams will run a stage machine and bet for a guess. They might be able to detect only 2 gestures and will react to only those two (e.g. object recognition to check if I see a laptop, or check if the elbow-joint in a skeleton has 90°, you bet that human is drinking). Action recognition? Doubt so, but quite robust point farming.

As @johaq said, the number of permutations is meaningless to an experienced competitor that knows what to expect. And let's be honest. Sleeping, drinking and reading are the ones a referee will choose.

In this, I side with @johaq in that the test looks boring from all non-scientific perspectives. It can be fixed by adding a purpose. Back in 2014 we had a granny standing up who would:
- let her blanket fall
- let her cane fall
- fall herself

Robot had to either offer the cane, pick up the blanket or call 911. The action makes times easier scoring, specially when you consider a noisy environment and not-that-good speakers, so Referee can't understand the robot (trust me, we often can't).

Being that said, we can still keep the test, with some minor mods to make it harder to cheat.

syedaraza:
@johaq and @kyordhel your comments make sense and based on your experiences you can tell better what appeals the audience.

I will see how this test can be modified to be more interactive. Otherwise, the idea of inquisitive robot as @johaq told above sounds interesting too.

syedaraza:
How about the robot recognizes the activity of a person and delivers the requested item (an accessory)?

Every one is doing an activity and have requested an item. The robot takes those items from the operator and delivers to the right person (based on the object and the recognized activity).

An example of activity-accessory pairs can be,
Reading -> Glasses
Writing -> Pencil/Pen
Using Laptop -> Headphones
Sleeping -> Pillow
Drinking -> Straw
Playing with football -> A pair of socks

If say delivering first four (randomly selected) items is hard but doable then delivering the rest of two items would become the bonus tasks.

If possible, it can use both hands to save time.

Would it be an interesting task for stage I?

kyordhel:
@syedaraza I think most people won't like a robot disturbing during an activity (I would be one of them), but for sake of the test, we will state our robot has been trained to do so.

On your suggestion
- **Reading** -> *Glasses:* OK
- **Writing** -> *Pencil/Pen:* Eraser or also glasses. I can't write without pen or pencil
- **Using Laptop** -> *Headphones:* OK
- **Sleeping** -> *Pillow:* This is a no go, if I'm sleeping and the robot wakes me up to offer me a pillow, I kill it.
- **Drinking** -> *Straw:* And trigger environmental activists? No way. What about some ice?
- **Playing with football** -> *A pair of socks:* Doesn't really make sense. Besides we are not playing football in the arena.

My 2¢
- **Watching TV while sit:** A bowl with nachos or a drink (superbowl) 
- **Watching TV while lying:** A pillow
- **Looking for a something in the Bookshelf:** Remind her that she left the book in the night stand
- **Drinking:** Either a snack or ask if finished drinking want some more or robot disposes the can/glass.
- **Writing:** Envelope and stamp (i.e. writing a letter to be sent via postal service).

syedaraza:
Maybe the description was not clear enough. The main idea is that everyone has already started an activity and during the activity they feel need of an item and request the robot to bring it to them (lets assume by sending a text message to the robot). So they will not be disturbed. In fact, they will be pleased because they requested it.

The rationale behind the proposed activity-item list was,
Writing -> Pencil/Pen (A person is writing using a pen and have requested a different color/style pen). But, I agree an envelop or a stamp would sound more interesting.
Sleeping -> Pillow (A person just lay-down the bed and closed eyes but now feeling a need of another pillow)
Drinking -> Straw (I thought it might be interesting to see if robot confuses a straw and with a pencil). But a snack will also work.
Playing with football -> A pair of socks (a person is ready to go outside to play soccer and is having a football in hand and requests to bring socks). I agree this is sort of made up and not a very clear need.

I was not sure if there will be a TV in the arena. If there is going to be one, then those TV related activities would be interesting.

> * **Looking for a something in the Bookshelf:** Remind her that she left the book in the night stand

It is an intriguing idea but I doubt it will align with the spirit of the main task.

PR #551, opened 2019-02-09
Added the Is the Refrigerator Running? task.

justinhart:


tasks/IsTheRefrigeratorRunning.tex
```diff
@@ -0,0 +1,37 @@
+\section{Is the Refrigerator Running [Housekeeper]}
+The people who live in the arena are so sloppy! The robot finds minor faults in the state of the arena and rectifies them. This is an advanced version of Clean Up for Stage II.
+
+% \subsection{Focus}
+% This test focuses on object perception, manipulation, and planning.
+
+\subsection{Main Goal}
+The robot wanders around the apartment looking for things that are wrong (we will provide a list of potentially-wrong things), and then fixes them. Examples: The refrigerator door is open, the pillows from the bed are on the floor, the television is on with nobody watching.
```
moriarty: As is I don't see this as much different than the `Clean Up` task from Stage I.

The list of potentially-wrong things would need to be provided in this PR for it to be merge-able.

justinhart: This was way more exciting when we discussed it as a group. We sort of had a brainstorming session to get people pumped about 2019, and the tasks that we submitted were a result of that. I'll ping my team to ask for some suggestions here.

kyordhel:
@justinhart Please edit the description of the PR explaining what this task is about (i.e. TL,DR; version).

tasks/IsTheRefrigeratorRunning.tex
```diff
@@ -0,0 +1,37 @@
...
+
+\noindent\textbf{Reward:} 2500pts (100pts per \textit{wrong thing} identified, 300pts per \textit{wrong thing} fixed).
+
+\subsection{Bonus rewards}
+\begin{enumerate}[nosep]
+	\item Fix everything. Announce satisfaction at a job well done. (250pts)
```
kyordhel: Meaning?

tasks/IsTheRefrigeratorRunning.tex
```diff
@@ -0,0 +1,37 @@
...
+	\item Tell the operator. Find the operator and tell them how it was fixed. (50 per item)
+\end{enumerate}
+
+
+\subsection{Setup}
+\begin{itemize}[nosep]
+	\item \textbf{Trash my home!:} Make all of the things wrong.
```
kyordhel: What and how?

tasks/IsTheRefrigeratorRunning.tex
```diff
@@ -0,0 +1,37 @@
...
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per object as follows:
+	\begin{itemize}[nosep]
+		\item \textbf{Verbal request for the fix:} Asking an operator to fix the thing that is wrong is severely penalized, and causes a 90\% point reduction for the fix.	
+	\end{itemize}
+\end{enumerate}
+
+\subsection{Referee instructions}
+The referee needs to
+\begin{itemize}
+	\item Place the objects in the room.
```
kyordhel: Which?

tasks/IsTheRefrigeratorRunning.tex
```diff
@@ -0,0 +1,37 @@
...
+	\item Mess up the room.
```
kyordhel: How?

tasks/IsTheRefrigeratorRunning.tex
kyordhel: We must provide a comprehensive list of stuff that might be wrong, and then clamp the e.g. 5 things the robot must look for (we mess up 7). As is, the main goal is so fuzzy I wouldn't know how to code it.

tasks/IsTheRefrigeratorRunning.tex
```diff
@@ -0,0 +1,37 @@
+\section{Is the Refrigerator Running [Housekeeper]}
+The people who live in the arena are so sloppy! The robot finds minor faults in the state of the arena and rectifies them. This is an advanced version of Clean Up for Stage II.
```
kyordhel: From here I thought the fridge was broken and I got a robot-technician to fix it.

# Task Title: Tidy up the apartment
# Description:
The robot patrols the apartment and fixes up the mess.

tasks/IsTheRefrigeratorRunning.tex
HareshMiriyala: Cleanup is about identifying objects in a room and classifying them as not-trash vs trash. After detection, the action to be carried out is - if its trash, throw them in the dustbin, else place them where they belong by navigating to the room and place it there. whereas, in this particular task, we propose a different action category, in which the robot interacts with the object - for example, the robot detects the fridge door is open, interacts with the door to close it, detects the television is running, but there's no human in the room to watch/hear the tv, looks for the remote, and switches off the television, sees that the fan/ lights are on but there's nobody in the room so it looks for the switch and turns it off , detects the electric stove is on as identified by the red coil color and theres no vessel on top of it - so it looks for the knob and turns it off. There can be several such examples where the robot needs to interact with the environment to fix it, rather than pick and place the object.

kyordhel: I agree with @moriarty in that the setup is similar to `Clean up`, in contrast to `Dishwasher`, `Help me carry`, etc.

On your suggestions:
1. OK
2. Hard, almost impossible in the Arena. I'm not sure we would be able to use the TV there
3. Fan is hard but might be okay, but lights is almost impossible in the arena
4. Don't even think about this
5. ...
    - Main door: doable
    - Window: impossible
    - Cupboard: doable
6. If my robot dares to do that, I dismantle it in the most painful way.

justinhart:
It may be difficult to make this precise enough in time for the rulebook. The core idea here is that we would use vision algorithms that go a bit beyond object recognition to identify the various states of the appliances and items in the arena. We may need to come up with a big list of these things and see if we can make this worth doing. I agree that, as stated, it isn't quite complete.

moriarty:
Shall this be closed?

We won’t be merging it for 2019

PR #554, opened 2019-02-09
Add the Smoothie Chef task

justinhart:


tasks/Bartender.tex
```diff
@@ -0,0 +1,29 @@
+\section{Bartender [Party Host]}
+The home owner will teach the robot to make a drink which guests at the party really like!
+
+\subsection{Focus}
+Learning from demonstration.
```
moriarty: I like this.

tasks/Bartender.tex
```diff
@@ -0,0 +1,29 @@
...
+
+\subsection{Main Goal}
+The home owner will teach the robot using demonstrations how to make a drink. The robot is then to repeat the task. A designated team member will show the home owner how to perform the demonstration. These drinks need to be \textit{good}, meaning that the robot must pour each mixer in for the right amount of time to get the right amount of mixer in.
```
moriarty: Getting a working blender in the arena and liquids might be problematic. Also the whole mixed drinks things in case RoboCup ever goes to a location which is dry... but the concept of the task, learning and programming by demonstration is very nice.

justinhart: Yeah. We were thinking that the robot would pick up bottles and pretend to mix drinks, rather than actually pour liquids, which is what we do in similar research activities at UT. Well, we don't use alcohol in those activities, but the idea holds.

kyordhel:
@justinhart Please edit the content of the PR explaining what this task is about (i.e. TL,DR; version).

tasks/Bartender.tex
```diff
@@ -0,0 +1,29 @@
+\section{Bartender [Party Host]}
+The home owner will teach the robot to make a drink which guests at the party really like!
```
kyordhel: How  about:

> *The home owner teaches the robot how to prepare drinks*

tasks/Bartender.tex
kyordhel: I find this overly complex, not to say impossible, for the following reasons:
1. We haven't addressed blenders or any other type of domestic appliance before.
2. We haven't handled liquids as part of regular tests before, let alone pouring them.
3. We haven't addressed properly pouring solids.
4. Measuring amounts of is still out of the chart.
5. Having an HSR handling liquids voids the warranty.

Further, the main goal is too complex. Short version required. Suggested:

> The ~home owner~ operator teaches the robot how to make a drink by demonstration. Once the operator has finished the demonstration, the robot has to repeat the same procedure it observed.

The team has no saying in this. The robot must clearly state how the operator must trigger the learning behavior. It must be natural.

Finally, I lack dialogue components people often used when instructed. After learning how to make a Mojito I had several questions so I asked.

tasks/Bartender.tex
```diff
@@ -0,0 +1,29 @@
...
+Learning from demonstration.
+
+\subsection{Main Goal}
+The home owner will teach the robot using demonstrations how to make a drink. The robot is then to repeat the task. A designated team member will show the home owner how to perform the demonstration. These drinks need to be \textit{good}, meaning that the robot must pour each mixer in for the right amount of time to get the right amount of mixer in.
+
+\noindent\textbf{Reward:} 2500pts (400 points per ingredient).
+
+\subsection{Bonus rewards}
+\begin{enumerate}[nosep]
+	\item Screwing a lid onto a screw-top blender (500 points)
```
kyordhel: Blenders are expensive, we are not sure teams will try this at all, and you need to add the blender to the objects section under General rules.

tasks/Bartender.tex
```diff
@@ -0,0 +1,29 @@
...
+\end{enumerate}
+
+
+\subsection{Setup}
+\begin{itemize}[nosep]
+	\item \textbf{Bar:} Place mixers onto the counter in the kitchen.
+	\item \textbf{Blender:} Place blender into the kitchen.
```
kyordhel: Unnecessary, the blender lives there

justinhart:
Now it's a smoothie chef placing fruits into a bowl, so there are no liquids, no alcohol (there was concern at UT), and no blender. I think it retains the good stuff about getting some rudimentary demonstrations into the system, and maybe in future years we can make it a bit tougher.

tasks/Bartender.tex
justinhart: No liquids because it's smoothies now, and picking up fruit? Maybe this is still good?

scoresheets/SmoothieChef.tex
```diff
@@ -0,0 +1,14 @@
+The maximum time for this test is 15 minutes.
```
johaq: Has to be 10 minutes so it fits in a slot with other stage 2 tests.

scoresheets/SmoothieChef.tex
```diff
@@ -0,0 +1,14 @@
...
+
+\begin{scorelist}
+	\scoreheading{Main Goal}
+	\scoreitem[5]{500}{Place fruit items into the bowl}
```
johaq: This seems like a LOT. Object rec on the bowl after demonstration then one pick and place operation in a fairly controlled environment (compared to let's say storing groceries) gives as much points as a full Stage 1 tasks.

tasks/SmoothieChef.tex
```diff
@@ -0,0 +1,29 @@
+\section{Smoothie Chef [Party Host]}
+The home owner will teach the robot to make a smoothie which guests at the party really like!
+
+\subsection{Focus}
+Learning from demonstration.
+
+\subsection{Main Goal}
+The home owner will teach the robot using demonstrations how to make a smoothie. The robot is then to repeat the task. A designated team member will show the home owner how to perform the demonstration. To avoid liquids and blenders, we will have the robot place the fruit corresponding to the desired smoothie into a bowl.
```
johaq: Main goal should be short. Maybe smth like: "The robot selects ingredients for a smoothie for which it learned the recipe."

tasks/SmoothieChef.tex
```diff
@@ -0,0 +1,29 @@
...
+\subsection{Setup}
+\begin{itemize}[nosep]
+	\item \textbf{Kitchen Counter:} Place plastic fruit onto the counter in the kitchen.
+	\item \textbf{Ingredient Bowl:} Place bowl into the kitchen.
+\end{itemize}
+
+\subsection{Referee instructions}
+The referee needs to
+\begin{itemize}
+	\item Witness the team member demonstration.
```
johaq: I'm not quite sure about this. So the team member demonstrates to the operater how to do it and then the operator to the robot? Is the first part included in the test's time?
One of the general approaches in @home is that only the robot is allowed to instruct the operator. I do not see why the robot could not explain how it needs things to be demonstrated. 
Otherwise we also had teams give like a handout for certain tests which contained for example how to address the robot in the restaurant test.

kyordhel:
@johaq I'm approving this PR despite your comments with which I agree. For time constraints I'm taking over and giving a final buff directly on 2019/Restructure

PR #558, opened 2019-02-19
Modified GPSR to fit Stage 1

johaq:
1. Only 3 commands
2. Distinguish between task and commands
3. Only one operator to keep already complicated setup manageable
4. One location for the robot to receive commands


kyordhel:
@johaq  please `git pull --rebase` before pushing anything else

johaq:
Yeah sorry. Had this stored a while ago and did not update. 
This is meant to move gpsr from stage 2 to stage 1. So i removed gpsr from stage 2 now.

kyordhel:
Can we keep both? GPSR in Stage I and EGPSR in Stage II?

kyordhel:
@johaq I'm stealing your mods and incorporating both tests.

Will push them directly.

PR #559, opened 2019-02-20
Fix Makefile for rule book compilation.

balkce:
Fixed an issue with Makefile that did not let met compile the rule book. This should work for everybody.


kyordhel:
Thanks!

Issue #578, opened 2019-03-07
Q: Start procedure for tests that do not begin with a door opening


RemiFabre:
The "3.7.8 Start signal" section doesn't cover the default start procedure when the test doesn't start by a door opening.

For example, in the test _Receptionist_ the robot should start by going towards the door and try opening it.

A simple solution would be that the team is responsible for providing a one-button start procedure so that the referee can start the test when everyone is ready. In most cases this would be pressing enter on the keyboard. Is this ok?

moriarty:
Yes, but without Keyboard.

The robot should communicate the start procedure. When the robot is ready to start (if using ROS- all nodes are up and ready) then it should indicate it is ready, and give some simple “one touch” instruction to the referee. The referee will do as they were told, or at least try to do what they understood.

Most teams using the Toyota HSR in DSPL, use a press of the wrist joint.

Other teams, use some speech command... but be careful because it’s noisy and some keywords are spoken frequently just before the test starts.

We always recommend a backup, such as an aruco marker.

The problem with Keyboards are that it’s too hard to control and we want to ensure that the robots are autonomous, if we see teams touching a Keyboard at the Externally Compute Resource Area (ECRA)- we will yell at them and potentially end their run.

moriarty:
Teleoping the robot, using a keyboard to press enter, and then pulling out the USB cable, should be the the “plan z”... or not allowed at all...
Although it has been allowed in the past... I think we should strongly discourage it.

I think we’d accept a PR which clarifies this rule.

kyordhel:
# TL;DR: Pressing Enter is [borderline] OK.
1. **Touching the robot afterwards causes immediate disqualification**.
    - This includes pulling/disconnecting a keyboard
2. Moving mouse and click are two actions, therefore not allowed.
3. The use of **wireless keyboards**/mouses counts as remote controlling, hence **disqualification**.
    - Remote computers can't be touched. Consequently, starting the robot from ECRA is remote controlling, i.e. forbidden.
4. The use of alternate HRI methods is highly encouraged.

## Explanation
I must say the rules seem quite clear to me
![image](https://user-images.githubusercontent.com/8583949/53954816-93016800-40d7-11e9-860d-79a92601ae15.png)

[...]

PR #580, opened 2019-03-07
Modify startup procedure in accordance with #578 as advised by @moriarty

kyordhel:
### Changelog
- Start button must be physical and on-the robot
- Alternative start signals must be allowed by TC
  - QR Codes allowed by default
  - Voice commands allowed by default
  - Software buttons and apps require revision
- It is forbidden to start with a keyboard/mouse or from ECRA


kyordhel:
@johaq many teams will fail to start by detectong the door, so I won’t enfoce it. 

Instead, there will be a penalty of 10% (100pts stage 1, 250 stage 2) for using start button). 

I’ll add it per test later on.

johaq:
Ok, sounds good to me.

PR #582, opened 2019-03-11
Add German Open note beside On-the-fly compile

moriarty:


moriarty:
This is to prevent too many emails going to @swachsmu asking where to find the RuleBook for German Open

MatthijsBurgh:
Why not create a German Open Release?

kyordhel:
Apologies in advance if I sound like a D.

Well, I think GO organizers could take the version (which hasn't really been modified since the inclusion of restaurant) and hang it in the GO website as official draft for the competition.

I think, that would be the best way to proceed.

PR #584, opened 2019-03-11
Fixes table of contents to include restaurant

kyordhel:
3 files changed:
- Rulebook.tex to include Restaurant in the chapter TOC
- Restaurant.tex to include the linking label
- SmoothieChef.tex to exclude an entry from the toc

To be merged after German Open


MatthijsBurgh:
The merge hold isn't needed anymore right? Because we have the release.

So the restaurant is in the GO rulebook, just not in the ToC?

Issue #587, opened 2019-03-12
Balance testing slots (fair competition)


kyordhel:
It seems the rulebook is not crystal clear in this aspect, so a clarification is needed.

In principle, all teams get the same number amount of testing slots, with a minimum of 2 per block. If a DSPL|OPL|SSPL block ends early and the referee has 30 minutes to spare, she can open an extra testing slot only if DSPL|OPL|SSPL has 6 teams or less. If DSPL|OPL|SSPL has 7+, open an extra slot is not possible because one team would be left out.

It is important to make a hard rule that all teams in a league receive the same number of testing slots per block. Yet is up to teams to use them or not (e.g. broken robot).

balkce:
Agreed. This clarification could go in Section 3.6.2 "Schedule". PR?

LoyVanBeek:
XPL? What's that? I assume x = {O, SS, DS}, right?

balkce:
> XPL? What's that? I assume x = {O, SS, DS}, right?

Yes. @kyordhel has a weird sense of writing style.

But you get used to it =P

kyordhel:
Sorry, I'm editing that to make it BNF compliant.

PR #589, opened 2019-03-13
Addresses #587 (balance of testing slots)

kyordhel:
- Updated time diagram for 3-days competition
- Explicitly stated all teams receive same number of testing slots.

### DIFF
--- a/general_rules/Organization.tex
+++ b/general_rules/Organization.tex
@@ -6,13 +6,10 @@
 
 The competition features a \iterm{stage system}. It is organized in two stages each consisting of a number of specific tasks. It ends with the \iterm{Finals}.
 
-Each \iaterm{stage} comprehends a set of tasks grouped in two thematic scenarios.
-% \iaterm{House Cleaner} and \iaterm{Party Host}.
-The \iaterm{Housekeeper} scenario features tasks related to cleaning, organizing, and giving maintenance; while the \iaterm{Party Host} scenario focuses in attending guests needs and providing general assistance during a party.
 
 \begin{enumerate}
 	\item \textbf{Robot Inspection:} For security, robots are inspected during setup days.
-  A robot must pass \iterm{Robot Inspection} test (see~\refsec{sec:robot_inspection}) in order to compete.
+	A robot must pass \iterm{Robot Inspection} test (see~\refsec{sec:robot_inspection}) in order to compete.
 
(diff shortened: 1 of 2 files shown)


benatuts:
I'm unclear how to interpret point 1 ("Thematic Scenario Blocks") in 3.6.2 with respect to the Morning of Day 2.

On Day 2 morning, are teams free to choose from either thematic scenario, or is it one slot of  Housekeeper followed by one slot of Party Host? If it is the latter, then which scenario is used if there is time remaining in the block to have an additional testing slot?

Thanks!

johaq:
Yes teams can choose from either scenario

general_rules/Organization.tex
```diff
@@ -29,26 +26,26 @@ \subsection{Stage system}\label{rule:stages}
...
-	Testing slots are randomly assigned to teams in each block.
+	\item \textbf{Slots:} In principle, all teams get the same amount of \iterm{testing slots} with a minimum of 2 per block.
+	If there is sufficient unoccupied remaining time for all the teams in a league, a referee can open an extra testing slot, allowing an extra run for all.
 
-	\item \textbf{Tests:} Teams must inform the OC in advance which task(s) will try to solve.
+	\item \textbf{Tests:} Teams must inform the OC in advance which tasks will be tested in each block.
 	Only one task can be attempted per test slot.
 
-	\item \textbf{Participation is default:} Teams have to indicate to the \iaterm{Organizing Committee}{OC} when they are \emph{skipping} a test slot. Without such indication, they may receive a penalty when not attending (see~\refsec{rule:not_attending}).
+	\item \textbf{Participation is default:} Teams have to indicate to the \iaterm{Organizing Committee}{OC} when they are \emph{skipping} a testing slot. Without such indication, they may receive a penalty when not attending (see~\refsec{rule:not_attending}).
```
MatthijsBurgh: Maybe a bit out of scope. But why 'may receive a penalty'? Just be strict.

kyordhel: Traditionally we apply a penalty for not attending. It is unacceptable not be there. You have to notify OC one hour in advance.

kyordhel:
@benatuts What you see in the rulebook is an example. The OC can modify those slots and the "freedom" to pick up an scenario will depend on logistic constraints as well as how teams are shuffled. It may be the case that the hybrid blocks takes place on the first day starting with Party Host and then Housekeeper.

PR #590, opened 2019-03-13
Task Selection Announcement

kyordhel:
Adds Task Selection Announcement as part of the procedure for team registration and qualification

### DIFF
--- a/general_rules/TeamRegistration.tex
+++ b/general_rules/TeamRegistration.tex
@@ -108,10 +108,22 @@ During the \iterm{qualification process} a selection will be made by the \iaterm
 	\item the contributions to the RoboCup@Home league.
 \end{itemize}
 
-\paragraph{Important note to Standard Platform Leagues:} Only unmodified robots may compete in Standard Platform Leagues. Any \textit{slight} modification made to the robot found in the Qualification Material will automatically disqualify the team, for which registration to the international competition will not be possible  (See~\refsec{rule:spl-mods}).
-
 % For getting considered in the evaluation, be sure to insert your team's name when adding information to the \iterm{RoboCup\char64Home Wiki}.
 
+\paragraph{Important note to Standard Platform Leagues:} Only unmodified robots may compete in Standard Platform Leagues. Any \textit{slight} modification made to the robot found in the Qualification Material will automatically disqualify the team, for which registration to the international competition will not be possible  (See~\refsec{rule:spl-mods}).
+
+\subsection{Participation confirmation}
+\label{rule:participation-confirmation}
+In order to max out the number of participating teams, qualified teams \emph{must} contact the \iaterm{Organizing Committee}{OC} to confirm (or cancel) participation.
+
+Confirming attendance implies that the team has sufficient resources to register, and commits to attend to the event.
+Qualified teams missing the \iterm{Participation Confirmation} may be disqualified.
+
+\subsection{Task Selection Announcement}
+\label{rule:task-selection-announcement}
+Aiming at ease test slot scheduling during the competitions, teams are required to notify the \iaterm{Organizing Committee}{OC} which tasks will be tested in each scenario/stage, at least one month before the competition.
+
+Task selection announcement is a non-binding decision, meaning that teams can opt-out from some tasks. However, it is not allowed to opt-in later on, as it is forbidden to do it for all tasks, since there are more tasks than testing slots.
 
 % Local Variables:
 % TeX-master: "../Rulebook"


MatthijsBurgh:
The maximum number of tests to announce should be limit to the number of timeslots. Otherwise teams can announce to do all tests.

MatthijsBurgh:
The teams’ deadline is one month. What is the deadline for (L)OC to announce the number (and type) of slots?

justinhart:
I agree with @MatthijsBurgh on this. The teams will need some kind of heads-up to know how many tasks to prepare.

kyordhel:
Let’s fix it to 3/3 in Stage 1 and 2/2 in Stage 2, excluding restaurant.
Picking less is allowed.

benatuts:
A month seems excessive and far more time than required to "facilitate scheduling of test slots during the competition".

kyordhel:
I am considering that most robots must be shipped 3~5 weeks before the competition to pass security inspection and customs. This means that one month in advance teams must have a very robust idea on which tasks they can solve with ease in lab conditions.

This month not only gives sufficient time to OC members (they are busy volunteers) to arrange an schedule, while also preventing teams from improvising tasks on the fly.

kyordhel:
@MatthijsBurgh 
>The teams’ deadline is one month. What is the deadline for (L)OC to announce the number (and type) of slots?

I'd say, ASAP, but I don't want to hardcode it. As for the worldcup, I'd say: _in this very PR_. My suggestion of (3/3, 2/2) came from the example schedule in §3.6.3 (pp 29-30). You get 1.5 housekeeper block and 1.5 Party Host one. With quick maths, if you have 2 testing slots per block 2*1.5 = 3. Stage II is straightforward due to Restaurant. Any additional testing slots can be used for repeating your worse execution.

@justinhart 
> I agree with @MatthijsBurgh on this. The teams will need some kind of heads-up to know how many tasks to prepare.

Understood, and agreed. As TC member you have the responsibility of helping to propose a solution to the problem, so all your suggestions are more than welcome. Above you can read my mindset, but I'd love to know what you think.

MatthijsBurgh:
> @MatthijsBurgh
> 
> > The teams’ deadline is one month. What is the deadline for (L)OC to announce the number (and type) of slots?
> 
> I'd say, ASAP, but I don't want to hardcode it. As for the worldcup, I'd say: _in this very PR_. My suggestion of (3/3, 2/2) came from the example schedule in §3.6.3 (pp 29-30). You get 1.5 housekeeper block and 1.5 Party Host one. With quick maths, if you have 2 testing slots per block 2*1.5 = 3. Stage II is straightforward due to Restaurant. Any additional testing slots can be used for repeating your worse execution.
> 
This allows for the teams to have to decide in one day. I would suggest that we add a default schedule. In the case that the schedule needs to be changed, the (L)OC should announce this minimal 6 weeks before the tournament. In case the announcement is closer to the tournament, the deadline of the teams moves accordingly, so we keep the 2 weeks to decide.

kyordhel:
> This allows for the teams to have to decide in one day. I would suggest that we add a default schedule. In the case that the schedule needs to be changed, the (L)OC should announce this minimal 6 weeks before the tournament. In case the announcement is closer to the tournament, the deadline of the teams moves accordingly, so we keep the 2 weeks to decide.

I'm not following you about why teams have to decide anything in one day. In any case, such decision sounds too rigid to me to be feasible, so let me clarify it.

1. One month in advance, teams announce which tasks they might test.
  - Only which, not when.
  - Choosing a task doesn't mean you will test it.
  - Not choosing a task means you will NOT test it.
2. During setup days, the OC releases the schedule with the blocks and the scenario of each block.
3. During Team Leader meeting, teams announce which tasks they are (almost) sure they will test.
4. Before the block (e.g. 2 hours) the OC adds the Task schedule and random team order
    - If no team will attempt a task, the task is not scheduled.
    - Teams can still change their mind
5. During the block, teams line up based on
    1. The task to be tested in that time slice
    2. The team order

All this is because some tasks require special objects (no point in spending funds on stuff that won't be used), and referees need to re-check the rules prior to the test. A complete random order is unfeasible for the referees who can't have the rules of 10 tests fresh on mind, so we schedule a task order. This is only possible if teams announce in advance what will they attempt.

So first you tell us: we prepared GPSR, Storing Groceries, and Take out the Garbage. During team leader meeting you say: our GPSR is not ready for tomorrow, so we assign you time in both Storing Groceries and Take out the Garbage. Then the test comes and everything goes wrong with Groceries so you want to repeat it ASAP. No problem, you get rescheduled. Or not, and you decide to drop GPSR and repeat Groceries the next day, which is also OK.

MatthijsBurgh:
About the teams to decide in one day. It wasn't clear enough.

If the (L)OC announces the number of tasks one day before the the deadline of the teams, Then the teams need to decide in one day. Therefore I suggested to set a deadline of 6 weeks, so teams allways have 2 weeks to decide.

kyordhel:
It is on the last commit for this PR.

image

So we are announcing it at the time of the merge. It becomes official and teams have a couple of months to decide.

general_rules/TeamRegistration.tex
```diff
@@ -108,10 +108,39 @@ \subsection{Qualification}
...
+The tasks a team can select should be according to the number and type of test slots of the competition (see~\Cref{rule:schedule,tbl:max-elegible-tests}).
+
+\begin{table}
+\centering
+\caption{Maximum number of elegible tests in Sydney 2019}
+\label{tbl:max-elegible-tests}
+\newcolumntype{C}[1]{>{\centering\arraybackslash}p{#1}}
+\begin{tabularx}{0.6\linewidth}{X C{0.2\linewidth} C{0.2\linewidth}}
+	\toprule
+	                  & Housekeeper & Party Host\\
```
MatthijsBurgh: As there are 6 slots in stage 1. Would you allow teams to 4 HK and 2 PH (or vice-versa)? Or would you want to limit it to equal numbers? (Stage 2 doesn't have 'free space'.)

kyordhel:
@balkce This requires merging based on what you said in #623

balkce:
I'm continuing here the conversation from #623, specifically @johaq input:

> I'm kind of against forcing teams to announce that early. Yes, as TC it is nice to know in advance but I know how much can still happen in the month prior to the robocup. I also think knowing so far in advance has actually very little organizational benefit. It is only good to know which objects need to be there (no need for yellow jacket if no teams want to do farewell) but due to gpsr and eegpsr we probably need every object anyway. In terms of setup knowing one day in advance was perfectly fine for setting up.

I think we are all in agreement that it would be good to know in advance which tests are going to actually be tested to reduce costs and logistical complexity. The problem here is **how** much in advance.

I think a compromise would be:

- Two weeks out, teams announce a "soft" list, from which they can opt out during the competition. No opt-ins are allowed.
- Confirmation of which tests they are going to attempt is made in the team leader a day prior.

This is more or less what is already in this PR (barring some date modifications).

If agreed by @johaq, @MatthijsBurgh (as participants of GO), @kyordhel (attending member of EC) and @justinhart and @awesomebytes (attending members of TC), I'll approve this request.

johaq:
That sounds good :+1:

kyordhel:
1. I think 1 month in advance is better, but I've no problem with 2 weeks.
2. I think the TL announcement is OK, but still teams should be allowed to opt-out after a run (e.g. failed at storing groceries so repeat instead of trying take out garbage).

balkce:
Have a look at the changes, as requested by @kyordhel . If nobody objects, this should be merged as soon as possible.

kyordhel:
I agree but as author I can't approve the PR

awesomebytes:
I was trying to approve and I got an error. It was just merging. I do think it all makes sense. 2 weeks is reasonable.

PR #629, opened 2019-05-13
Making Where is This a Stage II Task

balkce:
Upgraded the Where Is This task to be more adequate to Stage II, forcing the robot to provide directions AND guide the operator to the location referencing places along the way.

### DIFF
--- a/tasks/WhereIsThis.tex
+++ b/tasks/WhereIsThis.tex
@@ -1,19 +1,18 @@
 \section{Where is This? [Party Host]}
 \label{test:where-is-this}
-The robot has to explain to people where they can find places in the arena (e.g.~\emph{Where is the TV?}). The robot has either guide the operator there, or tell the operator how to get there (e.g.~\emph{It's at your right} or \emph{Turn around and go towards\dots}).
+The robot has to show people where they can find places in the arena (e.g.~\emph{Where is the TV?}). The robot has to tell the operator how to get there (in a summarized fashion) and then physically take the operator there in a type of tour guide.
 
 
 \subsection*{Main Goal}
-Give accurate directions or guide at least 3 people.
+Give accurate directions and guide at least 3 people.
 
 \noindent\textbf{Reward:} 300pts (100pts per direction/guidance).
 
 \subsection*{Bonus rewards}
 \begin{enumerate}[nosep]
-	\item Give accurate directions or guide 3 more people 600pts (200pts each).
+	\item Give accurate directions and guide 3 more people 600pts (200pts each).
 	\item Provide new instructions based on the previous ones 300pts (100pts each).
 	\item Provide directions to a naive operator 300pts (50pts each).
-	\item Provide audio recording and transcript of each interaction 600pts (100pts each).
 \end{enumerate}
 
 \subsection*{Setup}
@@ -32,28 +31,18 @@ Give accurate directions or guide at least 3 people.
 		\item \textbf{Follower's assistance:} Helping the robot while guiding a guest (e.g.~waving, shouting) causes a score reduction of 50pts.
 	\end{itemize}
 
-	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point}.
-	After guiding a person to its destination, it must return to the information point.
+	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point} and then guides them to the location.
+	After guiding a person, it must return to the information point.
 
 	\item \textbf{Naive Operators:} Optionally, questions can be asked by a \emph{Naive Operator}, i.e.~a person from the audience with no background on robotics. \emph{Naive Operators} are allowed to rephrase commands given by the referees.
 	\\\textbf{Remark:} Referees are not allowed to instruct naive operators on how to operate the robot. \textbf{Teams attempting to instruct or bias the operator will be disqualified}.
 
 	\item \textbf{Custom operator fallback:} If the robot consistently fails to understand the naive/professional operator (e.g.~3 times or more), teams can default to a custom operator.
 
-	\item \textbf{Data Recording:} Only when using Naive Operators, a team can get an additional scoring bonus by providing the recording and transcript of the issued commands.
+	\item \textbf{Guiding people:} The guiding phase of the test requires the robot to provide a type of tour guide of its explanation (pointing out reference locations at each step e.g.~\emph{The couch is at your right.}). This information can later be used as reference for later explanations. Blocked pathways may be encountered and should be used to make subsequent guides faster.
 
-	\item \textbf{Guiding people:} Instead of describing the location the robot may guide the person there but this will probably take more time.
-
-	\item \textbf{Returning guests:} All guests will come back later to ask for further instructions.
-	Example:
-	\begin{itemize}[nosep,leftmargin=1.2cm]
-		\item[Karen:] \emph{Where is the TV?}
-		\item[Robot:] \texttt{\small In the bedroom in front of the bed. Right door at the end of the corridor}
-		\item[Karen:] \emph{Where can I find the remote?}
-		\item[Robot:] \texttt{\small On the nightstand in the bedroom. Same room as the TV, Karen.}
-	\end{itemize}
-
-	If the robot can give new instructions based on the previous ones (i.e.~it remembers where the person was sent to), bonus points are awarded.
+	\item \textbf{Returning guests:} All guests will come back later to ask for other locations.
+	If the robot is able to guide an operator using information based on previous tours (i.e.~it remembers where the person was sent to or avoids blocked paths), bonus points are awarded.
 \end{enumerate}
 
 \subsection*{OC instructions}
--- a/tasks/WhereIsThis.tex
+++ b/tasks/WhereIsThis.tex
@@ -1,6 +1,6 @@
 \section{Where is This? [Party Host]}
 \label{test:where-is-this}
-The robot has to show people where they can find places in the arena (e.g.~\emph{Where is the TV?}). The robot has to tell the operator how to get there (in a summarized fashion) and then physically take the operator there in a type of tour guide.
+The robot has to explain and show people where they can find places in the arena (e.g.~\emph{Where is the TV?}). The robot has to tell the operator how to get there (in a summarized fashion) and then physically take the operator there in a type of tour guide.
 
 
 \subsection*{Main Goal}
--- a/tasks/WhereIsThis.tex
+++ b/tasks/WhereIsThis.tex
@@ -13,6 +13,7 @@ Give accurate directions and guide at least 3 people.
 	\item Give accurate directions and guide 3 more people 600pts (200pts each).
 	\item Provide new instructions based on the previous ones 300pts (100pts each).
 	\item Provide directions to a naive operator 300pts (50pts each).
+	\item Provide audio recording and transcript of each interaction 600pts (100pts each).
 \end{enumerate}
 
 \subsection*{Setup}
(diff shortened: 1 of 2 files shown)


tasks/WhereIsThis.tex
```diff
@@ -32,28 +31,18 @@ \subsection*{Additional rules and remarks}
 		\item \textbf{Follower's assistance:} Helping the robot while guiding a guest (e.g.~waving, shouting) causes a score reduction of 50pts.
 	\end{itemize}
 
-	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point}.
-	After guiding a person to its destination, it must return to the information point.
+	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point} and then guides them to the location.
+	After guiding a person, it must return to the information point.
```
johaq: We should add that the information point is not known and random for every team. Maybe:

The information point is not known beforehand. Right before a run at the entrance door, the team is told a position where they should start the robot.

balkce: Agreed. Added.

kyordhel:
The main idea of this test was to address high-level reasoning and language generation (previously unaddressed in @Home).

Based on feedback, it seems it was mostly implemented with state machines, which is not the goal. We can solve it with a generator but it would be too trivial. Here an example:

Locations: Stove, TV, 

image

Dialog:
- *Robot, where can I find the stove?*
- `Right behind you. Turn around and you will see it.`
- *And the TV?*
- `Which TV? The one of the living room or the one on the bedroom?`
- *The living-room one.*
- `Standing in front of the stove, turn left and walk until you reach the couch. You'll see the TV on your left.`
- *Awesome! What if I need to go to the toilet.*
- `Sitting on the TV couch you'll see two doors, the bathroom is behind the door on your left in a small cubicle.`

Notice that each direction is provided from the former one and the operator's perspective. This is important since it makes almost impossible to hardcode the test. I think guiding the operator is also fine and quite natural.

tasks/WhereIsThis.tex
```diff
@@ -32,28 +31,18 @@ \subsection*{Additional rules and remarks}
...
+	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point} and then guides them to the location. The information point is not known beforehand. Right before a run at the entrance door, the team is told a position where they should start the robot.
+	After guiding a person, it must return to the information point.
 
 	\item \textbf{Naive Operators:} Optionally, questions can be asked by a \emph{Naive Operator}, i.e.~a person from the audience with no background on robotics. \emph{Naive Operators} are allowed to rephrase commands given by the referees.
 	\\\textbf{Remark:} Referees are not allowed to instruct naive operators on how to operate the robot. \textbf{Teams attempting to instruct or bias the operator will be disqualified}.
 
 	\item \textbf{Custom operator fallback:} If the robot consistently fails to understand the naive/professional operator (e.g.~3 times or more), teams can default to a custom operator.
 
-	\item \textbf{Data Recording:} Only when using Naive Operators, a team can get an additional scoring bonus by providing the recording and transcript of the issued commands.
```
kyordhel: Why do we remove data recording? This must be kept as it is mentioned in the Scoresheet.

tasks/WhereIsThis.tex
```diff
@@ -32,28 +31,18 @@ \subsection*{Additional rules and remarks}
 		\item \textbf{Follower's assistance:} Helping the robot while guiding a guest (e.g.~waving, shouting) causes a score reduction of 50pts.
 	\end{itemize}
 
-	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point}.
-	After guiding a person to its destination, it must return to the information point.
+	\item \textbf{Information Point:} The robot meets guests and gives directions at the \emph{Information Point} and then guides them to the location. The information point is not known beforehand. Right before a run at the entrance door, the team is told a position where they should start the robot.
```
kyordhel: I think the info point must be announced 2 hours before the test so the robot enters through the door and reaches it (standard procedure). Two hours is insufficient to hardcode the test.

tasks/WhereIsThis.tex
```diff
@@ -1,19 +1,18 @@
 \section{Where is This? [Party Host]}
 \label{test:where-is-this}
-The robot has to explain to people where they can find places in the arena (e.g.~\emph{Where is the TV?}). The robot has either guide the operator there, or tell the operator how to get there (e.g.~\emph{It's at your right} or \emph{Turn around and go towards\dots}).
+The robot has to show people where they can find places in the arena (e.g.~\emph{Where is the TV?}). The robot has to tell the operator how to get there (in a summarized fashion) and then physically take the operator there in a type of tour guide.
```
kyordhel: **explain AND show**. Just guiding people is too trivial for a Stage II test.

tasks/WhereIsThis.tex
johaq: The point is that the start position is randomized for each team. I think it is not too much to ask to be flexible on the spot. Every team in Magdeburg manged that part,.

tasks/WhereIsThis.tex
balkce: @swachsmu brought up a good point in recording data of naive operators: we would need authorization from them (forms galore), and we would need to ensure that they would be able to request at any time for the removal of their recorded data. This also applies for all other tasks that involve naive operators (as mentioned in my notes).

To be consistent, I also removed it from the score sheet.

balkce: However, I'm open in writing a section of recording data of only participants (as they already filled out a form of the use of their voice and image as part of RoboCup activities).

tasks/WhereIsThis.tex
balkce: Agreed. Changed.

balkce:
I would like to have an agreement on this change soon, to leave teams enough time to implement.

kyordhel:
AFAIK you only need authorisation of a person if you’re going to publish:

1. Picture/voice for commercial purposes
2. Any biometrics linked to the person’a name or ID (e.g. Passport, social security, etc).
3. Picture (only face) if it will be world-wide publicly available along with other contextually sensitive information (e.g. news, google maps, etc). 

A person’s utterance of a command issued to a robot should not be a problem since the identity of the person is unknown and irrelevant. Besides, all people that attends to the event signs a form allowing robocup to record them because of the media.

MatthijsBurgh:
@kyordhel A naive operator could also be somebody from the public, I think they don't sign any agreement

kyordhel:
@MatthijsBurgh A Naive operator **must** be someone from the audience. I think visitors must sign something because they can be recorded by the media. I should confirm that.

swachsmu:
> AFAIK you only need authorisation of a person if you’re going to publish:
> 
> 1. Picture/voice for commercial purposes
> 2. Any biometrics linked to the person’a name or ID (e.g. Passport, social security, etc).
> 3. Picture (only face) if it will be world-wide publicly available along with other contextually sensitive information (e.g. news, google maps, etc).
> 
> A person’s utterance of a command issued to a robot should not be a problem since the identity of the person is unknown and irrelevant. Besides, all people that attends to the event signs a form allowing robocup to record them because of the media.

In Europe we have the General Data Protection Regulation GDPR about the protection of person-related data.  This is not only about publishing data (you need an agreement for that anyway), but also about processing and storing data where you can identify the person. With audio data of the voice you might be able to identify the person.  The purpose why you are storing the data must be transparent, a silent agreement of the recorded person is ok, but you are responsible to justify later that an agreement was given. Data recordings at public events where you are not focussing on a single person are ok, but if you record the audio data from a human-robot interaction, this is person related.

We need to ask us, what we want to do with the data recordings. If we want to build up data sets for later usage, we would need to publish the data (signed agreement would be necessary). If we do not want to keep it, why are we requesting it? What is the purpose?

Generally, I think selecting naive operators is really problematic. You just do not know, if you are getting people who are used to technology or not. This is ok as long as the robot is selecting the person (like in tour guide). If you select those persons as a referee, you biasing the test significantly.

kyordhel:
Two things:

1. If that’s the case, we should ban immediately all cloud-based ASR because google is definitively keeping the data. The same story goes for the rest: they keep the audio and use it for training.

2. That’s why the use of naive operators is optional, and a step that we sooner or later we need to take. We need the corpora. I know a referee might bias a test with a selection, but also I honestly trust our referees will play fair. I’d pick a random dude from the audience.

johaq:
> Two things:
> 
>     1. If that’s the case, we should ban immediately all cloud-based ASR because google is definitively keeping the data. The same story goes for the rest: they keep the audio and use it for training.

I actually think we should.

>     2. That’s why the use of naive operators is optional, and a step that we sooner or later we need to take. We need the corpora. I know a referee might bias a test with a selection, but also I honestly trust our referees will play fair. I’d pick a random dude from the audience.

This is super difficult. In Magdeburg I had the same volunteers from teams and some people I knew from the audience. The same group of people for every test. This is still not fair because the people practice and get used to the robots and the test so later starting teams have an advantage. 
Maybe this should be a more general discussion for a separate issue.

kyordhel:
@swachsmu @johaq Agreed. Let's discuss this in a separate issue.

swachsmu:
Regarding the cloud services. I do not think, we should ban those. I think there is difference between requesting data in the rulebook (responsibility of the league organization) and a team solution for a robot skill (responsibility of the team). Correct me, if I am wrong, but Google says, e.g., that they are only using your personal data for training, if you agree to subscribe to the Data Logging Program. Thus, the question which service fulfills some privacy protection is not simple at all. We cannot check, what teams are doing with their sensor data. Thus, we should leave it to the teams that they act according to scientific standards.

balkce:
I think that for the time being, I can re-instate the recording into this version of the test.

The "recording naive operators" issue topic affects other tests as well. If it is decided to remove it from tests with naive operators, we should do a PR for that specifically.

balkce:
I would like to have a decision on this so I can then modify it as is asked on #639 .

tasks/WhereIsThis.tex
moriarty: When we tell them the `Information Point` how do we communicate this to them?

1. Point and say "start the robot over there" ?
2. How do they move the robot to the robot their to start ? RViz navigation goal, robot teleop ?

tasks/WhereIsThis.tex
```diff
@@ -1,16 +1,16 @@
...
 
 \noindent\textbf{Reward:} 300pts (100pts per direction/guidance).
 
 \subsection*{Bonus rewards}
 \begin{enumerate}[nosep]
-	\item Give accurate directions or guide 3 more people 600pts (200pts each).
+	\item Give accurate directions and guide 3 more people 600pts (200pts each).
 	\item Provide new instructions based on the previous ones 300pts (100pts each).
 	\item Provide directions to a naive operator 300pts (50pts each).
 	\item Provide audio recording and transcript of each interaction 600pts (100pts each).
```
moriarty: This item is being discussed in the thread and not part of this PR diff. Did anyone do it in German Open?

I think the balance of points could lower the providing the transcript of each interaction to 50 pts each.

It seems like a lot to double the score of each direction/guidance when the effort and difficulty to record the data is relatively low, and it's a one time incurred effort. Whereas the behaviour of the recognition system is where we want to focus the points.

tasks/WhereIsThis.tex
kyordhel: I'd say 2 hours before the test. You can hard-code the Info Point position, but you can't really code all directions from there. Besides, is some sort of standard procedure. In addition, I think it's way more fair to have all robots getting questions from the same location.

tasks/WhereIsThis.tex
kyordhel: Provided transcripts are accepted only if the robot solved the task, so no point farming. Besides, that's the only way to build an @Home training corpora.

kyordhel:
@moriarty let's just add it under OC Instructions, 2 hours before test

balkce:
@kyordhel isn't it clear that the robot needs to point out each of the steps of its explanation during the guiding phase? The idea is that the robot is guiding AND explaining to the user the directions (ala "tour guide").

If it isn't, it may need to be more clear.

kyordhel:
Again, I don't think that guiding people around the arena should be considered at all as a Stage II test.

For instance the following may happen: 

- *Where is the bookcase?*
- ```Walk at the end of the corridor, turn left. I'll take you there.```

To me, this is ambiguous, and the guiding is trivial. I do not know whether the bookcase is in the corridor or in a room at my left hand. Ideal to me would be:

- *Where is the bookcase?*
- ```Walk at the end of the corridor, turn left. I'll take you there.```
- *Sorry, I don't get it. Can you take me there?*
- ```Sure!``` [Apply DeM score reduction]

The test only makes sense if the robot has to calculate and explain linked routes, as in:

- How do I find the stove?
- You'll see the stove on your left from the kitchen entrance
- And the food?
- The food is in the fridge which is in front of the stove. You'll see it at your right from the kitchen entrance.

For short, guiding would be ok to me if the robot should need to take the owner to wherever an object is (e.g. I can't find my glasses).

Issue #639, opened 2019-05-29
DeM reductions in scoresheets


kyordhel:
I recently found out that many tests have a potential score reduction (e.g. bypassing manipulation) not listed in the scoresheet. The Rulebook says that a 30% score reduction applies every time the robot requires human assistance, so that point is covered. However, it would come handy to have the most common reductions already written in the scoresheet and summing up nicely (e.g. 35 instead of 33.33333333... out of 100).

Here the list of test that require to be buffed

### Tests in Stage I
- [x] Carry My Luggage (Already correct in the RuleBook)
- [x] Clean Up (Already correct in the RuleBook)
- [x] Farewell (Already correct in the RuleBook)
- [x] Find My Mates
- [ ] General Purpose Service Robot
- [ ] Receptionist
- [x] Serving Drinks
- [x] Serve the Breakfast
- [x] Storing Groceries (Already correct in the RuleBook)
- [x] Take Out the Garbage (Already correct in the RuleBook)

### Tests in Stage II
- [x] Procter & Gamble: Clean the Table
- [ ] Enhanced General Purpose Service Robot
- [ ] Find My Disk
- [ ] Hand Me That
- [ ] Set the Table
- [ ] Restaurant
- [ ] Smoothie Chef
- [ ] Stickler for the Rules (aka RoboCop)
- [ ] Where is This?

MatthijsBurgh:
Receptionist is missing all DeM information

MatthijsBurgh:
Find My Disk doesn't have any penalty information

PR #646, opened 2019-06-04
Add DeM scores for cleanup in task description

MatthijsBurgh:
Addresses #639 cleanup


kyordhel:
Why did you closed it?

MatthijsBurgh:
It was already correct in the RuleBook. So only adding the points, while the percentage are already there is not needed IMO.

PR #649, opened 2019-06-04
HandMeThat penalty per question, max 3 questions

MatthijsBurgh:
Addresses #639 Hand Me That

### DIFF
--- a/tasks/HandMeThat.tex
+++ b/tasks/HandMeThat.tex
@@ -40,7 +40,7 @@ The robot identifies (touching or naming) each object at which the operator is p
 	Then
 	\begin{enumerate}[nosep]
 		\item The operator walks near to the object and points at it.
-		\item The robot can asks up to 3 questions as necessary.
+		\item The robot can asks as many questions as necessary.
 		\item The operator replies to each question (most likely with \emph{yes}, \emph{no}, \emph{I don't know}, etc).
 		\\\textbf{Remark:} The operator does not know the name of the object.
 	\end{enumerate}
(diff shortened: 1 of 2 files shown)


tasks/HandMeThat.tex
```diff
@@ -40,7 +40,7 @@ \subsection*{Procedure}
 	Then
 	\begin{enumerate}[nosep]
 		\item The operator walks near to the object and points at it.
-		\item The asks as many questions as necessary.
+		\item The robot can asks up to 3 questions as necessary.
```
kyordhel: Robot may ask more than 3 questions (waste of time). Question 4th and on won't be penalized anymore.

This makes sense since blind questioning is not allowed (fair play)

tasks/HandMeThat.tex
```diff
@@ -61,7 +61,7 @@ \subsection*{Additional rules and remarks}
 
 	\item \textbf{Incorrect guesses:} Incorrect guesses reduce the value of the correct guess by 200 points, each, the first two times. Guessing correctly on the third or fourth attempt is worth 100 points. After the fourth guess is worth no points.
 
-	\item\textbf{Colors and categories:} Asking for the color or category of a pointed object applies a penalty of 400 points for that particular object.
```
kyordhel: I would not remove this rule, otherwise the test becomes trivial

MatthijsBurgh: This is converted to the clarifying questions. So asking for color is also a reduction of 150.

MatthijsBurgh:
> Missing
> 
> `\item \textbf{Colors and categories:} Asking for the color or category of a pointed object applies a penalty of 400 points for that particular object.`

So asking for color/categories is a bigger penalty than another clarifying question?

kyordhel:
Yes, it is. It technically narrows the Q/A to 2 questions (e.g. only one object is red).

Issue #664, opened 2019-06-18
Bonus reward on main goal doubts


awesomebytes:
Hello,

I open this issue in a general fashion cause it applies to more than one test.

In Clean Up, there is a Bonus Reward for `Opening the entrance door autonomously`. The Main Goal is stated as: `Find all misplaced objects in a room and bring them to their predefined locations.`.

My assumption is that for getting this Bonus Reward a participant needs to find and bring ALL 5 misplaced objects to be able to have the Bonus Reward applied.

If that is so, I can only interpret that the next Bonus Reward `Moving a tiny object` or `Moving a heavy object` apply the same rule (instead of earning that bonus reward because of having moved that specific object).

In Receptionist the Main Goal is to `Introduce and allocate two newcomers in a party`. With a Bonus Reward of `Opening the entrance door to a guest (200pts each)`. By my previous interpretation if the door was opened once for the first guest it can only be scored as Bonus Reward when both guests are introduced and seated.

In Take Out the Garbage the same happens with the Main Goal being `The robot takes out the trash bags from the two bins in the apartment.` and the Bonus Reward of `Removing the lid of a bin` only will apply if taking out both garbage bag.

Also on Where is This? the Main Goal is `Give accurate directions and guide at least 3 people`. The Bonus Rewards state `Provide directions to a naive operator 300pts (50pts each).` and `Provide audio recording and transcript of each interaction 600pts (100pts each).`. By my previous interpretation, in the moment you get the 3rd person you unlock all those bonus points, otherwise, you don't get any.

Am I correct to interpret the rules in this strict way?

This may be happening in more tests, but these are the ones that this question came up.

Thank you very much.

johaq:
No, partial scoring enables scoring of bonus points. There is the partial scoring remark:
1. Partial scoring: The main task allows partial (per guest) scoring.
I think this was an oversight and should be added for these tasks. I'll get on that.

kyordhel:
@awesomebytes as a TC member you should be answering questions, not making them.

1. **Clean Up**. It's *five objects or nothing*, since you can handover the objects to the robot (reduction of 80%). Opening the door counts only if the robot cleans out these 5 objects. If you use DeM for the tiny/heavy, there is no bonus (i.e., you can't request the person to tell you where the object is).

2. Correct.

3. Correct.

4. Correct.

[...]

PR #665, opened 2019-06-18
Added partial scoring remark where appropriate

johaq:
@awesomebytes pointed out in #664 that some tasks did not allow partial scoring for the main task. I added the remark for all the task where I think the intent of the task is to be scored partially.

### DIFF
--- a/tasks/CleanUp.tex
+++ b/tasks/CleanUp.tex
@@ -27,6 +27,7 @@ Find all misplaced objects in a room and bring them to their predefined location
 
 \subsection*{Additional Rules and Remarks}
 \begin{enumerate}[nosep]
+	\item \textbf{Partial scoring:} The main task allows partial (per object) scoring.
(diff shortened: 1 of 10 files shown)


kyordhel:
# NO-GO

I simply can't approve these changes. Reasons as follow:

1. The rulebook is **FINAL**. Allowing partial scoring is a drastic change, way too big and should have been buffed between the Draft and FINAL versions.
2. Teams already made their task selection (perhaps some already shipped the robot). Such a big change would cause teams to reconsider their decisions.
3. Many tasks were designed to prevent point farming. The partial scoring rule was not added on purpose.
4. Altering several tests in one PR goes against our guidelines.

I think the TC had enough time after the German Open to review the rulebook based on @balkce's observations. This is simply not the time for such modifications.

Therefore, I'm closing this PR. It can be re-opened for the 2020 rulebook.

Issue #659, opened 2019-06-14
Request: Mint DOIs for finalized rulebooks


nickswalker:
I would like to reference the rulebook in a way that is both very easy for readers to follow while still durable (i.e. citation with a link that's guaranteed to not break). One way to do this would be to mint DOIs for finalized, official PDFs of the rulebook using something like [Zenodo](http://zenodo.org). There may be alternative archival repositories to consider as well, I am just familiar with Zenodo.

To check:
1. Are there DOIs that someone had issued set up already that just aren't visible? If so, I can put in a PR to list them in the readme
2. If not, would there be support for the idea of archiving rulebooks in this fashion? At the least, it provides a convenience. At best, it could make it easier to track back references to the rules.

balkce:
Just playing devil's advocate for a moment: If you're going for something that is "easy for readers to follow", I'm not sure a DOI link is the best to go: the provided index are bunch of numbers that require either some HTML (for a simple click) or to copy/paste. If you're looking for "a link that's guaranteed to not break", I'm not sure doi.org can guarantee that, or any of us for that matter (anybody remember the hi5 social network? or altavista?).

[taking off the devil's advocate hat] Ok. Seriously, though: would the official archive http://www.robocupathome.org/rules be close enough to what you're looking for? Or are you looking for a link for each one of the rule books?

At one point, I remember somebody made a link to one of the rule books via purl.org. Arxiv.org could be another possibility. Or you can go ultra-jank and use bitly or tiny.cc.

In any case, I agree that we should add a link to the past rule books to the Readme file of this GitHub, regardless of what linking method is used.

nickswalker:
I think the desirable properties for reference links are

* Will continue to work indefinitely
* Attached specifically to a single document
* The single document cannot be changed, or changes are officially noted

I think the league website works well for most purposes, but isn't appropriate for references. Because the actual rulebooks are hosted on Google Drive, there's no way to deep link directly to the official copy (URL is too long and gross), it's possible for them to be removed by error or intentionally, and it's possible for them to be changed in the future.

Also, since I first posted, I see that this note has appeared:
> Notice:
>The official RoboCup@Home website has been moved to http://athome.robocup.org.
> robocupathome.org is no longer maintained.

[...]

PR #678, opened 2019-07-18
Clarify licensing

nickswalker:
Archiving rulebooks (#659) involves declaring a license for the documents. This PR proposes applying a permissive [MIT license](https://choosealicense.com/licenses/mit/) to the source of the rulebook, and to apply [CC-BY](http://creativecommons.org/licenses/by/4.0/) to all official rulebook PDFs.


kyordhel:
I'm not sure whether the MIT license fits the needs for the RoboCup@Home Rulebooks. As a matter of fact, I think its way too permissive. Therefore, before approving this merge, I'd like to double check whether there is a more convenient license or not.

Below my concerns.

1. It is not software, but a lean set of rules and regulations
2. The community is allowed to fork, copy, modify, redistribute, etc; however
    2.1. They must cite the original work
    2.2. It must be free and remain open
3. The authors cannot be made liable for anything related to the use or misuse of the information contained in the rulebook and its offspring.
4. The creation and use of modified versions of this rulebook in local tournaments (@Home Opens) must be notified to the organization of the league.

We, as an open community, have placed our hearts, minds, and time in the creation of these rulebooks when aiming towards fostering research and development in robotics. Therefore, I think it's important that these efforts remain open and accessible to the world.

Perhaps a Creative Commons license would suit best?

nickswalker:
1. I can't find definitive guidance on this, but I think the contents of the repository would be described as "mixed": it's mostly the rules and regulations, but there's also Makefiles, TeX macros, automation and the such. From [what I read](https://choosealicense.com/non-software/):
  > If your project contains a mix of software and other material, you can include multiple licenses, as long as you are explicit about which license applies to what parts of the project

Here's an [example of a similar situation](https://github.com/github/choosealicense.com/blob/gh-pages/README.md#license)

2. 👍  Looking at the root Makefile, I see [it is already GPLv2](https://github.com/RoboCupAtHome/RuleBook/blob/6c6d58e854ff57a5d995f3fb65674bc0ec85201e/Makefile#L10) (derivative work must be open source, must adopt the same license). How about expanding that explicitly to the software parts of the rulebook repo, then adopting a "share alike" CC ([CC-BY-SA 4.0](https://choosealicense.com/licenses/cc-by-sa-4.0/))license for the content?
3. 👍 
4. Interesting. No existing license has this kind of stipulation (of course 😛) so we'd have to add it ourselves. One way this is already enforceable is that local events cannot use the protected RoboCup name and trademark without permission from the federation. We could add a sentence in the Readme emphasizing this.

I'll update this PR to reflect these comments.

README.md
```diff
@@ -58,6 +58,12 @@ The RoboCup @Home Organizing Committee is always looking for volunteers to assis
 
 Contributions and improvements to the rulebook are always welcome in the form of pull requests.
 
+### License
+
+The content of this project itself is licensed under the Creative Commons Attribution Share Alike 4.0 International license, and the underlying source code used to format and display that content is licensed under the GPL v2 license.
+
+Note that neither of these licenses grant rights to the RoboCup trademark. If you would like to host a sanctioned event, please contact the league organizers.
```
kyordhel: > Note that neither of these licenses grant rights to the RoboCup trademark.

... or to the use of the RoboCup logo.

> If you would like to host a sanctioned event, please contact the league organizers.

If you would like to host a sanctioned event including, but not limited to competitions, workshops, conferences, and educational projects, you must contact the RoboCup Federation.

I'm dubtious about this previous line. While RoboCup and the RoboCup logo are registered trademarks, I don't think the same is valid for RoboCup@Home and our logo. Furthermore, *the league organizers* refers to the @Home league organization (e.g. any OC member), while not to the president of the RoboCup Federation, so we need to be careful here.

Perhaps @iocchi can tell us better.

nickswalker:
1. That's right, I left it off because it is plain text
2. That is correct
3. 👍 
4. Ah, including that stipulation in the rulebook itself is a good idea. I see a couple of pages on the Federation site that explain processes: 
"Use of RoboCup Logo": https://www.robocup.org/logo_guidelines
"Proposals for RoboCup Related Events" section here: https://www.robocup.org/site_proposals
> RoboCup Related Events are those organized by the RoboCup Regional committees.
The committee must submit a proposal to Trustees to be approved following the instructions below. 

It seems the official process is: interested person gets in touch with Regional Committee. RC probably finds the nearest @Home expertise to help them coordinate a proposal to the Trustees. If Trustees accept, the event becomes official and can use the branding.

> IMHO, is unacceptable to acknowledge a local tournament as an official RoboCup@Home Open if only a portion of the rulebook is being used (remember that participation in open tournaments is factor for qualification).

> Therefore, iif you want to have a recognized event, you must issue an international CFP and use an TC approved version of the rulebook, implying you have to notify the @Home organization. Any other use of the rulebook, well, is creative commons so we don't care.

It seems like all that's required to accomplish this ^ is for the trustees to enforce these requirements as a part of the official proposal process.

For now, I will update the readme language with references to existing official processes, and wait on adding any additional expectations of official events until there is more input from the league trustees.

LICENSE
```diff
@@ -0,0 +1,72 @@
+GNU GENERAL PUBLIC LICENSE
```
moriarty: It sounds like we're going to do a dual license.

Please add a simple LICENSE and two more individual files LICENSE.GNU and LICENSE.CC-BY-SA-4.0

I agree with a dual license. I would have done even stricter CC-BY-NC-SA-4.0 but we should leave it CC-BY-SA-4.0 this is better for getting potential future sponsors. eg. Toyota or Softbank or PAL or Fetch want to sell robots, therefor their contributions could be seen as for a commercial purpose- Bigger companies like Google are very strict on letting their employees contribute when there is NC.

README.md
```diff
@@ -58,6 +58,12 @@ The RoboCup @Home Organizing Committee is always looking for volunteers to assis
 
 Contributions and improvements to the rulebook are always welcome in the form of pull requests.
 
+### License
+
+The content of this project itself is licensed under the Creative Commons Attribution Share Alike 4.0 International license, and the underlying source code used to format and display that content is licensed under the GPL v2 license.
```
moriarty: Add this information in the LICENSE file, and add two more files which the GLPv2 and CC-BY-SA-4.0

moriarty:
Thanks @nickswalker

nickswalker:
It would be nice to have buy-in from folks who were involved in earlier rulebooks as we should adopt these licenses to uniformly cover all versions. Not sure if anyone regularly talks to some of the folks already tagged, but to sum up:
* We want to apply these licenses to all @Home rulebooks
* We discussed additional limitations on the official use of the rules (allow international participation, use rules in whole) that we would like to be upheld when the Trustees consider approving any event related to @Home

moriarty:
Yeh- that’s why I didn’t directly merge I was waiting for someone else to approve.

kyordhel:
Those guys are Tijn van der Zant (@robotijn), Dirk Holz (@dirkholz), Caleb Rascon (@balkce), Loy van Beek (@LoyVanBeek), and I. There were others, but they were the top contributors (the code behind teh rulebook was mosltly crafted by Dirk, and he chose GPLv2).

You might want to ask them by email to ensure a response, but I'm 100% confident that they will agree with this licensing.

PR #681, opened 2019-07-25
Add helper script to compile in Docker

moriarty:
Usage:

    ./latexdockercmd.sh pdflatex Rulebook.tex

Requires Docker to be installed- but not LaTeX.

## Description

I'm lazy and don't like installing things.
This allows me to edit the RuleBook faster.
I assume other people are lazy too.

## Other comments
This is just a helper script, it makes it really easy for people to contribute to the RuleBook. I have only tested it on Linux but in theory anything with Docker installed will work, no need to install LaTeX. LaTeX is large, so the first run will download LaTeX in a docker and run everything cleanly inside of the container.

The script mounts the present working directory with the correct user which is nice becuase it doesn't mess with git or local editors etc.

### DIFF
--- a/latexdockercmd.sh
+++ b/latexdockercmd.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+IMAGE=blang/latex:ubuntu
+exec docker run --rm -i --user="$(id -u):$(id -g)" --net=none -v "$PWD":/data "$IMAGE" "$@"


kyordhel:
Marvelous! Thank you @moriarty 

Just out of curiosity, why just `pdflatex` once and not make?
Makefile requires run
```bash
pdflatex
makeindex
makeindex
pdflatex
pdflatex
```

So even though it will make the rulebook, indexes may not be generated

But you tested it so I’m approving the PR

moriarty:
I didn't take a really close look at the indexes but this works good enough for me to make and test quick edits to the rulebook and see the results.

I have had it in my workspace for a while and should have committed it sooner :laughing:

Issue #685, opened 2019-07-29
Reduce Number of Tasks


johaq:
Based on the competition in Sydney we want to reduce the number of tasks to:
2 tasks in Housekeeper and Party Host for Stage 1 (GPSR is mandatory)
1 task in Housekeeper and Party Host for Stage 2 (EEGPSR and Restaurant are mandatory)

Please write down which tasks you think should be kept. 

We will collect proposals until 05.08. Everyone is encouraged to participate, TC members are expected to. After that we will vote on the most popular choices.

johaq:
- Stage 1
   - Housekeeper
       - Take out the Garbage: high participation numbers this year and easy to understand
       - Storing Groceries: Imo the most useful task of the remaining ones and I think we really need manipulation tasks in this scenario
   - Party Host
       - Carry my Luggage: Person tracking and following are imo very important skills
       - Receptionist: Useful and high participation numbers this year

- Stage 2
   - Housekeeper
       - Set the Table: useful and easy to understand, good participation this year
   - Party Host
       - New: I am not really happy with any of the Stage 2 Party Host tasks and would prefer a new one.

semaph0r:
I agree on what @johaq have choosen for stage 1 and 2. 

AFAIK person tracking and following is not as stable as it should be, so emphasizing these may push teams in the right direction.

Regarding Stage 2 Party Host: What I would like to see is some sort of situation aware navigation in a highly dynamic environment, which party scenarios are. For example, find your way through a dynamic crowd and choose a strategy ad-hoc to decide whether to ask people to move or to find an alternative solution. For sure, a robot shall be as less interrupting as possible, but sometimes he needs to behave like a human waiter and say: Sorry, I have delicious food, may I pass by, please? Sure this may be hard to grade, but dealing with close proximities with robots is a thing.

Also dealing with good contact/forbiden contact during manipulation and the whole scenario might be an option.
Of course regarding this, safety has to be our top priority. But allowing the robot some sort of proximity and "contact interaction" with the environment (not people, this might be too dangerous as of now) would be a more natural apporach.

moriarty:
August 5th 2019 is the deadline. Tomorrow.

Which timezone?

peterpena:
- Stage 1

    - Housekeeper
        - Storing Groceries: It is a hard manipulation problem and has the potential for 
           human-robot interaction. The current 'Storing Groceries' needs to be expanded to contain 
[...]

PR #686, opened 2019-08-17
Make Take out the Garbage more challenging

nickswalker:
## Description

Makes _Take out the Garbage_ harder

- Require lid removal
- Add nearby floor garbage which must be placed in the bin

## Other comments
From discussion in #685, there is enthusiasm about both _Clean up_ and _Take out the Garbage_, but we will only keep one. There has also been some agreement that _Garbage_ will need to be harder. One path could be to add an element of _Clean up_ to _Garbage_. 

Not sure about how the score balancing would look, I just put down something roughly feasible. Bonuses would have to be reworked too.

### DIFF
--- a/scoresheets/TakeOutGarbage.tex
+++ b/scoresheets/TakeOutGarbage.tex
@@ -3,18 +3,19 @@ The maximum time for this test is 5 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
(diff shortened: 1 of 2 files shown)


nickswalker:
We voted to keep Clean Up #694, so we'll look at adding an element of Garbage to that task instead.

PR #682, opened 2019-07-25
REMOVE: Serve Breakfast task

moriarty:
## Description

We need to reduce the number of tasks.
- 7 teams opted into this task
- 2 attempted this task in Sydney
- 0 teams scored on this task

## Other comments
At least a few of the @RoboCupAtHome/technical-commmittee needs to agree on this, and maybe if needed discuss it. But otherwise it's ready to merge as soon as you've talked about it.

PR #690, opened 2019-08-29
REMOVE: Set Table

moriarty:
## Description

This task is admittedly one of the harder cuts to make.
But we must make the RoboCup@Home Competition clearer.
Perhaps it will be added back in the future, or an extension of test.
The robotics related components of this test are covered in other tests.

Related to #685
- 10 teams opted into this task before Sydney
- 1 teams attempted this task in Sydney
- 1 teams scored on this task in Sydney

Similar to other removal PRs: #682 #688 #689


kyordhel:
For the record: I'm against removing this one. It plays along with clean the table.

LoyVanBeek:
This challenge requires the very useful skill of manipulating small objects, which would be a reason to keep it. 
It's super clear to judge and for the audience, anyone can relate to what is happening. I even think there's some demand for a robot that can do this task well.

AFAIK TechUnited was the only team trying this and getting points, and we skipped grabbing the cutlery from a drawer and instead got it from the referee. Only the easy part was left, namely placing & dropping the cutlery.

To make the task more feasible, I'd put the cutlery on a different surface eg. the kitchen counter or a shelf and skip the drawer. 

The HSR should be perfectly capable of grabbing cutlery, with a camera in hand and force-sensing in the wrist.

moriarty:
@nickswalker ew sorry I tried the web editor to fix that merge conflict and I won't do that again...

moriarty:
Merge this with `Squash and merge` and it should be fine.

PR #695, opened 2019-08-30
Update stage TOCs

nickswalker:
Remove entries for already removed tasks

## Description

I failed to catch these dangling references in the first batch of task removals

### DIFF
--- a/Rulebook.tex
+++ b/Rulebook.tex
@@ -70,7 +70,6 @@
 
 \section*{Party Host tasks}
 \testtocentry{test:carry-my-luggage}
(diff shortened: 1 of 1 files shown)


kyordhel:
Each PR removing a test must also update the TOC. There's no need for this. If we decide to rollback, this will make it impossible.

nickswalker:
Well, that was my bad; all of those PRs are already merged. This just sets a clean base before removing the rest of the tasks. This needs to be merged before the remaining, then the other PRs will cleanly rebase (e.g. they'll have two commits each with contents just relevant to the task being removed).

PR #704, opened 2019-09-23
Storing Groceries: Reduce points for opening shelf door

johaq:
This PR contains minor spelling and grammar changes.
The main change is reducing bonus points for opening the door from 300 to 100.
In the TC meeting we considered making the door opening mandatory, but personally I think the league is not quite there yet. Afaik no OPL or SSPL robots opened the door (correct me here since I was not in Sydney). 

Something else I considered: Only award points for door opening when 2 or more objects are correctly moved to encourage teams to work on the main goal. I opted not to go for this since it adds an exception to a general rule which should be avoided as much as possible imo. But I could be convinced otherwise.

### DIFF
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -8,7 +8,7 @@ The maximum time for this test is 5 minutes.
 	\penaltyitem[5]{-100}{Receiving human help (move object)}
 
 	\scoreheading{Bonus rewards}
-	\scoreitem{300}{Opening the shelf door without human help}
+	\scoreitem{100}{Opening the shelf door without human help}
 	\scoreitem{100}{Moving a \emph{tiny} object}
 	\scoreitem{100}{Moving a \emph{heavy} object}
 
--- a/tasks/StoringGroceries.tex
+++ b/tasks/StoringGroceries.tex
@@ -1,18 +1,18 @@
 \section{Storing Groceries [Housekeeper]}
 \label{test:storing-groceries}
-The robot stores groceries into a pantry shelf while paying attention to sorting objects in their appropriate place, i.e.~storing an apple next to other fruits.
+The robot stores groceries into a pantry shelf while adhering to how objects are arranged in it, i.e. storing an apple next to other fruits.
 
 % \subsection*{Focus}
 % This test focuses on the detection and recognition of objects and their features, as well as object manipulation.
 
 \subsection*{Main Goal}
-Move objects 5 from a table into a shelf, grouping them by category or similarity.
+Move 5 objects from a table into a shelf, grouping them by category or similarity.
 
 \noindent\textbf{Reward:} 500pts (100pts per object).\\
 
 \subsection*{Bonus rewards}
 \begin{enumerate}[nosep]
-	\item Opening the shelf door (300pts)
+	\item Opening the shelf door (100pts)
 	\item Moving a \emph{tiny} object (100pts)
 	\item Moving a \emph{heavy} object (100pts)
 \end{enumerate}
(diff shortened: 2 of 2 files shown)


semaph0r:
Can we count door opening of the cabinet as part of the main goal? IMHO it is fair to award a team with some points if they were able to identify the door handle and open the doors. In fact, the robot is preparing it's environment to fulfill the complete task. Due to that, I vote for awarding team, which are able to open a fully closed cabinet. Even though, we are not tied to this and may remove this extra points if teams are confident in opening doors.

tasks/StoringGroceries.tex
```diff
@@ -1,18 +1,18 @@
 \section{Storing Groceries [Housekeeper]}
 \label{test:storing-groceries}
-The robot stores groceries into a pantry shelf while paying attention to sorting objects in their appropriate place, i.e.~storing an apple next to other fruits.
+The robot stores groceries into a pantry shelf while adhering to how objects are arranged in it, i.e. storing an apple next to other fruits.
```
kyordhel: please keep an `~` or `\,` after abbreviations such as e.g., i.e., Dr., Prof., etc. That's a good practice in LaTeX (aka Linter warning) to keep words together and increase readability.

scoresheets/StoringGroceries.tex
```diff
@@ -8,7 +8,7 @@
 	\penaltyitem[5]{-100}{Receiving human help (move object)}
 
 	\scoreheading{Bonus rewards}
-	\scoreitem{300}{Opening the shelf door without human help}
+	\scoreitem{100}{Opening the shelf door without human help}
```
kyordhel: I'd keep it in 300 for SSPL and perhaps OPL. DSPL has no big issue.

tasks/StoringGroceries.tex
```diff
@@ -1,18 +1,18 @@
...
 \subsection*{Main Goal}
-Move objects 5 from a table into a shelf, grouping them by category or similarity.
+Move 5 objects from a table into a shelf, grouping them by category or similarity.
 
 \noindent\textbf{Reward:} 500pts (100pts per object).\\
 
 \subsection*{Bonus rewards}
 \begin{enumerate}[nosep]
-	\item Opening the shelf door (300pts)
+	\item Opening the shelf door (100pts)
```
kyordhel: Same as above

kyordhel:
@semaph0r We could count door opening as part of the main goal yes, but I rather not for one single reason: **not screwing up teams**.

Adding this feature as part of the main goal means that if you fail to open the door, the test is over for you. Objects are pretty much standard when it comes to manipulation (all cans, bottles, and boxes are the same), but doors are not, so Burdeaux can be a bit more tricky than Sydney. Further, we should not raise the bar in 2020.

I'd keep that change for 2021 tho.

johaq:
Maybe do:
Tiny Object 200pts
Heavy Object 200pts
(Both never attempted or completed afaik)
Place on lowest shelf 50pts
Place on highest shelf 50pts

johaq:
I rounded out bonus points to 1000. I think giving more points for the heavy and tiny object is fair because so far no team did these. I'm against adding door opening to the main goal for the reason @kyordhel stated. I also think it would kinda screw up the concise task description

semaph0r:
> Place on lowest shelf 50pts
> Place on highest shelf 50pts
Are those rewards? What's the intended challenge?
Even though I can not find these in the PR, am I blind?

johaq:
Were just ideas to round out bonus challenge to 1000 points. Decided against those. Sorry, if I was not clear.

kyordhel:
@johaq if we're adding tiny and heavy objects, we need to be sure the description and constraints is added to the pertinent section in General Rules, for example

- Tiny object: an object which has no more than 5cm (2 inch) on each side
- Heavy object: any object weighting more than 750gr

Please edit first comment to summarize the changes applied in the PR

johaq:
I'm not adding those, just upping the points. Heavy and tiny object is defined in 3.3.6.

johaq:
Obsolete with #722 
Closed

PR #705, opened 2019-09-23
Carry my Luggage: Add DeM rule for finding operator (optional obstacles)

johaq:
This PR contains two main changes among some minor grammar and spelling ones:
1. Add DEM rule for finding the operator when lost. This is mainly for completeness sake, since this is a behavior I see some robots go for, so we have no discussions on site to which category of DEM this belongs.
2. Make placement of obstacles optional. I think we had this discussion on site in Magdeburg. Since avoiding obstacles is a bonus goal, teams should be able to choose not to go for them and focus on the main goal.

### DIFF
--- a/tasks/CarryMyLuggage.tex
+++ b/tasks/CarryMyLuggage.tex
@@ -1,12 +1,12 @@
 \section{Carry My Luggage [Party Host]}
 \label{test:carry-my-luggage}
-The robot helps the operator to carry some luggage to the car which is parked outside.
+The robot helps the operator to carry their luggage to a car which is parked outside.
 
 % \subsection*{Focus}
 % Object identification and manipulation, person-following, and social navigation.
 
 \subsection*{Main Goal}
-The robot helps the operator to carry a bag to the car parked outside.
+The robot helps the operator to carry a bag to a location outside the arena.
 
 \noindent\textbf{Reward:} 500pts.
 
@@ -21,9 +21,9 @@ The robot helps the operator to carry a bag to the car parked outside.
 \begin{itemize}[nosep]
 	\item \textbf{Location}: The test takes place inside and outside the arena.
 
-	\item \textbf{Start Location}: The robot starts at a predefined location in the living room.
+	\item \textbf{Start Location}: The robot starts at a predefined location.
 
-	\item \textbf{Bags:} At least two bags are placed between nearby the operator (within 2m distance and visible to the robot).
+	\item \textbf{Bags:} At least two bags (see \ref{rule:scenario_objects_list}) are placed near the operator, visible to the robot.
 
 	\item \textbf{Operator}: The operator is standing in front of the robot pointing at the bag to be carried out.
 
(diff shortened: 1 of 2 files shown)


tasks/CarryMyLuggage.tex
```diff
@@ -33,11 +33,12 @@ \subsection*{Additional rules and remarks}
...
 	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied as follows.
 	\begin{itemize}[nosep]
-		\item Handing-over the bag (-100 pts).
-		\item Finding the operator while following her
+		\item Handing the bag over (-100 pts).
+		\item Finding the operator while following them:
 		\begin{itemize}[nosep]
 		\item Natural interaction (e.g.~wave and calling) 100pts score reduction.
 		\item Non-natural interaction (e.g.~raising both hands and jumping) 200pts score reduction.
+		\item Moving back (e.g.~ask the operator to step back in front of the robot) 300pts score reduction.
```
YuqianJiang: Why does this request get more score reduction than non-natural interactions? I think it can just count as a non-natural interaction.

johaq: In the past it was straight out forbidden for the operator to move back whereas requesting to raise both arms was fine. So I thought it should be penalized more heavily. What do the others think? @RoboCupAtHome/technical-commmittee

nickswalker: A distinction between natural and moving-back makes a lot of sense to me; I would rather clap than have to walk back because effort. Distinction between non-natural and moving back is less clear; I think I would rather move back than jump up and down, but raising both hands sounds easier than walking... So it may fit fine in that category, but the name would have to better describe what is implied; more-than minimal user effort required

tasks/CarryMyLuggage.tex
```diff
@@ -48,10 +49,10 @@ \subsection*{Additional rules and remarks}
 		\item a small object on the ground,
 		\item a hard-to-see 3D object,
 		and
-		\item a small area blocked using retractable barriers.
+		\item a small area blocked using retractable barriers. Teams can choose to have no obstacles placed if they do not want to go for bonus points. Teams can not pick and choose which obstacles are placed. Either none or all of them.
```
nickswalker: "Retractable barrier" has [a specific meaning](https://www.amazon.com/Retractable-Wall-Barrier-Black-Finished/dp/B009YEE22U) that may or may not be what you were going for.

> Teams can choose to have no obstacles placed if they do not want to go for bonus points. Teams can not pick and choose which obstacles are placed. Either none or all of them.

The last two sentences are obvious enough from the first sentence that I think it's fine to drop them

johaq: I did not add that but yes, that is what is meant there. Simply because we often have those at a competition.

ARTenshi: I would put it back:

"Teams can choose to have no obstacles placed if they do not want to go for bonus points. Teams can not pick and choose which obstacles are placed. Either none or all of them."

Because some teams expected these obstacles to be there but they weren't, at least as a reminder for TC to discuss during the Team Leaders meeting.

johaq:
Obsolete with #722 
Closed

PR #706, opened 2019-09-25
Clean Up: add take out the garbage bonus goal

johaq:
Among some grammar and spelling changes, the main change is to replace opening the door with taking out the garbage as bonus goal. Scoring of bonus goals has been modified to be consistent with storing groceries. I think that taking out the garbage is more in line with the objective of the task than opening the door. Also, opening the door is already a bonus goal in the receptionist task.

### DIFF
--- a/tasks/CleanUp.tex
+++ b/tasks/CleanUp.tex
@@ -65,7 +65,7 @@ The referee needs to
 \subsection*{OC instructions}
 2 hours before the test:
 \begin{itemize}
-	\item Announce the starting location of the robot.
+	\item Announce the location of the garbage collection zone.
 \end{itemize}
 
 \subsection*{Score sheet}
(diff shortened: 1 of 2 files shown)


johaq:
I just notices I needed to add instruction to announce collection zone. Also announcing starting location is already covered in general rules

tasks/CleanUp.tex
```diff
@@ -65,7 +65,7 @@ \subsection*{Referee instructions}
 \subsection*{OC instructions}
 2 hours before the test:
 \begin{itemize}
-	\item Announce the starting location of the robot.
+	\item Announce location of garbage collection zone.
```
kyordhel: Announce **the** location of **the** garbage...
Ist nicht deutsch!!! Du musst Artikel verwenden :)

johaq: Ah I dont't think they are really required. But maybe a bit nicer to read.

nickswalker:
I like these changes!

A have a high level suggestion for this task: rename it to "Tidy Up". "Clean Up" is almost a synonym, but it has a connotation of cleaning something dirty (maybe see [this Wikipedia list of types of cleaning](https://en.wikipedia.org/wiki/Cleaning) to get the sense). Toyota named their [very similar task](https://worldrobotsummit.org/wrs2020/challenge/download/Rules/DetailedRules_Partner_EN.pdf) Tidy Up as well.

johaq:
Obsolete with #722 
Closed

PR #707, opened 2019-09-25
Receptionist: New bonus goal and clarifications

johaq:
Two main changes:
1. Replace seating the oldest person with describing a guest as bonus goal. We rarely have a clearly oldest person as volunteer at the competition so I think describing a guest is not only more practical but also more useful.
2. Clearing up some misunderstandings that occured at GO:
- Is the favorite drink of John also announced beforehand
- How does the robot know when guests arrive


johaq:
Which change do you disagree with? 
It fits on two pages.

kyordhel:
Removing politeness towards elderly and the elder detection feature, but I get your point, we can't ensure grumps is going to be there even though we have Jesus Savage around

PR #708, opened 2019-09-25
GPSR: Remove recording bonus goal

johaq:
Main change: I removed the bonus score for recording audio. This was always confusing to teams in how it works. Also not that useful imo and not understandable for audiences. GPSR also rounds out to a 1000 points now in line with other Stage 1 tasks.
Some clarifications: 
-Which abilities can make up a task
-DEM scoring
-What does successful execution mean for scoring bonus of leaving the arena

Removed description of task start because it is covered in general rules.

### DIFF
--- a/scoresheets/GPSR.tex
+++ b/scoresheets/GPSR.tex
@@ -6,9 +6,8 @@ The maximum time for this test is 5 minutes.
 	\penaltyitem[3]{50}{Using custom operator or bypassing ASR}
 
 	\scoreheading{Bonus rewards}
-	\scoreitem[3]{100}{Understand command given by naive operator}
-	\scoreitem[3]{100}{Provide audio recording and transcript}
-	\scoreitem{150}{Autonomously leaving the arena}
+	\scoreitem[3]{50}{Understand command given by naive operator}
+	\scoreitem{100}{Autonomously leaving the arena}
 
 	%\setTotalScore{1000}
 \end{scorelist}
--- a/tasks/GPSR.tex
+++ b/tasks/GPSR.tex
@@ -1,20 +1,19 @@
 \section{General Purpose Service Robot [Housekeeper]}
 \label{test:gpsr}
-Similar to a modern smart-speaker, the robot can be asked to do anything from the Stage~I of this rulebook or any previous rulebook.
+The robot can be asked to do anything involving abilities required in Stage~I of this rulebook.
 
 % \subsection*{Focus}
 % This test focuses on the detection and recognition of objects and their features, as well as object manipulation.
 
 \subsection*{Main Goal}
-Execute each of the 3 commands requested by the operator.
+Execute 3 commands requested by the operator.
 
 \noindent\textbf{Reward:} 750pts (250 points per command)\\
 
 \subsection*{Bonus rewards}
 \begin{enumerate}[nosep]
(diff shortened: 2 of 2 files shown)


kyordhel:
The audio recording thingy comes is part of the roadmap, aiming to create a training corpora of natural interactions. I wouldn't like to see it removed (perhaps lowering the score or zero-it but make the data retrieval mandatory would work).

@justinhart, @okadahiroyuki, @balkce, you are also in audio/nlp. What do you think about removing recordings?

johaq:
I think that if that was the goal we would at least need to define some standards for the corpus. Like how to record, file format standards, encoding etc.

kyordhel:
> I think that if that was the goal we would at least need to define some standards for the corpus. Like how to record, file format standards, encoding etc.

That would come handy. We had that already for the data recording in Sp&AD test but w need to do it explicit. Same as with the standard objects, that info should go in the wiki.

johaq:
It would also need to be labeled. Do the teams also have to do that? I'm not convinced this is something that should be "enforced" by the rulebook. If this is someone's research topic feel free to use robocup as an opportunity to collect recordings but you need to convince the teams and ideally provide them with the required software for recording yourself.

kyordhel:
Audio + transcript, no labeling.

Ideally, as a scientific competition, @Home should also produce datasets

johaq:
Hm, ok. I still think this is something that should not belong to a specific task as a bonus goal. Why only do recordings of GPSR? Commands might be complex but other tasks might offer different and interesting challenges like multiple persons standing in front of the robot, persons talking to each other in front of the robot etc.
I can talk to a speech guy about what would be interesting and annotation standards. I do think this is something that does not belong in the rulebook but should rather be something extra we try to acquire on the side. Use GO as test run for example.

nickswalker:
I agree that without a useful standard, there is little utility to gathering data from teams. It is more effort to standardize than it would be for a determined group to produce the data on their own. Even with complete buy in and a standard format, it is unlikely we would be able to gather a meaningful amount of data ([comparable speech datasets](http://spandh.dcs.shef.ac.uk/chime_challenge/CHiME5/data.html) sit around 50 hours of audio, 90K utterances).

Another obstacle is that mid-competition interactions _are not natural and do not reflect conditions that exist in any home in the world_. The league needs to produce datasets that non-RoboCup researchers would find interesting, but the audio environment is a deadly combination of difficult and too specific to have broader relevance.

kyordhel:
The data recording step is crucial for the NLP roadmap (doi: 10.1007/978-3-030-27544-0_18). It may seem pointless in the first stages but these datasets are cornerstones for a real GPSR with unconstrained interaction (i.e. no command generator whatsoever). I'm well aware of all you're pointing out.

Can we remove it? Yes, surely! Nobody is forcing us to follow a roadmap, so TC can vote against and approve the PR. I would advise otherwise and rather add specs. Roadmap aside, I think it's about time that @Home starts producing some data and tech, because so far we're the only tech-consumer-only league in all RoboCup.

moriarty:
I think the bonus points can be removed after reading the pros and cons listed above.

If it is truly important than teams who research in this area can communicate with themselves and collaborate with themselves to organize and collect this data, but it doesn’t need to be pushed onto all the other teams.

johaq:
Obsolete with #722 
Closed

PR #716, opened 2019-12-31
Use personal access token for deployments

nickswalker:
## Description

Turns out, the provided GITHUB_TOKEN doesn't trigger GitHub Pages builds (even though it can push successfully). [It's not clear](https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/td-p/26869) if this is a bug or a feature intended to prevent recursive triggers.

- Use a personal access token when pushing to `gh-pages`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,5 +89,5 @@ jobs:
       if: github.ref == 'refs/heads/master'
       uses: ad-m/github-push-action@v0.5.0
       with:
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
Not working yet: https://github.com/RoboCupAtHome/RuleBook/runs/368922564#step:7:8

nickswalker:
My bad; had accidentally entered the secret under the wrong name. Fixed now. Thanks for patience, all.

PR #722, opened 2020-02-16
Overhaul and Trimming of Rules

johaq:
This is my push to get a final version for 2020. 
Before you all panic, though this PR seems a lot, most changes are:
Removal of rules that were never enforced
Removal of sections that were duplicate or even contradictory
Making things more coherent. In terms of formatting and structure

One of my big goals was to reduce the size of the rulebook significantly which was requested a lot in the past. I reduced the rulebook from 72 to 60 pages, from ~20.8k to ~14.5k words.

Overview of changes:
- Removal of skill certificates. Don't make sense without basic ability tests.
- Removal of SPL focuses. Don't make sense when all leagues have the same rules
- Removed lean set of rules concept. Self-explanatory
- Removed desired abilities in concepts. Already covered in Introduction and in test focuses
- Moved website tips to wiki. Rulebook is no place for tips and tricks
- Removed task announcements. Not necessary with reduced number of tasks
- Removed powered fridge from scenario. Afaik we never had a powered fridge or used it in a test
- Furniture specs were way too narrow. Contradicts key concept of non-standard scenario
- Formalize changes to arena during setup days. Was usually done but not in the rules
- Removed contradictions in External Computing. No wireless allowed but laptops ok
- Reduced number of external computing devices to 2. 5 cannot be supported in the ECRA
- Usage of Laptop in DSPL not mandatory. Why would we punish teams for this?
- Moving DSPL addons. They are not external devices
- Removed scoring limitations for tests. Those are more internal guidelines
- Added partial scoring to tests when applicable. Therefore, changed bonus goals to be attainable when more than 50% were scored in the main goal
- Removed not trying to solve and just score. This is confusing to new teams and means we designed the test poorly. Also covered in exploiting the rules.
- Removed gesture section. Imo just not necessary. Gestures should be easy enough to be explained by the robot. If you need the robot inspection to explain gestures to the referees the gesture is too complicated.
- Split referee and volunteer section. Was confusing
- Removed default time limits. Useless if every test states its own time limit regardless
- Removed restart. No test allows one
- Removed faking robot rule. Never applied and not enforceable
- Removed outstanding performance. Never awarded in recent years. Contradictory to other parts emphasizing wanting teams to focus on the main task
- Removed grouping of teams. Never used
- Renaming of poster session and poster teaser sessions to reflect how these terms are usually used at the competition
- No strict time limit of robot inspection. Makes no sense, since teams can have lots of ext devices and similar. Time taken dependent on referees understanding and noting them done. 
- Introduced focus section for each test. Was proposed by @justinhart if I recall correctly
- Make scoring equal for stage 2 tests. 1000+750 (except EGPSR)
- Rework Restaurant test to fit other tests.
- Rename to P&G test
- Removed open challenge. If we want this, there needs to be a place in the schedule
- Added team leader meeting. Usually done but not in the rules

TODOS: 
- Add names of contributors
- Decide on themes for the finals


johaq:
Made formatting and scoring of DeM more consistent between tests.
Incorporated feedback by @swachsmu

moriarty:
This one is too long to review from my phone 🤣 I’m sure it’s fine but someone should do a full review if this is the final copy.

nickswalker:
Made it through the first three chapters tonight and appreciated how much more concise the text was. Great work. I think we should cull back the table of contents to make things more navigable, mostly by reducing the entries for chapters 1 and 2. The Clean Up -> Tidy Up rename was breaking the score sheet build so also fixed that and put the changes on [a branch](https://github.com/RoboCupAtHome/RuleBook/tree/johaq/master) if you want to add those commits.
Will look through the rest in the coming days.

scoresheets/StoringGroceries.tex
```diff
@@ -3,14 +3,14 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[5]{100}{Move an object next to their peers in the shelf}
-	\penaltyitem[5]{-30}{Receiving human help (point at target location)}
-	\penaltyitem[5]{-100}{Receiving human help (move object)}
+	\scoreitem[5]{100}{Move an object next to similar objects in the shelf}
+	\penaltyitem[5]{-40}{Request object locations by gesture}
+	\penaltyitem[5]{-60}{Request object interaction}
```
LoyVanBeek: I'd specify 'interaction by a human'

tasks/CleanTable.tex
```diff
@@ -1,80 +1,72 @@
-\section{Clean the Table [Housekeeper]}
+\section{P\&G Challenge}
```
LoyVanBeek: Do they still sponsor us?

tasks/HandMeThat.tex
```diff
@@ -1,91 +1,61 @@
-\section{Hand Me That [Party Host]}
+\section{Hand Me That}
 \label{test:hand-me-that}
-A guest at the party speaks English, but with only a limited vocabulary. The robot will assist them in obtaining things that they gesture for.
+A guest asks the robot to identify objects in the household.
```
LoyVanBeek: This is a confusing (and disappointing?) test for the audience, as they expect the robot to actually bring the pointed-at object. 
Maybe it already helps for the robot to ask if you want that object and for the operator to only confirm this for the 5th object. Then, getting to the 5th object and awarding points for the 5th object to be grasped is an incentive to get to 5 objects fast.

general_rules/AudienceInteraction.tex
```diff
@@ -0,0 +1,7 @@
+\section{Audience Interaction}
+\label{sec:rules:audience}
+Only some tests require direct interaction with the audience. In order to make all tests more appealing (see \ref{sec:concepts:appel}) and engaging, informing the audience about what is happening is important.
+
+\subsection{Visualization}
+\label{sec:rules:audience:visualization}
+All scenarios will have a screen, visible to the audience, that the teams can use to display, e.g., a robot's state, world model, and current plan. An example of such a visualization is the \VizBox{} by Loy van Beek (\url{https://github.com/LoyVanBeek/vizbox}). Using and enhancing the capabilities of the \VizBox{} or developing another open source visualization is greatly appreciated and considered a contribution to the \RoboCup\AtHome{} community.
```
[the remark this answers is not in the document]
johaq: Section 3.3.3 Furniture lists that a TV is needed. Is that what you mean? Or just having an additional separate document you could give to LOC?

general_rules/ContinueRules.tex
```diff
@@ -1,117 +1,65 @@
-\section[Deus ex Machina]{Deus ex Machina: Bypassing features with human help \\ \small Because the show must go on}
-\label{rule:continue}
-Robots can't score unless they accomplish the main goal of a task.
-However, in many real-life situations, a minor malfunction may prevent the robot from accomplishing a task.
-To prevent this situation, while fostering awareness and human-robot interaction, robots are allowed to request human assistance during a test.
+\section{Deus ex Machina}
+\label{sec:rules:deusexmachina}
+To allow teams to participate in tests, where the robot is only missing one ability to complete the main goal, and to allow robots to continue in case of minor malfunctions, a robot can request human assistance during a test, called \DEM{}. Help can not be requested to complete bonus goals.
```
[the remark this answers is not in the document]
johaq: Yeah, I can see that. I believe part of the idea was to allow newer teams to still score, with partial scoring removed. That's why I wrote it that way.

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
-	\item living room, and
-	\item kitchen.
+	\item Bedroom
+	\item Dining Room
+	\item Living Room
+	\item Kitchen
 \end{itemize}
-Depending on the Local Organization, there may be multiple apartments which may be different to each other.
-Robot must be prepared to perform any task in any arena, not the same arena every time.
+There is one \Arena{} for each league and an additional one for setup and training shared by all leagues.
```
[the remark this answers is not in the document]
johaq: Fair. In my mind the section describes the world cup. Smaller competitions can adjust depending on their resources. But this raises a good point, that smaller competitions should probably publish in advance where they might deviate from the rulebook. The way I see it, German Open only offers OPL anyway so one arena is fine, but it should be stated that no training arena will be available.

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
 	\centering
 	\subfloat[Typical arena]{\label{fig:scenario_arena}\includegraphics[height=46mm]{images/typical_arena.jpg}} ~
 	\subfloat[Typical objects]{\label{fig:scenario_objects}\includegraphics[height=46mm]{images/typical_objects.jpg}}
-	\caption{Scenario examples: (a) a typical arena, and (b) typical objects.}
+	\caption{Example \RoboCup\AtHome{} scenario.}
 	\label{fig:arena}
 \end{figure}
 
-
-\subsubsection{Cupboard}
```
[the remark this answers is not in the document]
johaq: No, an over specified cupboard is not needed. Cupboard is still in the list of needed furniture, but in my mind there is no need to prescribe what distance the shelfs have to have from each other. Robots should be able to deal with any normal cupboard.

scoresheets/StoringGroceries.tex
johaq: changed

tasks/CleanTable.tex
johaq: Renewed I guess. EC could give more details.

tasks/HandMeThat.tex
johaq: Yes, I already started a discussion in the TC group about changing this test to a robot does inventory test. Ca: Operator shows coke, robot needs to find out how many coke there are in the kitchen, bonus if it finds out how many drinks in general there are. Question is, are teams still fine with a heavy rework of a test at this point in time?

johaq:
Incorporated feedback by Nick and Loy.
Merged proposed changes by Nick

general_rules/AudienceInteraction.tex
```diff
@@ -0,0 +1,7 @@
+\section{Audience Interaction}
+\label{sec:rules:audience}
+Only some tests require direct interaction with the audience. In order to make all tests more appealing (see \ref{sec:concepts:appel}) and engaging, informing the audience about what is happening is important.
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/HandMeThat.tex
LoyVanBeek: Nope, I would not rework a test in such a major way for such little reward at this point.

johaq: Yeah fair enough. I think I still want to change the name. Currently its not really describing the test. That would probably also make it less confusing for audiences.

scoresheets/GPSR.tex
```diff
@@ -2,13 +2,14 @@
...
-	\scoreitem[3]{100}{Provide audio recording and transcript}
-	\scoreitem{150}{Autonomously leaving the arena}
+	\scoreitem[3]{250}{Complete a command}
+	\penaltyitem[3]{50}{Receiving command not from the instruction point}
+	\penaltyitem[3]{50}{Custom Operator}
+	\penaltyitem[3]{75}{Alternative HRI}
+	
+	\scoreheading{Bonus Rewards}
+	\scoreitem[3]{50}{Understand command given by naive operator}
+	\scoreitem{100}{Autonomously leaving the arena after completing all 3 commands}
```
nickswalker: I've never really understood the reasoning for this bonus; if the robot was able to complete three arbitrary commands, it can surely navigate to a waypoint outside the arena. Would rather see these points put on something more interesting; either more points for the naive operator or something new.

tasks/CarryMyLuggage.tex
```diff
@@ -1,84 +1,82 @@
-\section{Carry My Luggage [Party Host]}
```
nickswalker: I notice the theme tags have been dropped. Even though most are obvious, I think there should be a listing somewhere to be unambiguous. A placement with less emphasis on the this page makes sense to me

Rulebook.tex
```diff
@@ -62,27 +62,29 @@
...
+\section*{Party Host}
 \testtocentry{test:carry-my-luggage}
 \testtocentry{test:receptionist}
 
 % Uncomment to get aesthetic improvement
-\cleardoublepage
-\input{tasks/CarryMyLuggage}
 \newpage
-\input{tasks/CleanUp}
+\input{tasks/TidyUp}
```
nickswalker: Why change the order? It's not clear from the titles alone why they're ordered this way (it's because of the themes, but the themes aren't listed anywhere so I don't think the reader would get it).

tasks/Receptionist.tex
```diff
@@ -1,74 +1,79 @@
...
 
 \noindent\textbf{Reward:} 500pts (250pts per guest).
 
-\subsection*{Bonus rewards}
+\vspace{-5pt}
+\subsection*{Bonus Rewards}
 \begin{enumerate}[nosep]
-	\item Opening the entrance door to a guest (200pts each)
-	\item Describe the first guest to the second guest  (100pts)
+	\item Opening the entrance door to a guest (200pts each).
```
nickswalker: Same deal

scoresheets/Receptionist.tex
```diff
@@ -1,14 +1,13 @@
...
 	\scoreheading{Main Goal}
-	\scoreitem{100}{Introduce the 1st guest to John}
-	\scoreitem{150}{Offer an empty seat to the 1st guest}
-	\scoreitem{100}{Introduce the 2nd guest to John and 1st guest}
-	\scoreitem{150}{Offer an empty seat to the 2nd guest}
-	\penaltyitem[2]{50}{Misunderstanding the name of a guest}
+	\scoreitem[2]{250}{Introduce a new guest to every other guest and offer a seat}
+	\penaltyitem[2]{50}{Continue with wrong name or drink}
+	\penaltyitem[2]{75}{Alternative HRI}
+	\penaltyitem[2]{200}{Recognizing People}
```
nickswalker: This item should be named more similarly to the storing groceries DeM penalties; at first glance it looks like recognizing people somehow causes a penalty. "Request person recognition"

scoresheets/GPSR.tex
semaph0r: +1

Rulebook.tex
moriarty: Maybe they're now ordered in increasing difficulty? Or just moving GPSR to the end out of tradition.

scoresheets/GPSR.tex
johaq: The reward is not for being able to leave, but to be able to solve 3 commands in time. Leaving just signifying that the bonus goal has been achieved. So the bonus is more for being really fast. It's difficult to make bonuses for GPSR since you don't know which commands will be generated.

tasks/CarryMyLuggage.tex
johaq: Listing is at the start of Stage 1 and Stage 2 chapters

Rulebook.tex
johaq: I believe they were just ordered inconsistently. The order in the toc was different to the order in the text and I just happened to make the TOC fit the text. No real reason for the order.

general_rules/ContinueRules.tex
```diff
@@ -1,117 +1,65 @@
...
-	\item \textbf{Request help:} The robot must indicate loud and clear that it requires human assistance. It must be clearly stated:
-	\begin{compactitem}
-		\item The nature of the assistance
-		\item The particular goal or desired result
-		\item How the action must be carried out (when necessary)
-		\item Details about how to interact with the robot (when necessary)
-	\end{compactitem}
+	\item \textbf{Request Help:} The robot has to indicate loud and clear that it requires human assistance. It must clearly state:
+	\begin{enumerate}
+		\item The nature of the assistance.
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/TidyUp.tex
```diff
@@ -0,0 +1,68 @@
...
+	
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per object as follows:
+	\begin{itemize}[nosep]
+		\item \textbf{Request Object Locations by Speech:} Telling the robot where an object can be found or should be placed causes a reduction of 30pts.
+		
+		\item \textbf{Request Object Locations by Gesture:} Pointing at an object to be picked up or where to place an object causes a reduction of 40pts.
+		
+		\item \textbf{Request Object Locations by Guiding:} Guiding the robot nearby the location where an object can be found or should be placed causes a reduction of 20pts.
+		
+		\item \textbf{Request Object Interaction:} Having a human pick up or place an object causes a reduction of 60ptst.
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/StoringGroceries.tex
```diff
@@ -1,39 +1,44 @@
-\section{Storing Groceries [Housekeeper]}
+\section{Storing Groceries}
 \label{test:storing-groceries}
-The robot stores groceries into a pantry shelf while paying attention to sorting objects in their appropriate place, i.e.~storing an apple next to other fruits.
+The robot stores groceries into a pantry shelf while adhering to how objects are arranged in it, i.e., storing an apple next to other fruits.
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/StoringGroceries.tex
```diff
@@ -42,36 +47,41 @@ \subsection*{Setup}
...
+\subsection*{Additional Rules and Remarks}
 \begin{enumerate}
-		\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per object as follows:
+	\item \textbf{Partial scoring:} The main task allows partial (per object) scoring.
+	\item \textbf{Deus ex Machina:} Score reduction for requesting human assistance is applied per object as follows:
 	\begin{itemize}[nosep]
-		\item Telling or pointing out to the robot where to place an object results in a score reduction of 30pts.
+		\item \textbf{Request Object Locations by Gesture:} Pointing out where to place an object causes a reduction of 40pts.
+		
+		\textbf{Request Object Interaction by a Human:} Having a human pick up or place an object causes a reduction of 60ptst.
```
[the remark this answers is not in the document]
johaq: Fixed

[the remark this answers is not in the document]
johaq: Matched Tidy Up to this

tasks/GPSR.tex
```diff
@@ -1,34 +1,34 @@
...
-% \subsection*{Focus}
-% This test focuses on the detection and recognition of objects and their features, as well as object manipulation.
+\noindent \textbf{Focus:} \SysI.
 
 \subsection*{Main Goal}
-Execute each of the 3 commands requested by the operator.
+Execute 3 commands requested by the operator.
 
-\noindent\textbf{Reward:} 750pts (250 points per command)\\
+\noindent\textbf{Reward:} 750pts (250 points per command).
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/Receptionist.tex
johaq: Fixed

scoresheets/Receptionist.tex
johaq: Changed

scoresheets/GPSR.tex
nickswalker: Ah, I see. It makes sense when explained. I would suggest lowering the bar ever so slightly then to being exactly that; "completing all 3 commands."

tasks/StoringGroceries.tex
```diff
@@ -1,39 +1,44 @@
...
+The robot stores groceries into a pantry shelf while adhering to how objects are arranged in it, e.g., storing an apple next to other fruits.
+
+\noindent \textbf{Focus:} \SysI, \MAN, \CV, \OR.
 
 % \subsection*{Focus}
 % This test focuses on the detection and recognition of objects and their features, as well as object manipulation.
 
 \subsection*{Main Goal}
-Move objects 5 from a table into a shelf, grouping them by category or similarity.
+Move 5 objects from a table into a shelf, grouping them by category.
```
YuqianJiang: Removing "or similarity" here makes it a bit inconsistent with "move an object next to similar objects in the shelf" in the scoresheet. For example, is it allowed to group unknown objects by size, shape, color, etc.?

johaq: Yeah, grouping by size, shape, color is NOT ok. At least that is not how I usually arrange my pantry. That's why I changed it there. Will adjust the scoresheet as well.

Rulebook.tex
nickswalker: I think the reasoning for the old TOC being out of order was that, if you're flipped to it, you just read the page number for the task you want (or click on it) and it's fine if that number is out of sequence. The TOC doesn't even need to be alphabetized really because there aren't many entries. What's more important is that the tasks are ordered alphabetically in page sequence, because if you open to Carry my luggage, you can then know for sure which direction you need to page/scroll to get to GPSR.

tasks/Restaurant.tex
```diff
@@ -1,114 +1,79 @@
...
-\noindent\textbf{Reward:} 500pts.
+\noindent\textbf{Reward:} 1000pts (500pts per order)
 
-\subsection*{Bonus rewards}
+\subsection*{Bonus Rewards}
 \begin{enumerate}[nosep]
-	\item Detect a customer's service request: 200pts (100pts each)
-	\item Reach a customer's table without prior guidance/training: 200pts (100pts each)
-	\item Use an unattached tray to transport an order: 500pts (250pts each).
+	\item Use an unattached tray to transport an order(250pts each)
```
nickswalker: space after order

scoresheets/Restaurant.tex
```diff
@@ -1,24 +1,16 @@
...
-\small\begin{scorelist}
+\small\begin{scorelist}[startbutton=false]
 	\scoreheading{Main Goal}
-	\scoreitem{500}{Take and serve an order}
-	\penaltyitem[2]{200}{Being guided to the \emph{Kitchen-bar} or to a customer's table}
-	\penaltyitem{100}{Not making eye-contact when taking an order}
-	\penaltyitem[4]{50}{Bypassing object manipulation (handover)}
-	\penaltyitem[5]{100}{Bypassing object manipulation}
+	\scoreitem[2]{500}{Take and serve an order}
+	\penaltyitem{25}{Request help by receiving directions}
```
nickswalker: Request guidance should be here too

nickswalker: These should all be 2x-ed

tests/Finals.tex
```diff
@@ -1,85 +1,77 @@
...
-%To avoid logistical issues during the last day of the competition, the \iterm{Finals} are divided into two sets of demonstrations: the Bronze Competition and the RoboCup @Home Grand Finale.
-%The Bronze Competition is a set of demonstrations that are carried out before the RoboCup @home Grand Finale. Here, all the leagues run in parallel, with the fourth and third highest scored teams competing for the bronze.
-%Finally, the two teams with the highest score in each League present their demonstrations in a serialized manner during the RoboCup @Home Grand Finale.
-
-Even though each league has its own first, second and third place, the \iterm{Finals} are meant to show the best of all leagues to the jury members as well as the audience and, thus, warrants a single schedule slot.
+\label{chap:finals}
+\raggedbottom
+The competition ends with the \FINAL{} on the last day, where the two teams with the highest total score compete. The \FINAL{} are conducted as an open, themed demonstration. The \FINAL{} of each league is scheduled at different times.
 
 \section{Structure and Theme}
```
nickswalker: Since this section is very compact now, would recommend using \section*{} to cut down TOC entries.

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \section{Scenario}
```
nickswalker: "Scenario" is a strange word choice here. More accurate would be "Setting" or "Environment", but probably best would just be "Arena".

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
 
 \begin{enumerate}
-	\item \textbf{\iterm{Known objects}:} Objects previously known by the robot and that it can identify and manipulate.
-	There are two kinds of known objects:
+	\item \textbf{\KnownObjects{}:} Objects previously known by the robot, split into:
 	\begin{enumerate}
-		\item \textbf{\iterm{Regular objects}:} Objects with no noticeable difference among peers (e.g.~soda can, cereal box, cutlery, etc).
-		\item \textbf{\iterm{Alike objects}:} Objects which are different one from another, but still considered by people to be the same (e.g.~apple, sandwich, cloth, etc.).
+		\item \textbf{\RegularObjects{}:} Objects that always look the same.
+		\item \textbf{\AlikeObjects{}:} Objects that look different to the provided image but are still considered the same by people (e.g. differently colored apple, cloth with different pattern).
```
nickswalker: The "alike/regular" distinction doesn't appear elsewhere in the rulebook this year, would recommend dropping it, as it's a non-obvious name.

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
-	\item \textbf{\iterm{Fragile object}:} An easy-to-break object, (e.g.~chocolate egg).
-	\item \textbf{\iterm{Amorphous object}:} An flexible object that may take an infinite number of shapes (e.g.~cloth, magnetic puzzle, etc.).
-	\item \textbf{\iterm{Garbage bag}:} A tie-able garbage bag.
+	\item \textbf{\iterm{Bags}:} Lightweight. With stiff, vertical handles (See~\reffig{fig:scenario_container_bag}).
+	\item \textbf{\iterm{Trays}:} Tray or basket, intended for two-handed manipulation (See~\reffig{fig:scenario_container_tray}).
+	\item \textbf{\iterm{Pourable}:} An object whose content can be poured (e.g.~jug).
+	\item \textbf{\iterm{Heavy Object}:} Weight between 1.0kg and 1.5kg (e.g.~water bottle).
+	\item \textbf{\iterm{Tiny Object}:} A lightweight object, no bigger than 5cm (e.g.~teabag).
+	\item \textbf{\iterm{Fragile Object}:} An easy-to-break object (e.g.~egg).
+	\item \textbf{\iterm{Amorphous Object}:} A flexible object that may appear in different shapes (e.g.~cloth).
```
nickswalker: The standard term here would be "deformable." Random google result exemplifying common usage: https://arxiv.org/abs/1910.13439

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
 \begin{enumerate}
-	\item \textbf{\iterm{Known objects}:} Objects previously known by the robot and that it can identify and manipulate.
-	There are two kinds of known objects:
+	\item \textbf{\KnownObjects{}:} Objects previously known by the robot, split into:
 	\begin{enumerate}
-		\item \textbf{\iterm{Regular objects}:} Objects with no noticeable difference among peers (e.g.~soda can, cereal box, cutlery, etc).
-		\item \textbf{\iterm{Alike objects}:} Objects which are different one from another, but still considered by people to be the same (e.g.~apple, sandwich, cloth, etc.).
+		\item \textbf{\RegularObjects{}:} Objects that always look the same.
+		\item \textbf{\AlikeObjects{}:} Objects that look different to the provided image but are still considered the same by people (e.g. differently colored apple, cloth with different pattern).
+		\item \textbf{\StandardObjects{}:} Objects chosen from the \YCBData{}\footnote{\url{http://www.ycbbenchmarks.com/object-set/}}. They are published 6 months in advance on the \RoboCup\AtHome{} website\footnote{\url{https://athome.robocup.org/standard-objects}} so that they can be aquired and trained beforehand.
```
nickswalker: Instead of the other two categories, would have "non-standard" and say that they are purchased in the country and made available during setup days (reference to set up days section).

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
 	\end{enumerate}
 
-	\item \textbf{\iterm{Unknown objects}:} Any other object that is not known beforehand but can be grasped or handled.
+	\item \textbf{\UnknownObjects{}:} Any other object that is not known beforehand but can be grasped or handled (e.g. arena decorations).
```
nickswalker: Wouldn't the arena decorations be largely known? It should be clearer that these are additional "held out" objects that the TC can insert into tasks/arena during competition days.

general_rules/ContinueRules.tex
```diff
@@ -1,117 +1,65 @@
...
-\end{itemize}
-
-
 
 \subsection{Scoring}
-\label{rule:continue_scoring}
-There is no limit in the amount of times a robot can request human assistance, but score reduction applies every time it is requested.
-
+\label{sec:rules:demscoring}
+The amount of times a robot can request human assistance is not limited, but score reduction applies each time. The score for a goal or partial goal cannot be negative due to \abb{DEM} reductions. Possible \abb{DEM} requests are stated for each test with the corresponding score reductions. Other requests must be announced one day before the test in the \TLM{} where the \abb{TC} will decide on the scoring. In general, points will be deducted increasingly for:
```
nickswalker: \abb{TLM}, or just Team Leader Meeting. Don't think I've ever seen anyone say TLM.

nickswalker:
For people who want PDFs of this or any PR, reminder that you can click into the GitHub Actions entry for a given commit and download the built PDF from its artifacts listing.

tasks/CleanTable.tex
```diff
@@ -1,80 +1,72 @@
...
+		\item \textbf{Request Object Locations by Gesture:} Pointing at an object to be picked up or where to place an object causes a reduction of 40pts.
 
-	\item \textbf{Dishwasher:} The team decides whether the robot will place the objects in the dishwasher's rack or in the official tray.
-
-	\item \textbf{Dishwasher door:} The dishwasher door is open and with the racks pulled out by default.
-	The team leader can, however, request the dishwasher to be closed and score additional points for opening it. If the robot fails to open the door, it must clearly state it and request the referee to open it.
+	\end{itemize}
 
-	% \item \textbf{Award:} Teams that carry out this task are eligible to win the \textit{Procter \& Gamble Dishwasher Challenge Award}, described in \refsec{award:pandg}.
+	\item \textbf{Award:} Teams that carry out this task are eligible to win the \PGAward{} and a prize sponsored by P\&G. The winner is chosen by representatives of P\&G.
```
semaph0r: We don't know any criteria the P&G representative is using for choosing a winner. Shall we ask them to be more transparent or leave it as is like was in the OpenChallenge?

tasks/EGPSR.tex
```diff
@@ -1,89 +1,76 @@
...
-Execute at least one of the 3 commands requested by the operator.
+Execute 3 commands requested by the operator.
 
-\noindent\textbf{Reward:} 1500pts (500 points per task)\\
+\noindent\textbf{Reward:} 1500pts (500 points per command)
 
-\subsection*{Bonus rewards}
+\subsection*{Bonus Rewards}
 \begin{enumerate}[nosep]
-	\item Attend when called (125pts).
```
semaph0r: This is removed bc it's too simple?

tests/Finals.tex
```diff
@@ -1,85 +1,77 @@
...
 
 \begin{enumerate}
-\item\textbf{League-internal jury:} The league-internal jury is formed by the Executive Committee. The evaluation of the league-internal jury is based on the following criteria:
+\item\textbf{League-Internal Jury:} Formed by the \EC{}. The evaluation is based on the following criteria:
   \begin{compactenum}
-  \item Efficacy/elegance of the solution
-  \item Innovation/contribution to the league of the additional tasks solved
-  \item Difficulty of the overall demonstration
+  \item Efficacy/elegance of the solution.
+  \item Innovation/contribution to the league.
```
semaph0r: Contribution as in "new standard" for upcoming competitions or like scientific contribution as we discussed to raise output for scientific research?

general_rules/AudienceInteraction.tex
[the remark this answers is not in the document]
semaph0r: I think such information for LOC are not in the scope of the rulebook. They should be kept in a separate leaflet for LOC and maybe OC. The arena requirements then could also be compiled into a handout together with the NetSpec.

tasks/StoringGroceries.tex
johaq: Adjusted

tasks/EGPSR.tex
```diff
@@ -1,89 +1,76 @@
...
+	
+	\item \textbf{Instruction Point:} An instruction point is assigned inside the arena where the robot has to go to receive a command.
 
-	\item \textbf{Start location:} The robot starts outside the arena. When the door opens, the robot moves towards the \textit{Standby Point}.
+	\item \textbf{Operators:} A \emph{professional} operator (i.e.~the referee) gives the commands. Optionally, teams can choose to receive commands by a \emph{naive} operator who has no background in robotics, i.e., an audience member. If the robot consistently fails to understand a command, the team can use a \CustomOperator{}
 
-	\item \textbf{Operators:} A \emph{professional operator} (i.e.~the referee) commands the robot by default.
+	\item \textbf{Operators:} A \emph{professional} operator (i.e.~the referee) gives the commands. Optionally, teams can choose to receive commands by a \emph{naive} operator who has no background in robotics, i.e., an audience member. The team can decide to use a \CustomOperator{} outright or after the robot fails to understand the professional or naive operator.
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/HandMeThat.tex
```diff
@@ -1,91 +1,62 @@
...
-	\item \textbf{Groups of Objects:} A group consists of 2--5 random standard objects (see~\refsec{rule:scenario_objects}), separated one from another for about 2.5--10cm.
-	The average distance between the starting position and each group ranges between 50cm and 150cm.
 
+\subsection*{Additional Rules and Remarks}
+\begin{enumerate}[nosep]
+	\item \textbf{Partial Scoring:} The main task allows partial (per object) scoring.
+	
+	\item \textbf{Deus ex Machina:} Score reduction applies per object as follows:
+	\begin{itemize}[nosep]
+		\item \textbf{Asking Yes or No Questions:} Asking yes or no questions about the object (e.g. Is the object yellow?), reduce the score by 50pts.
```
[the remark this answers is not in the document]
johaq: Fixed and re-named

scoresheets/EGPSR.tex
```diff
@@ -2,14 +2,15 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[3]{500}{Perform each task}
-	\penaltyitem[3]{150}{Using custom operator or bypassing ASR}
+	\scoreitem[3]{500}{Complete a command}
+	\penaltyitem[3]{50}{Receiving command not from the instruction point}
```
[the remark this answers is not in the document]
johaq: Fixed

scoresheets/EGPSR.tex
```diff
@@ -2,14 +2,15 @@
...
+	\penaltyitem[3]{100}{Custom Operator}
+	\penaltyitem[3]{150}{Alternative HRI}
 
-	\scoreheading{Bonus rewards}
+	\scoreheading{Bonus Rewards}
 	\scoreitem[3]{150}{Understand command given by naive operator}
-	\scoreitem[3]{100}{Provide audio recording and transcript}
-	\scoreitem{150}{Find the operator when called}
 	\scoreitem{100}{Autonomously leaving the arena}
```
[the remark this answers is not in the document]
johaq: Fixed

scoresheets/Receptionist.tex
```diff
@@ -1,14 +1,13 @@
...
 
-	\scoreheading{Bonus rewards}
+	\scoreheading{Bonus Rewards}
 	\scoreitem[2]{200}{Opening the entrance door to a guest}
```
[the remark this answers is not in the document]
johaq: Fixed

scoresheets/Receptionist.tex
```diff
@@ -1,14 +1,13 @@
...
 	\scoreitem{100}{Describing the first guest to the second guest}
```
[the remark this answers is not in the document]
johaq: Fixed

tasks/Restaurant.tex
johaq: Fixed

scoresheets/Restaurant.tex
johaq: Fixed

johaq: Some yes. Guiding can happen more times. The robot could request guidance to the table where someone wants to order and back to the bar.

tests/Finals.tex
johaq: Imo it's fine like it is

general_rules/Scenario.tex
johaq: I think scenario is fairly established but I'm willing to change it if other TC members agree.

general_rules/Scenario.tex
johaq: It is important because alike objects can occur when known objects are required (Storing Groceries for example). This basically means: Known objects can look differently to how they are pictured in the object list.

general_rules/Scenario.tex
johaq: Changed

general_rules/Scenario.tex
johaq: As I said, imo regular and alike is an important distinction.

general_rules/Scenario.tex
johaq: Changed it to hopefully make it clearer. Unknown basically means anything not in the provided object list.

general_rules/ContinueRules.tex
johaq: Changed

general_rules/AudienceInteraction.tex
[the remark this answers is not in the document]
johaq: To clarify. You think we should remove most of the scenario section from the rules?

tasks/HandMeThat.tex
johaq: Changed name

scoresheets/GPSR.tex
johaq: Changed bonus

Rulebook.tex
johaq: Ordered alphabetically

tasks/CleanTable.tex
johaq: You should ask the EC if they can ask for more detailed criteria. Otherwise, I'd guess that representatives will explain it at the competition if necessary.

tasks/EGPSR.tex
johaq: Mainly, because I don't want to stand in the arena screaming a command for 5 minutes while the robot tries to find me.

tests/Finals.tex
johaq: Both

general_rules/Scenario.tex
YuqianJiang: I think "Arena" and "Environment" are better. All of this subsection seems to be about the arena setup.

tasks/WhatisThat.tex
```diff
@@ -0,0 +1,64 @@
+\section{What is That?}
+\label{test:what-is-that}
+A guest asks the robot to identify objects in the household.
+
+\noindent \textbf{Focus:} \SysI{}, \HRI{}, \GestRec{}, \CV{}, \OR{}.
+
+
+\subsection*{Main Goal}
+The robot names 5 objects, the operator is pointing at.
```
YuqianJiang: Maybe change to "The robot names 5 objects that the guest points at."

tasks/WhatisThat.tex
```diff
@@ -0,0 +1,64 @@
...
+\end{enumerate}
+
+
+\subsection*{Setup and Procedure}
+\begin{enumerate}[nosep]
+	\item \textbf{Location:} This test takes place inside the \Arena{}.
+
+	\item \textbf{Groups of Objects:} There are five groups of 2--5 \KnownObjects{} randomly placed on the floor.
+	
+	\item \textbf{Guest:} The guest is one of the referees. They will step in front of the robot after it enters the \Arena{} and guide them to a group of objects where they point at one.
```
YuqianJiang: This sentence is a bit confusing because of the bad grammar. Does the guest choose a group of objects and guide the robot, or does the robot guide the guest? I guess it should be the former?

tasks/WhatisThat.tex
```diff
@@ -0,0 +1,64 @@
...
+\end{enumerate}
+
+\newpage
+
+\subsection*{Referee Instructions}
+
+The referee needs to
+\begin{itemize}[nosep]
+	\item Place the groups of objects.
+	\item Chose which objects to point at.
```
YuqianJiang: Chose -> Choose

kyordhel:
### Some EC-level restrictions you **CAN'T** change

> Removal of skill certificates. Don't make sense without basic ability tests.

Trustees want us to issue more certificates. If you remove these, you have to come with new ones.

> Removed lean set of rules concept. Self-explanatory

Not for newbies

> Removed powered fridge from scenario. Afaik we never had a powered fridge or used it in a test

Big NO GO. We need one and use it. Is a service for teams to have a place to keep their food, and where we cool down the drinks for the welcome reception (when there is money for it).

> Furniture specs were way too narrow. Contradicts key concept of non-standard scenario

Keep those that set the working area of robots (e.g. HSR grasp distance)

> Removed not trying to solve and just score. This is confusing to new teams and means we designed the test poorly. Also covered in exploiting the rules.

Keep this one. We have had MANY teams faking tests in the past, even some trying to teleop. Just because they weren't publicly shamed doesn't mean this didn't happen. This rule stays.

> Removed restart. No test allows one

Restaurant allows restart

> Removed faking robot rule. Never applied and not enforceable

Also stays

> Rename to P&G test

Stays because P&G is an official sponsor of the league

> Removed open challenge. 

Past year we had Open Challenge and we allotted time for it. Since was an official demo, it has to be in the rules

Neither of these points are open to discussion. Make the rules shorter and more concise if you want, but they stay in the rulebook.

scoresheets/Receptionist.tex
```diff
@@ -1,16 +1,15 @@
...
 
-\begin{scorelist}
+\begin{scorelist}[startbutton=false]
 	\scoreheading{Main Goal}
-	\scoreitem{100}{Introduce the 1st guest to John}
-	\scoreitem{150}{Offer an empty seat to the 1st guest}
-	\scoreitem{100}{Introduce the 2nd guest to John and 1st guest}
-	\scoreitem{150}{Offer an empty seat to the 2nd guest}
-	\penaltyitem[2]{50}{Misunderstanding the name of a guest}
+	\scoreitem[2]{250}{Introduce a new guest to every other guest and offer a seat}
```
YuqianJiang: Is there a penalty for offering an occupied seat, or does that lose all 250pts?

general_rules/Scenario.tex
```diff
@@ -1,202 +1,122 @@
...
-A shelf, rack, or bookcase is required in RoboCup@Home.
-The shelf can be any shelf-like furniture in which objects can be placed.
-\begin{itemize}
-	\item[\textbf{Doors:}] The shelf must have at least one door (preferrably a vertical one) covering up to one half of it.
-	\item[\textbf{Drawers:}] The shelf must have no drawers.
-	\item[\textbf{Shelves:}] The shelf must have 5 shelves or layers between 0.0m and 1.80m from the ground, with a minimum distance of 30cm between shelves or layers.
-\end{itemize}
-
-\subsubsection{Fridge}
-Fridges must not be smaller than 120m. At least one powered and functioning fridge is required.
```
moriarty: I think both @dirkholz and @robotijn will agree that at least one powered and functioning fridge is extremely important to RoboCup@Home and one of the features of the league which helps retain the committee members.

It's required.

moriarty:
> For people who want PDFs of this or any PR, reminder that you can click into the GitHub Actions entry for a given commit and download the built PDF from its artifacts listing.

Thanks @nickswalker that's exactly what I needed.

johaq:
> Trustees want us to issue more certificates. If you remove these, you have to come with new ones.

Well, the current ones CANNOT be awarded. There is no objective way to tell which team made the most points by navigating etc. Therefore they make no sense. I'll put it on the agenda for TC to come up with new ones.

> Big NO GO. We need one and use it. Is a service for teams to have a place to keep their food, and where we cool down the drinks for the welcome reception (when there is money for it).

If this is the purpose of the fridge and it is not necessary for any test then this does not belong in the scenario section. I'd argue it does not belong in the rulebook at all but should be part of our requirements to LOC

> Keep those that set the working area of robots (e.g. HSR grasp distance)

Is there realistically any cupboard, where the HSR cannot reach anything?

> > Removed not trying to solve and just score. This is confusing to new teams and means we designed the test poorly. Also covered in exploiting the rules.

This is imo the worst rule in the entire rulebook because it is so confusing to newer teams. Teleoperating is already covered in two other rules 3.7.3 and 3.7.4 and in both cases it is much clearer what is meant. Any other kind of cheating is also covered in other rules. Please give me an actual example of a case where the other cheating rules don't apply and we need this rule.

> Restaurant allows restart

Should all Stage 2 tests allow restarts or just restaurant? If so, why only restaurant?

> > Removed faking robot rule. Never applied and not enforceable
> 
> Also stays

Pretty much the same applies as above. Confusing as hell and not enforceable. Any sort of cheating is covered in other rules. Provide an example where only this rule applies and I as a referee can actually enforce it.

> > Removed open challenge.
> 
> Past year we had Open Challenge and we allotted time for it. Since was an official demo, it has to be in the rules

As I wrote, I think this should be part of the schedule then. When do we allow time for this? After Stage 2. Evening before Finals?

scoresheets/Receptionist.tex
johaq: Loose all points imo. The robot can just check with the guest if the seat it offered is actually free and look for a different one otherwise.

YuqianJiang: Okay so as long as the person gets seated in the end, there is no penalty; if the person never gets an empty seat, no points. Maybe change "offer a seat" to "get the guest seated" to be clear?

nickswalker:
> If this is the purpose of the fridge and it is not necessary for any test then this does not belong in the scenario section. I'd argue it does not belong in the rulebook at all but should be part of our requirements to LOC

+1. New teams buy furniture based on what the rulebook says, so this section should reflect only things that are necessary for teams to prepare and compete.

general_rules/Scenario.tex
nickswalker: I see, that makes sense.

My suggestion then is change the adjectives to better describe this distinction. "Regular" [has many meanings](https://www.merriam-webster.com/dictionary/regular). The first definition many will think of is the one that is synonymous with "normal": "this object is regular; there's nothing special about it." This is not what we want.

The desired definition is the one that is synonymous with ["consistent"](https://www.merriam-webster.com/dictionary/consistent) ("free from variation"). "Consistent" has fewer interpretations, and the first thing that pops to mind when describing an object as "consistent" is the correct interpretation ("this object is consistent; it looks the same every time").

"Alike" has the correct definition (synonymous with "similar"), but in practice people use it when they want to communicate that the differences are effectively negligible (i.e. a synonym for "same" or "identical": "oh those politicians are all alike"). The simplest alternative is "similar." Others might be "akin" or "resemblant", but people don't usually say these words.

scoresheets/Restaurant.tex
```diff
@@ -1,24 +1,17 @@
...
+	\penaltyitem{200}{Request guidance to a table}
+	\penaltyitem[2]{350}{Deliver a partly wrong order}
+	\penaltyitem[2]{500}{Deliver a fully wrong order}
+	\penaltyitem[2]{100}{Handing over an order instead of placing}
 
-	\scoreheading{Second order (bonus reward)}
-	\scoreitem{500}{Take and serve an additional order}
-	\penaltyitem[2]{200}{Being guided to the \emph{Kitchen-bar} or to a customer's table}
-	\penaltyitem{100}{Not making eye-contact when taking an order}
```
RemiFabre: Just making sure that this was an intentional removal? I think it was a nice add-on. I we think that the test is too hard, this could be turned into a bonus when done instead than a malus when not done?

scoresheets/Restaurant.tex
```diff
@@ -1,24 +1,17 @@
...
-	\penaltyitem[4]{50}{Bypassing object manipulation (handover)}
-	\penaltyitem[5]{100}{Bypassing object manipulation}
-
-	\scoreheading{Additional bonus rewards}
-	\scoreitem[2]{100}{Detect calling or waving customer}
-	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training}
-	\scoreitem[2]{300}{Use an unattached tray to transport an order}
+	\scoreheading{Bonus Rewards}
+	\scoreitem[2]{250}{Use an unattached tray to transport an order}
```
RemiFabre: Could we clarify what an "unattached tray" is? I've seen a team putting a basket inside their robot's gripper, does that qualify as unattached? If so, then no one will ever not use that technique and this bonus will always be scored when an order is succeeded (similar to the bonus "leave the arena" in GPSR). =>If this is the intent, then I would just delete any mention of "trays" and give 750 points per order achieved.
The restaurant is a very hard test (only 1 run, unknown location, laymen all around...) and is the emblematic test of @home in my opinion (it's probably the test that has the biggest audience). Therefore, I believe that the points should be tuned in a way that entice most teams to prepare for it.

DavidPL1: In my opinion it's clear that if a robot starts with the tray or some other container attached to it (even loosely like putting a basket in the opened gripper before starting the test) it can be considered as an attachment to the robot, as it does not require any autonomous interaction by the robot. But I think it could be helpful to define some basic rules on how to achieve the bonus, for instance, would the bonus apply if the barman hands over the tray on request or does the robot need to pick up the tray by its own and if so, where will the tray be placed.

tasks/EGPSR.tex
```diff
@@ -1,89 +1,76 @@
...
+	
+	\item \textbf{Minimal Help:} To score this bonus, the robot needs to complete the commands while requesting relatively few \abb{DEMs} compared to other teams. This is up to the consideration of the \abb{TC}.
```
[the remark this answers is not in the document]
RemiFabre: I also like the idea but it seems similar to just removing more points for each DeM and I suspect it will be quite annoying for the TC to handle on the day. If you want to go through with this, I suggest deciding how to handle it before hand, e.g:
"100 points for the team that used the least amount of DeM (several teams will score the bonus if the number of DeM is tied).

kyordhel:
Yet again

# Some EC-level restrictions you WILL NOT change

1. Removal of skill certificates. 
2. Removed lean set of rules concept. 
3. Removed powered fridge from scenario.
4. Furniture specs were way too narrow.
5. Removed not trying to solve and just score.
6. Removed restart. 
7. Removed faking robot rule.
8. Rename to P&G test
9. Removed open challenge.

@johaq I appreciate your hard work and commitment in shrinking the Rulebook, but I'm not discussing any of this points. I'm telling plain and clear that neither of these changes will be done in this pull request, so please revert them ASAP. We are being permissive enough with a PR that changes over 40 pages.

I'm not asking for explanations or opening this for discussion. Is an explicit request. If you want to change any of these 9 points, open a pull request for the topic and it will be discussed there, maybe changed, but definitively not here.

general_rules/Scenario.tex
johaq: These make a lot of sense to me. Also not arguing with a native speaker on this :)
Changed to Consistent and Similar Objects

tasks/WhatisThat.tex
johaq: Guest is definitely better. Question for native speaker:
The robot names 5 objects, the guest is pointing at. 
or
The robot names 5 objects that the guest is pointing at. 
?

tasks/WhatisThat.tex
johaq: Hopefully, made it clearer

tasks/WhatisThat.tex
johaq: fixed

scoresheets/Receptionist.tex
johaq: Yeah, this is somewhat difficult to decide for each test. I think here, I'm fine with it because the person is naturally there and it is reasonable for the robot to ask whether the person wants to sit there. The reward for a team that gets it right on the first try is that they have more time.

scoresheets/Restaurant.tex
johaq: Yes intentional. Eye contact is somewhat difficult to determine for some platforms.

scoresheets/Restaurant.tex
johaq: What David said. Robot starts with the try in hand or wherever -> tray is considered attached. Robot picks up the tray during the test -> tray is not considered attached.

johaq:
I hopefully addressed every point of @kyordhel  except for the Open Challenge which I will do when scheduling is clear.

scoresheets/Restaurant.tex
RemiFabre: Is the barman allowed to handover the tray of the robot asks?

johaq: Yes, but I think in order for the robot to score the bonus points it needs to grab the tray itself.

johaq:
Added open challenge

johaq:
I still need review approvals to merge as decided in last meeting

PR #754, opened 2022-06-20
2022

kyordhel:
From [alex-mitrevski/2022/draft](https://github.com/RoboCupAtHome/RuleBook/commit/abce6cfd8b56291965def166c6f2d9763b18cc7c). this PR is to check conflicts with master for a safe merge.


justinhart:
Looks good.

justinhart (approved):

MatthijsBurgh:
@justinhart please resolve all the conflicts before approving....

alex-mitrevski:
@kyordhel As there are quite a few conflicts since `master` had all the changes for 2020, while `2022` is derived from the 2019 rulebook, my suggestion would be not to try resolving them, also because all the changes on `master` would be lost that way.

I think the easier thing to do would be to push the current `master` to a different branch (e.g. `2020`) so that those changes are not lost; then we can revert `master` to the last 2019 commit (by "we" I mean someone who can force push to `master`) so that `2022` can be directly merged. Not very elegant, but probably the least painful given the circumstances.

Another alternative would be to change the default branch (to one that starts from the 2019 rulebook); then this PR can be directed to that one instead.

What do you think?

MatthijsBurgh:
I agree with @alex-mitrevski

MatthijsBurgh:
@kyordhel Do you agree with me and @alex-mitrevski?

justinhart:
I agree on this. I think that @kyordhel has the access to update the
website.

MatthijsBurgh:
I have created the 2020_master branch, which points to the current master.

The master is protected, so I can't force push to the master branch to reset it to the 2022 branch. Somebody with the correct permissions should do that.

Maybe the REAMDE does require a small update. I suggest to do that after resetting the master.

kyordhel:
@MatthijsBurgh I agree with what @alex-mitrevski. Whatever it takes to have the rulebook released.

I want to note that I do NOT have the permissions to merge. Nobody does. Git simply does not allow us to merge conflicting branches.

I'll try to fix as much as I can.

Issue #767, opened 2022-07-24
Reduce Number of Tests, Make All Mandatory


johaq:
I think back in 2018 we decided we wanted to make the competition less stressful for teams.
That initially meant reducing the number of tests and therefore the number of test slots.
This idea evolved into just reducing the number of test slots but letting teams choose which tests to perform.
This meant instead of reducing the number of tests we actually drastically increased the number of tests so
teams had a choice. 

**I think reversing this descision is the most important change we need to make for next year!**

Letting teams choose the test introduced several problems to the competition:
* Teams' scores are not at all comparable anymore
    * This goes for individual test scores, i.e. teams got 1000 pts for taking pictures of people and placing them on a map, but another got 650 points for picking and placing 4 objects and pouring milk and cereal into a bowl. As well as for the overall scores.
* The league is even more diffictult to follow for the audience
     * Why does one robot store objects into a cabin and the next robot talks to some people?
* Running a testslot becomes a nightmare for referees
    * Everytime a team wants to do storing groceries I need to redecorate the arena? Not to mention that switching constantly between tests just drastically increases the chance of the referee making a mistake. We already have a problem of not having enough people to referee but now we basically created a scenario where a first year referee has to run an incredibly difficult test slot, whereas in the past they could start out by running a Person Recognition test slot for example.

I propose switching back to the old system where every test had its own test slot. To keep in mind the earlier idea of reducing the amount of stress for teams, I propose to reduce the number of tests to:
Stage 1:
* Storing Groceries
* Carry my Luggage
* Receptionist
* Set the Table or Take out the Garbage
* GPSR

Stage 2:
* Restaurant
* Clean the Table
* Smoothie Chef
* EGPSR

I think this gives the easiest to understand tests for the audience while still covering a wide variety of robot skills.
This also gives us back the ability to award points depending on how difficult something is and not based on trying to reach the set amount of points of a test.

moriarty:
I agree.

MatthijsBurgh:
I agree with the idea. I am not sure yet about which challenges.

This of course starts with the focus of the League. Then determine skills to test in the challenges. Then select the challenges from current set or create new ones. But I let this to @PetervDooren.

PetervDooren:
Another aspect that is improved with this is that teams will not pass on difficult but interesting challenges such as the GPSR.

MatthijsBurgh:
Try to find a good balance between keeping the league accessible for new teams and interesting and challenge for established teams

[...]

PR #771, opened 2022-07-29
Remove Scenarios. Cut Number of Tests

johaq:
This is currently being discussed in #767 
Please keep in mind that the test selection is by no means meant to be final. This PR is meant to give us a base to work from moving forward.

### DIFF
--- a/general_rules/Organization.tex
+++ b/general_rules/Organization.tex
@@ -23,31 +23,17 @@ The competition features a \iterm{stage system}, namely it is organized in two s
 Note that the same task can be performed multiple times during \SONE{} and \STWO{} (see~\refsec{rule:score_system}).
 In case of having no considerable score deviation between a team advancing to the next stage and a team dropping out, the TC may announce additional teams advancing to the next stage.
 
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\subsection{Thematic Scenarios}
-\label{rule:themes}
-
-Each stage includes a set of tasks that are grouped in two thematic scenarios:
-\begin{enumerate}
-	\item The \iterm{Housekeeper} scenario features tasks related to cleaning, organization, and maintenance of a household.
-	\item The \iterm{Party Host} scenario focuses on providing general assistance during a party by attending to the needs of the guests.
-\end{enumerate}
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Schedule}
 \label{rule:schedule}
 
 \begin{enumerate}
-	\item \textbf{Thematic scenario blocks:} Two \Testblocks{} are scheduled per day, lasting between two and three hours.
-	All teams have at least 2 \Testslots{} assigned per \Testblock{}, during which they can perform any task of their choice from the block's assigned scenario.
+	\item \textbf{Test Blocks:} For each test one Test Block is scheduled during which each (qualified) team gets at least two Test Slots assigned. With the Restaurant test being the exception where only one Test Slot is assigned.
 
-	\item \textbf{Test slots:} In principle, all teams get the same amount of \Testslots{} with a minimum of 2 per block.
-	In case there is sufficient unoccupied time, a referee may open an extra testing slot, thereby allowing an extra run for all teams.
+	\item \textbf{Test Slots:} During a Test Slot a team has the amount of time specified for the test to attempt to solve it. 
 
-	\item \textbf{Tests:} Teams must inform the OC in advance about the tasks that they want to perform in each block.
-	Note that only one task can be attempted per \Testslot, but different tasks may be attempted in different \Testslots{} within a \Testblock{} (i.e. one task can be attempted in the first slot, while another one in the following slot, if desired).
-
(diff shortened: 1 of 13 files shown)


general_rules/Organization.tex
```diff
@@ -23,31 +23,17 @@ \subsection{Stage System}\label{rule:stages}
...
 
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-\subsection{Thematic Scenarios}
-\label{rule:themes}
-
-Each stage includes a set of tasks that are grouped in two thematic scenarios:
-\begin{enumerate}
-	\item The \iterm{Housekeeper} scenario features tasks related to cleaning, organization, and maintenance of a household.
-	\item The \iterm{Party Host} scenario focuses on providing general assistance during a party by attending to the needs of the guests.
-\end{enumerate}
```
PetervDooren, line 34: I think it would be good to keep the thematic scenarios as these are two different contexts in which the robot is employed. Otherwise it might become unclear what the goal of the robot is.

general_rules/Organization.tex
```diff
@@ -23,31 +23,17 @@ \subsection{Stage System}\label{rule:stages}
...
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection{Schedule}
 \label{rule:schedule}
 
 \begin{enumerate}
-	\item \textbf{Thematic scenario blocks:} Two \Testblocks{} are scheduled per day, lasting between two and three hours.
-	All teams have at least 2 \Testslots{} assigned per \Testblock{}, during which they can perform any task of their choice from the block's assigned scenario.
+	\item \textbf{Test Blocks:} For each test one Test Block is scheduled during which each (qualified) team gets at least two Test Slots assigned. With the Restaurant test being the exception where only one Test Slot is assigned.
```
PetervDooren, line 32: #TODO add what was discussed last TC meeting. The GPSR slot after stage1 can be used to re-take other tests.

general_rules/Organization.tex
johaq: I think leaving in scenarios when they have no impact on scheduling or rules is confusing for new readers

general_rules/Organization.tex
johaq: Added

kyordhel:
Disagree. Not approving this PR. Reasons below.

- Do NOT apply this patch. It was hell of a work adapting the rulebook from tests to scenarios/slots. If you're going back to plain tests just branch from 2018 or 2019 and replace the tests. The result will be much cleaner. Otherwise the Rulebook will be quite confusing and with inconsistent terminology and rules.
- We have three leagues. Although Trustees want all 3 leagues running the same test the dire truth is SSPL is dying and Peppers can't grasp. It makes no sense to have Set the Table and Storing Groceries for SSPL.
  - Add more Stage I tests and restrict them for SSPL, the others will be OPL/DSPL.
  - Add an SSPL alternative for clean the table.
  - SSPL focus is heavy HRI since the robot can barely do anything else.
- Make sure scoresheets reflect you have 1 go with restart or 3 runs rather than multiple tries with flexible test-slots.

**Final note:** Once @LoyVanBeek tried to make scoring consistent among tests and it became a nightmare. Already tried, didn't work. Here I want to advice TC to read the existing papers made by former Trustees and Execs. Focus on scoring subgoals so teams aim to complete the task, and not on point farming.

kyordhel (changes requested):
Check comment above

johaq:
Please point out any existing inconsistencies. 
I believe we have not discussed yet whether we want restarts or set amount of tries. This PR is not intended to be final but smth to work of of. 
I have no problem with keeping all the previous Party Host tests for the SSPL.

ARTenshi:
To contribute to the discussion, these are the tasks where SSPL teams scored (an X per team that scored) before leaving all of the "Party Host" tasks for that league:

Housekeeper

Stage I:
CleanUp
GPSR
Storing Groceries
Serve Breakfast
Take Out the Garbage

Stage II:
Clean the table
EGPSR
Find my Disk
Set the table


Party host:

Stage I:
Carry my luggage
Farewell
Find my mates           X
Receptionist            XXX
Serving Drinks

Stage II:
Hand me that
Restaurant
Smoothie chef
Sticker for the rules   XX
Where is this

mleonetti (changes requested):
I  am fine with most of the changes, but would leave the "scenarios" as containers for distinguishing HRI tasks from robot vs environment tasks. I don't remember discussion the removal of scenarios, but that may be me.

mleonetti:
guys, we need to progress this discussion faster. I know I proposed this before to no avail, but a EC/TC meeting would let us figure this out in 1 hour instead of several weeks.

LoyVanBeek:
https://doodle.com/en/ :-)

mleonetti:
> https://doodle.com/en/ :-)

how did we not think of that!

tomkimsour:
Has this PR been discussed since november ?

johaq:
Tasks now in line with latest decision

PetervDooren (approved):

johaq:
> I am fine with most of the changes, but would leave the "scenarios" as containers for distinguishing HRI tasks from robot vs environment tasks. I don't remember discussing the removal of scenarios, but that may be me.

I think keeping scenarios is confusing if they do not have any real bearing on rules and scheduling

PR #777, opened 2023-02-01
Schedule and Scoring Clarification

johaq:
-Clarified one attempt per test slot.
-Removed the 1000/2000 pt rule for stage 1/2 tests. Since points are clarified for each test anyway and gpsr tests are exceptions. Imo was just confusing and did not add valuable information.

Per last TC meeting:
-Updated rule for gpsr slot. Teams can attempt test but not score.
-Remove rule about about gpsr requirement for stage 2. Meaning of participation is vague and GPSR is encouraged now anyway with overall schedule change.  And do we really cancel stage 2 if no team participates in GPSR?
-Teams score for their best attempt. This could change depending on upcoming TC discussions. How much do we value consistency?

### DIFF
--- a/general_rules/Organization.tex
+++ b/general_rules/Organization.tex
@@ -8,14 +8,14 @@ The competition features a \iterm{stage system}, namely it is organized in two s
 \begin{enumerate}
 	\item \textbf{Robot Inspection:} For security, robots are inspected during the \SetupDays.
 	A robot must pass the \RobotInspection{} test (see~\refsec{sec:robot_inspection}) so that it is allowed to compete.
-
+	
 	\item \textbf{\SONE:} The first days of the competition are called \SONE.
 	All qualified teams that have passed the \RobotInspection{} can participate in \SONE.
-
-
+	
+	
 	\item \textbf{\STWO:} The best \emph{50\% of teams} after \SONE{} advance to \STWO. If the total number of teams is less than 12, up to 6 teams may advance to \STWO.
 	In this stage, tasks require more complex abilities or combinations of abilities.
-
+	
 	\item \textbf{\FINAL:} The best \emph{two teams} of each league, namely the ones with the highest score after \STWO, advance to the \FINAL.
 	The final round features only a single task integrating all tested abilities.
 	In order to participate in the \FINAL, a team must have solved at least one of the \STWO{} tasks.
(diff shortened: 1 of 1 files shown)


general_rules/Organization.tex
```diff
@@ -29,13 +29,13 @@ \subsection{Schedule}
 \label{rule:schedule}
 
 \begin{enumerate}
-	\item \textbf{Test Blocks:} For each test one Test Block is scheduled during which each (qualified) team gets at least two Test Slots assigned. With the Restaurant test being the exception where only one Test Slot is assigned.
-
-	\item \textbf{Test Slots:} During a Test Slot a team has the amount of time specified for the test to attempt to solve it. 
-
+	\item \textbf{Test Blocks:} For each test one Test Block is scheduled during which each (qualified) team gets at least two, maximum five Test Slots assigned. With the Restaurant test being the exception where only one Test Slot is assigned.
+	
+	\item \textbf{Test Slots:} During a Test Slot a team has one attempt to solve the test in the test's specified amount of time
```
PetervDooren, line 34: There is a rule that allows teams to request a restart if not more than 50% of the time has passed yet. This conflicts with that but I dont think that rule is ever used, so I would opt to remove that instead.

general_rules/Organization.tex
```diff
@@ -8,14 +8,14 @@ \subsection{Stage System}\label{rule:stages}
 \begin{enumerate}
 	\item \textbf{Robot Inspection:} For security, robots are inspected during the \SetupDays.
 	A robot must pass the \RobotInspection{} test (see~\refsec{sec:robot_inspection}) so that it is allowed to compete.
-
+
```
PetervDooren, line 11: please remove these tabs

general_rules/Organization.tex
```diff
@@ -128,13 +128,10 @@ \subsection{Scoring System}
 \label{rule:score_system}
 
 Each task has a main objective and a set of bonus scores.
-To score in a test, a team must successfully accomplish the main objective of the task; bonuses are not awarded otherwise.
+To score in a test, a team must successfully accomplish a main objective of the task; bonuses are not awarded otherwise.
```
PetervDooren, line 131: So what does this mean exactly? a team must score more than 0 points on the main objective? Then we should formulate is as such.

PetervDooren (changes requested):

PetervDooren:
@johaq Are you still developing this?

general_rules/Organization.tex
PetervDooren: This rule has been removed as of #778

PR #778, opened 2023-02-13
Remove unused rules

PetervDooren:

## Description
Removed rules which are not enforced in practice to clean up the rulebook and better reflect the situation at an actual tournament

Closes issue #

- Removed restart rule: this rule is seldom used in practice.
- Removed the rule on moderation. LOC rarely arranges for this so don't need to add this rule. If we do want to provide moderation the LOC can specify it on their platform. 
- Updated the allocation of referees: In practice the main referee is a member of the TC and one or more team members are asked to assist. But the protocols in the rulebook are not used.
- clarify that the referee will press the emergency button, rather than ask the team to stop the robot.
- Allow multiple teammembers in the arena for the purpose of removing the robot: this is already done, despite the fact that it is technically not allowed.
- remove rule which states that a team must complete at least one stage II task to compete in the finals. This will never happen, and if it does something has gone terribly wrong in tuning the difficulty of challenges.
- fix minor \iterm \iaterm bug

## Other comments
If this request is marked as a draft, please describe your plans or any specific feedback you would like

### DIFF
--- a/general_rules/Procedure.tex
+++ b/general_rules/Procedure.tex
@@ -75,12 +75,6 @@ Hence, interaction should be as with any other human being.
 	This is known as \iterm{functional touching}.
 	However, the robot must clearly announce the collision-like interaction and kindly request not being stopped.\\
 	\textbf{Remark: } Referees can (and will) immediately stop a robot in case or suspicion of \emph{dangerous} behavior.
-
-	\item \textbf{Robot-Robot avoidance:} If two robots encounter each other, they both have to actively try to avoid the other robot.
-	\begin{enumerate}
-		\item A robot which is not going for a different route within a reasonable amount of time (e.g., \SI{30}{\second}) is removed.
-		\item A non-moving robot blocking the path of another robot for longer than a reasonable amount of time (e.g., \SI{30}{\second}) is removed.
-	\end{enumerate}
 \end{enumerate}
 
 
--- a/general_rules/Procedure.tex
+++ b/general_rules/Procedure.tex
@@ -4,8 +4,8 @@
 \subsection{Safety First!}
 \label{rule:safetyfirst}
 \begin{enumerate}
-	\item \textbf{Emergency Stop:} At any time when operating the robot inside and outside the scenario the owners have to stop the robot immediately if there is a remote possibility of dangerous behavior towards people and/or objects.
-	\item \textbf{Stopping on request:} If a referee, member of the Technical or Organizational committee, an Executive or Trustee of the federation tells the team to stop the robot, there will be no discussion and the robot must be stopped \emph{immediately}.
+	\item \textbf{Emergency Stop:} At any time when operating the robot inside and outside the scenario the owners have to stop the robot immediately if there is a possibility of dangerous behavior towards people and/or objects.
+	\item \textbf{Stopping on request:} If a referee, member of the Technical or Organizational committee, an Executive or Trustee of the federation stops the robot (by pressing the emergency button) there will be no discussion. Similarly if they tell the team to stop the robot, the robot must be stopped \emph{immediately}.
 	\item \textbf{Penalties:} If the team does not comply, the team and its members will be excluded from the ongoing competition immediately by a decision of the RoboCup@Home \iaterm{Technical Committee}{TC}. 	Furthermore, the team and its members may be banned from future competitions for a period not less than a year by a decision of the RoboCup Federation Trustee Board.
 \end{enumerate}
 
--- a/general_rules/Procedure.tex
+++ b/general_rules/Procedure.tex
@@ -72,7 +72,7 @@ Hence, interaction should be as with any other human being.
 	\item \textbf{\iterm{Major collisions}:} If a robot crushes into something during a test, the robot is immediately stopped.	Additional penalties may apply.
 
 	\item \textbf{\iterm{Functional touching}:} Robots are allowed to apply pressure on objects, push away furniture and, in general, interact with the environment using structural parts other than their manipulators.
-	This is known as \iaterm{functional touching}.
+	This is known as \iterm{functional touching}.
 	However, the robot must clearly announce the collision-like interaction and kindly request not being stopped.\\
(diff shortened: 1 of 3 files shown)


general_rules/Procedure.tex
```diff
@@ -72,15 +72,9 @@ \subsection{Collisions}
 	\item \textbf{\iterm{Major collisions}:} If a robot crushes into something during a test, the robot is immediately stopped.	Additional penalties may apply.
 
 	\item \textbf{\iterm{Functional touching}:} Robots are allowed to apply pressure on objects, push away furniture and, in general, interact with the environment using structural parts other than their manipulators.
-	This is known as \iaterm{functional touching}.
```
MatthijsBurgh, line 75: Why is this change needed? `iaterm` is a valid command, see https://github.com/RoboCupAtHome/RuleBook/blob/5185332e20fa3fd4444a9930e4350ed2a1033290/setup/abbrevix.tex#L51

general_rules/Procedure.tex
```diff
@@ -72,15 +72,9 @@ \subsection{Collisions}
...
+	This is known as \iterm{functional touching}.
 	However, the robot must clearly announce the collision-like interaction and kindly request not being stopped.\\
 	\textbf{Remark: } Referees can (and will) immediately stop a robot in case or suspicion of \emph{dangerous} behavior.
-
-	\item \textbf{Robot-Robot avoidance:} If two robots encounter each other, they both have to actively try to avoid the other robot.
-	\begin{enumerate}
-		\item A robot which is not going for a different route within a reasonable amount of time (e.g., \SI{30}{\second}) is removed.
-		\item A non-moving robot blocking the path of another robot for longer than a reasonable amount of time (e.g., \SI{30}{\second}) is removed.
-	\end{enumerate}
```
MatthijsBurgh, line 83: I think this valid as the number of multi robot situations have been increasing

general_rules/Procedure.tex
```diff
@@ -228,34 +196,6 @@ \subsection{Time limits}
 	\item \textbf{Show must go on:} On special cases, the referee may let the robot continue the test for demonstration purposes, but no additional points will be scored.
 \end{enumerate}
 
-
-
-\subsection{Restart}
```
MatthijsBurgh, line 233: Restarts have been used in Stage 2 tests in the pasts. As these tests have a longer duration than stage 1.

general_rules/Procedure.tex
PetervDooren: Because with iaterm the result in pdf is "functional touching (.)"
As far as I can tell iaterm takes two arguments.

general_rules/Procedure.tex
```diff
@@ -187,23 +172,6 @@ \subsection{Operators}
 Penalty may apply when using a custom operator.
 
 
-\subsection{Moderator}
```
MatthijsBurgh, line 190: So do we eliminate the usage of any other language than English?

general_rules/Scenario.tex
```diff
@@ -176,7 +176,6 @@ \subsection{Predefined Rooms and Locations}
 \label{rule:scenario_locations}
 
 Some tests in the RoboCup@Home league involve a \PredefinedLocation{} where people or objects can be found.
-Each \PredefinedLocation{} has an assigned \LocationClass{} (for instance, a \textit{Couch} and an \textit{Arm Chair} belong to the \textit{Seating} class).
```
MatthijsBurgh, line 179: Any reason to remove this definition?

PetervDooren: It has not been provided in past tournaments. There is also no mention of it in the rest of the rulebook. So this is a definition without use

MatthijsBurgh: Fair

general_rules/Procedure.tex
PetervDooren: Where? In none of the tests do we use multiple robots at once. The only times this may happen is during the setup days (in which case the robots do not follow these rules) or in the final (where the robots belong to the same team)

general_rules/Procedure.tex
PetervDooren: No, that is not what this rule says. This rule is meant to have someone of the team be available to provide live commentary to the audience. Which is not/rarely organised by the LOC.

general_rules/Procedure.tex
MatthijsBurgh: I have seen situations in which two robots run the restaurant challenge at the same time. And multiple robots can be used in Open/Final challenges.

general_rules/Procedure.tex
MatthijsBurgh: Oh sorry didn't read carefully enough

general_rules/Procedure.tex
PetervDooren: In Open and Final challenges the robots belong to the same team. We do not need to specify how the robots should interact in that case since the team will opt for a working solution with two of their own robots.
As for the restaurant challenge, I have not seen it in 2019 or 2022. Don't know much about years before that. But it seems strange for that to be the only reason to include these rules. The restaurant also does not state that multiple robots may test simultaneously.

general_rules/Procedure.tex
PetervDooren: I don't have enough experience to say whether or not this rule helps for that.
Another question I had was whether or not a team may use this rule when their robot fails at the door. The rule does not specify how much time a team has to fix their robot.
I think this will make running a test slot more of a hastle while there are also multiple attempts per test slot.

MatthijsBurgh: I leave this one for you to discuss with other TC members.

general_rules/Procedure.tex
MatthijsBurgh: I leave this one for you to discuss with other TC members.

ARTenshi (approved):
Changes have been discussed at the TC's slack channel and no extra feedback was provided.

PR #783, opened 2023-03-29
Score update clean the table

PetervDooren:
**Note: created this PR based on suggestions by @ARTenshi Luis Contreras**

## Description

Clean the table as described now seems too easy for a stage 2 task. Basically, the robot can request to be handed all objects and place them in a random deposit in the dishwasher and still score high; furthermore, by placing all objects in the same hardcoded rack, by definition, some of them will get the bonus points for the correct location. Finally, this applies to the dishwasher tab if it can be handed, as well. This is more a "Put the objects in the robot hand and deposit them in any rack" task. This is not suitable for a Stage II test.

- Make picking up the dishwasher tab mandatory for the bonus points
- remove partial scoring for picking up an object
- award bonus points for correct placement only if all are correctly placed, to avoid it happening by coincidence
- increase deus ex machina penalty

### DIFF
--- a/scoresheets/CleanTable.tex
+++ b/scoresheets/CleanTable.tex
@@ -2,18 +2,17 @@ The maximum time for this test is 10 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[5]{60}{Picking tableware and cutlery for transportation to the dishwasher}
-	\scoreitem[5]{180}{Placing the tableware and cutlery inside the dishwasher}
+	\scoreitem[5]{240}{Placing the tableware and cutlery inside the dishwasher}
 
 	\scoreheading{Bonus Rewards}
 	\scoreitem{200}{Opening the dishwasher door}
 	\scoreitem{100}{Pulling out the dishwasher rack}
-	\scoreitem[5]{40}{Placing an item correctly in the dishwasher}
-	\scoreitem{300}{Placing the dishwasher tab inside the dishwasher}
+	\scoreitem{200}{Placing all items correctly in the dishwasher}
+	\scoreitem{300}{Taking and placing the dishwasher tab inside the dishwasher}
 
 	\scoreheading{Deus Ex Machina Penalties}
-	\penaltyitem[5]{25}{Pointing at an object or telling the robot where an object is}
-	\penaltyitem[5]{60}{Handing an object over to the robot}
+	\penaltyitem[5]{100}{Pointing at an object or telling the robot where an object is}
+	\penaltyitem[5]{140}{Handing an object over to the robot}
 
 	%\setTotalScore{1000}
 \end{scorelist}


scoresheets/CleanTable.tex
```diff
@@ -2,18 +2,17 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[5]{60}{Picking tableware and cutlery for transportation to the dishwasher}
```
PetervDooren, line 5: To encourage picking up the objects I think we should keep the partial scoring, we may spread the focus more evenly between pickup and place. Removing the partial scoring only discourages going for pickup as you wont get points anymore if the placing fails afterwards.

scoresheets/CleanTable.tex
```diff
@@ -2,18 +2,17 @@
...
-	\scoreitem[5]{180}{Placing the tableware and cutlery inside the dishwasher}
+	\scoreitem[5]{240}{Placing the tableware and cutlery inside the dishwasher}
 
 	\scoreheading{Bonus Rewards}
 	\scoreitem{200}{Opening the dishwasher door}
 	\scoreitem{100}{Pulling out the dishwasher rack}
-	\scoreitem[5]{40}{Placing an item correctly in the dishwasher}
-	\scoreitem{300}{Placing the dishwasher tab inside the dishwasher}
+	\scoreitem{200}{Placing all items correctly in the dishwasher}
```
PetervDooren, line 10: While some objects may be placed correctly by accident I think this makes it too easy for things to accidentaly go wrong. If one objects placement goes wrong you lose out on 200 points. I would try to avoid these high risk high reward situations. This may discourage teams from even trying to go for a correct placement.

scoresheets/CleanTable.tex
```diff
@@ -2,18 +2,17 @@
...
+	\scoreitem{300}{Taking and placing the dishwasher tab inside the dishwasher}
```
PetervDooren, line 11: I think this is fair for bonus points.

I do wonder if the dishwasher tab can be placed anywhere in the dishwasher or if it has a small sepate compartment it must be placed in. That depends on the type of dishwasher and is not specified. For this many bonus points I would be fine with the latter.

scoresheets/CleanTable.tex
```diff
@@ -2,18 +2,17 @@
...
 
 	\scoreheading{Deus Ex Machina Penalties}
-	\penaltyitem[5]{25}{Pointing at an object or telling the robot where an object is}
-	\penaltyitem[5]{60}{Handing an object over to the robot}
+	\penaltyitem[5]{100}{Pointing at an object or telling the robot where an object is}
+	\penaltyitem[5]{140}{Handing an object over to the robot}
```
PetervDooren, line 15: I think these are the biggest ways we can improve the challenge. Increasing the penalty for handover. 

I am assuming the penalty for handing over an object is cumulative with not receiving points for pickup.

PetervDooren (commented):
I agree that it is too easy to score points on this challenge with a trivial solution "handover and place anywhere" I think some of these changes do not support fixing this problem

scoresheets/CleanTable.tex
ARTenshi: Pick-and-place is now a basic skill (or should be), is more a problem than a solution, i.e. to decide when the limits of a "pick" and a "place" are is ambiguous and hard to evaluate objectively.

scoresheets/CleanTable.tex
ARTenshi: We should remember this is a bonus, as we don't give partial points for "trying" to open the dishwasher's door or the rack, we should not give partial points for "trying" to classify objects (or, in this case, give free points even if a team is not trying to do the classification itself as it wouldn't make any sense to give points to a team if the dishwasher's door opens by accident).

scoresheets/CleanTable.tex
ARTenshi: True. Adding "per delivered object" should work.

PetervDooren:
Similar to #784, requires further discussion within the TC

PR #784, opened 2023-03-29
Alt update scoring clean the table

PetervDooren:
## Description

Alternative solution to PR #783 

- make placing correctly main challenge
Redistribute scoring to:
- increase importance of pickup
- increase importance of placing correctly

### DIFF
--- a/scoresheets/CleanTable.tex
+++ b/scoresheets/CleanTable.tex
@@ -13,7 +13,7 @@ The maximum time for this test is 10 minutes.
 
 	\scoreheading{Deus Ex Machina Penalties}
 	\penaltyitem[5]{25}{Pointing at an object or telling the robot where an object is}
-	\penaltyitem[5]{-}{Handing an object over to the robot (no points are awarded for pickup)}
+	\penaltyitem[5]{0}{Handing an object over to the robot (no points are awarded for pickup)}
 
 	%\setTotalScore{1000}
 \end{scorelist}
--- a/scoresheets/CleanTable.tex
+++ b/scoresheets/CleanTable.tex
@@ -2,18 +2,18 @@ The maximum time for this test is 10 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[5]{60}{Picking tableware and cutlery for transportation to the dishwasher}
-	\scoreitem[5]{180}{Placing the tableware and cutlery inside the dishwasher}
+	\scoreitem[5]{120}{Picking tableware and cutlery for transportation to the dishwasher}
+	\scoreitem[5]{80}{Placing the tableware and cutlery inside the dishwasher}
+	\scoreitem[5]{80}{Placing an item in the correct place in the dishwasher}
 
 	\scoreheading{Bonus Rewards}
 	\scoreitem{200}{Opening the dishwasher door}
 	\scoreitem{100}{Pulling out the dishwasher rack}
-	\scoreitem[5]{40}{Placing an item correctly in the dishwasher}
-	\scoreitem{300}{Placing the dishwasher tab inside the dishwasher}
+	\scoreitem{300}{Picking and Placing the dishwasher tab inside the dishwasher}
 
 	\scoreheading{Deus Ex Machina Penalties}
 	\penaltyitem[5]{25}{Pointing at an object or telling the robot where an object is}
-	\penaltyitem[5]{60}{Handing an object over to the robot}
+	\penaltyitem[5]{-}{Handing an object over to the robot (no points are awarded for pickup)}
 
 	%\setTotalScore{1000}
 \end{scorelist}


scoresheets/CleanTable.tex
```diff
@@ -2,18 +2,18 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[5]{60}{Picking tableware and cutlery for transportation to the dishwasher}
-	\scoreitem[5]{180}{Placing the tableware and cutlery inside the dishwasher}
+	\scoreitem[5]{120}{Picking tableware and cutlery for transportation to the dishwasher}
```
ARTenshi, line 5: According to my experience in Tidy Up, it is hard to evaluate the taking part only without the correct delivery, as a robot can drop the object and take a different one or keep taking the same and keep scoring without performing the task; that is why it is preferred to grant points if the take and delivery is performed and leave the partial points to the penalties section.

scoresheets/CleanTable.tex
```diff
@@ -2,18 +2,18 @@
...
 	\scoreitem{200}{Opening the dishwasher door}
 	\scoreitem{100}{Pulling out the dishwasher rack}
-	\scoreitem[5]{40}{Placing an item correctly in the dishwasher}
-	\scoreitem{300}{Placing the dishwasher tab inside the dishwasher}
+	\scoreitem{300}{Picking and Placing the dishwasher tab inside the dishwasher}
 
 	\scoreheading{Deus Ex Machina Penalties}
 	\penaltyitem[5]{25}{Pointing at an object or telling the robot where an object is}
-	\penaltyitem[5]{60}{Handing an object over to the robot}
+	\penaltyitem[5]{0}{Handing an object over to the robot (no points are awarded for pickup)}
```
ARTenshi, line 16: It goes in accordance to the comment  above. Is is hard to score partial deliveries. The reduction should be applied only if the take and delivery is performed.

scoresheets/CleanTable.tex
PetervDooren: I dont know what exactly the guidelines are on when an object is considered 'picked up'. the intention here is that the robot has control over the object and can start moving to the goal. So just lifting and then dropping would not count as picking up. But I get that there is ambiguity there. The main reason to do partial scoring this way for me is that it rewards teams even if accidents happen later on. Otherwise the robot would have to accomplish the full delivery to get any points.

ARTenshi: >The main reason to do partial scoring this way for me is that it rewards teams even if accidents happen later on.

The idea of the complete delivery is to make robots aware of these accidents and act in consequence (e.g. by taking the object again if the robot drops it in any part of the delivery process – table or floor).

Therefore, multiple grasping of the same object might occur and having the score open (although "it is not the intention of the challenge") will make it hard to evaluate.

PetervDooren: Having robots be aware of these accidents is good. But this test is not designed to assess that capability. We cannot guarantee that every robot will have such an accident. So whether or not a complete delivery takes place will depend on luck mainly. Most of the time a successful pickup will mean a succesfull delivery. In the off chance that a team has an accident during transport I would grant them the points for the pickup. As they did succesfullly do that part of the task.

PetervDooren: Mainly I want to encourage teams to go for the pickup, which is what is often bypassed right now in favour of a handover. To do this I suggest to increase the score you can get for the pickup.

akinobu1998: In my opinion, the teams should go for the pickup (the rulebook encourages doing it), otherwise, they would tend to bypass picking up to get scores by placing.
I want to clarify the phrase, “Picking tableware and cutlery **for transportation**”. Let’s say a robot grasped an object but accidentally drops the object during transportation, does this situation considered to be successful for pickup? (I also saw this kind of situation in Tidy up under WRC rules.)
I suggest separate pickup+delivery and placement (If we separate pickup, delivery, and placement (three actions), the teams will get points just by lifting the object randomly.).

ARTenshi: Leaving anything to "judges' consideration" has always been a problem during the competition (teams and TC discussing their interpretations, and it will be even harder considering the number of qualified teams https://athome.robocup.org/2023-qualified-teams/ ); furthermore, pick-and-place is a very basic skill that we should take for granted in Stage II. "Clean the Table" with partial points is a Stage I task so, we are trying to make it a Stage II by considering full deliveries, complete classification, and so on.

ARTenshi: Taking cutlery or dishes has been tested in other competitions like WRS where those objects are considered with the same points as any other item (like a big bottle, a small dice, or a flat credit card).

Check 6.4 Smoothie Chef. That's more a Stage II task, where all points are for correct deposits (not for taking for transportation, delivery, and bonus for correct object). Furthermore, that task not only requires a spoon manipulation but pouring a teaspoon of sugar in a blender).

PetervDooren: It is to late to make such a big change now. This is about the scoring of Clean the table and whether or not this version is better than the current one on master. As a comparison:

|                        | master | this PR | PR #783   |
|------------------------|--------|---------|----------|
| Full delivery          | 280    | 280     | 240*     |
| Only pickup            | 60     | 120     | 0        |
| Only place (handover)  | 160    | 160     | 100      |
| only place incorrectly | 100    | 80      | 100      |

* 280 if all objects are placed correctly

ARTenshi: Basically, we have to decide if, for Stage II, we want a pick-and-place task or a pick-a d-maybe-place task. The main goals of this task are defined as

"Main goals: All tableware and cutlery on the table is placed inside the dishwasher (five objects in total)."

If this is the pick-and-maybe-place paradigm, we should change this goal.

PetervDooren: This refers to bonus points from for example opening the dishwasher. It is unrelated to what should be considered the main task. 
This is a matter of how we want to implement partial scoring for this challenge. 

But this discussion is bringing us nowhere. Please state which option you prefer and why. Then we can see if we have consensus. 
I prefer this version, I am willing to make it so doing only the handover scores less points (100) to encourage teams to complete the whole cycle. But I do want to keep a partial score for pickup so teams that can only pick up can still show what they can do.

ARTenshi: >This refers to bonus points from for example opening the dishwasher. 

This refers to the whole scoring to enforce teams to complete the core of the task and not only to "hunt" points. Ask other TCs or the EC if you have doubts.

>Please state which option you prefer and why.


I am for having pick-and-place as a single skill to make it clearer the evaluation and to avoid misinterpetations.

This is a Stage II task. Historically, Stage II has been harder than Stage I (above, I referred you to Smoothie Chef). Stating that a pick-and-place task (similar to Stage I Storing Groceries) is harder because we changed the objects is based in a subjective impression; to be objectives, the YCB cutlery and dishes object have been successfully used for robot manipulation through the years – moreover, in WRS using the HSR robot, most objects in the YCB dataset have been used (except for those that might be a serious risk) and have been scored with the same value (from bottles, cans and boxes to flat cards, small dices, and transparent glasses). Therefore, this is still a Stage I task reloaded and I can't find any way to meke it worth for Stage II.

PetervDooren:
This topic requires further discussion within the TC. Closing for now.

PR #785, opened 2023-03-29
Updated Deus ex Machina and Setup

ARTenshi:
Updated the definitions in Deus ex Machina section to avoid fixed scoring regulations and leave those to the specificites of each task and their own scoring systems.

Also, commented out the Welcome Reception subsection as it is not a technical aspect to be regulated in this rulebook. It has been left opened in case the EC strongly suggest to get it back.

### DIFF
--- a/Setup.tex
+++ b/Setup.tex
@@ -1,6 +1,8 @@
 \chapter{Setup and Preparation}
 \label{chap:setup_and_preparation}
-Prior to the RoboCup@Home competition, all arriving teams will have an opportunity to set up their robots and prepare for the competition in a \iterm{Setup \& Preparation} phase. This phase is scheduled to start on the first day of the competition, namely when the venue opens and the teams arrive. During the \SetupDays, teams can assemble and test their robots. On the last setup day, a \WelcomeReception{} will be held. To foster the knowledge exchange between teams a conference-like \PS{} takes place during the reception. Additionally, all teams have to get their robots inspected by members of the TC to be allowed to participate in the competition.
+Prior to the RoboCup@Home competition, all arriving teams will have an opportunity to set up their robots and prepare for the competition in a \iterm{Setup \& Preparation} phase. This phase is scheduled to start on the first day of the competition, namely when the venue opens and the teams arrive. During the \SetupDays, teams can assemble and test their robots.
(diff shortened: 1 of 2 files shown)


Setup.tex
```diff
@@ -42,26 +44,34 @@ \section{General Setup}
...
-			\item organizing snacks and drinks, and
-			\item inviting officials, sponsors, the local organization, and the trustees of the RoboCup Federation to the event.
-		\end{enumerate}
-	\item \textbf{Poster presentation:} During the \WelcomeReception, the teams give a poster presentation on their research focus, recent results, and their scientific contribution.
-	Both the poster and the teaser talk are evaluated by a jury (see~\refsec{sec:poster_teaser_session}).
-\end{enumerate}
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%
+% This Welcome Reception was suggested by the EC but it is not a technical aspect
+% It is still open for the TC to get it back is the EC strongly request it
```
PetervDooren, line 50: Lets add a timestamp so we know when this decision was made
```suggestion
% It is left commented for the TC to get it back is the EC strongly request it. 2023
```

PetervDooren (changes requested):
Regarding the score reduction I still have a few points: 
1. are the percentages still consistent with the rulebook?
2.  I think it might be a good idea to give indications of how much reduction should be given. For when we are making scoresheets for future tests.

ARTenshi:
I agree but it requires a further discussion ans consensus that might not happen in the next couple of days.

ARTenshi:
It should be done before releasing the scores as, if it becomes a general guideline, we should check all scoring sheets comply with it (and, by modifying the Deus ex Machina sections, we should check the final sums and modify other scores and so on). As it is written now, each "Deus ex Machina" section follows its own considerations.

PetervDooren:
The way this PR is currently written there is no difference between the different deus ex machinas. Whereas the latter should be more heavily penalised than the first ones. I would still like to see an indication of how severely these are penalised, even if we do not want to give a numeric amount.

PR #791, opened 2023-05-10
Fine Tuning 2023 Release

johaq:
Hello everybody,

Based on the Portugal Open we noticed some inconsistencies and unclear wordings in the rules.
These are the changes that should not affect a teams preparation since Bordeaux is so soon but are in my opinion still important to include in the 2023 release for a smoother run competition:

GPSR & EGPSR:
-Made the wording for non-expert operators clearer and added an example for illustration.
-Changed the deus ex machina rule to be applied accoding to other deus ex machina rules in the respective stage (also adjusted penalty for EGPSR 500 -> 800)

Serving Breakfast:
-Changed penalty for pointing at an object (should not be full penalty if the robot still picks up the object)
-Penalty for not picking and not placing now in order with points in main goal

Storing Groceries:
-Penalty for not picking and not placing now in order with points in main goal
-Penalty for recognizing and clearly indicating where to place an object but not placing now in order with points in main goal
-Human pointing where to place an object should not give full penatly if the robot still manages to place it there accurately

Clean the Table
-Penalty for pointing at an object consistent with serving breakfast
-Penalty for not picking and not placing in order with points in main goal

Restaurant
-Merged detecting and reaching table to one bullet point since just detection without reaching cannot be verified by the referee
-Skipping picking and placing needs to add up to 200 since picking and placing x3 needs to add up to 600. I changed it to 100/100 can be discussed if we want different split
-Being told where a guest is now gives -100 since it skips detection part

Side Note: I would like to change the penalty for skipping manipulation to be only applicable 2 times per order even for orders with 3 objects. As it stands teams can be penalized more for getting an order with 3 objects which is already harder to understand. I think we should still change this but I understand it is a bigger change. If you agree with this change and think it i still a fine change to make I will add it to the PR.

Stickler for the Rules
-Approaching a guest not breaking the rule should be fine only telling someone they are breaking a rule which they are not breaking should apply a penalty

### DIFF
--- a/scoresheets/Receptionist.tex
+++ b/scoresheets/Receptionist.tex
@@ -4,7 +4,7 @@ The maximum time for this test is 5 minutes.
 	\scoreheading{Main Goal}
 	\scoreitem[2]{250}{Introduce a new guest to every other guest and offer a seat}
 	\scoreitem{50}{Look at the person talking}
-	\scoreitem{50}{Look at the person being described}
+	\scoreitem{50}{Look at the person the robot is introducing the guest to}
 	\scoreitem{50}{Look in the direction of navigation}
 	\penaltyitem[2]{50}{Continue with wrong name or drink}
 	\penaltyitem[2]{50}{Persistent inappropriate gaze - away from conversational partner}
--- a/scoresheets/Restaurant.tex
+++ b/scoresheets/Restaurant.tex
@@ -6,21 +6,21 @@
 
 \small\begin{scorelist}
 	\scoreheading{Regular Rewards}	
-	\scoreitem[2]{100}{Detect calling or waving customer (per order).}
-	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training (per order).}
+	\scoreitem[2]{200}{Detect calling or waving customer and reach their table.}
+
 	\scoreitem[2]{600}{Take and serve an order.}
 
 	\scoreheading{Bonus Rewards}
-	\scoreitem[2]{200}{Use an unattached tray to transport (per order).}
+	\scoreitem[2]{200}{Use an unattached tray to transport.}
 	
 	\scoreheading{Regular Penalties}
-	\penaltyitem[2]{100}{Not making eye-contact when taking an order (per order).}
+	\penaltyitem[2]{100}{Not making eye-contact when taking an order.}
 
 	\scoreheading{Deus ex Machina Penalties}
-	\penaltyitem[2]{200}{Being guided to a table (per order).}
-	\penaltyitem[6]{100}{Bypassing manipulation (per object).}
-	\penaltyitem[6]{50}{Each object handover (per object).}
-	\penaltyitem[2]{25}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
+	\penaltyitem[2]{200}{Being guided to a table.}
+	\penaltyitem[6]{50}{Asking Barman to handover object.}
+	\penaltyitem[6]{50}{Handing over the object to the guest.}
+	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
 
 \end{scorelist}
 
--- a/scoresheets/Restaurant.tex
+++ b/scoresheets/Restaurant.tex
@@ -6,7 +6,8 @@
 
 \small\begin{scorelist}
 	\scoreheading{Regular Rewards}	
-	\scoreitem[2]{200}{Detect calling or waving customer and reach their table.}
+	\scoreitem[2]{100}{Detect calling or waving customer.}
+	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training.}
 
 	\scoreitem[2]{600}{Take and serve an order.}
 
@@ -18,8 +19,8 @@
 
 	\scoreheading{Deus ex Machina Penalties}
 	\penaltyitem[2]{200}{Being guided to a table.}
-	\penaltyitem[6]{100}{Asking the Barman to handover object to the robot.}
-	\penaltyitem[6]{100}{Guest needing to take the object from a tray or the robot's hand}
+	\penaltyitem[4]{100}{Asking the Barman to handover object to the robot.}
+	\penaltyitem[4]{100}{Guest needing to take the object from a tray or the robot's hand}
 	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar}.}
 
 \end{scorelist}
--- a/scoresheets/Restaurant.tex
+++ b/scoresheets/Restaurant.tex
@@ -18,8 +18,8 @@
 
 	\scoreheading{Deus ex Machina Penalties}
 	\penaltyitem[2]{200}{Being guided to a table.}
-	\penaltyitem[3]{100}{Asking the Barman to handover object to the robot.}
-	\penaltyitem[3]{100}{Guest needing to take the object from a tray or the robot's hand}
+	\penaltyitem[6]{100}{Asking the Barman to handover object to the robot.}
+	\penaltyitem[6]{100}{Guest needing to take the object from a tray or the robot's hand}
 	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar}.}
 
 \end{scorelist}
--- a/scoresheets/Restaurant.tex
+++ b/scoresheets/Restaurant.tex
@@ -18,9 +18,9 @@
 
 	\scoreheading{Deus ex Machina Penalties}
 	\penaltyitem[2]{200}{Being guided to a table.}
-	\penaltyitem[6]{50}{Asking Barman to handover object.}
-	\penaltyitem[6]{50}{Handing over the object to the guest.}
-	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
+	\penaltyitem[3]{100}{Asking the Barman to handover object to the robot.}
+	\penaltyitem[3]{100}{Guest needing to take the object from a tray or the robot's hand}
+	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar}.}
 
 \end{scorelist}
 
--- a/scoresheets/SticklerForRules.tex
+++ b/scoresheets/SticklerForRules.tex
@@ -17,11 +17,11 @@
 	\scoreitem[4]{100}{Making eye-contact, politely clarify to the guest what rule is being broken.}
 
 	\scoreheading{Regular Penalties}
-	\penaltyitem[4]{100}{Approach a guest who is not breaking a rule.}
+	\penaltyitem[4]{100}{Talking to a guest about a rule they are not breaking.}
 
 	\scoreheading{Deus ex Machina Penalties}
-	\penaltyitem[4]{100}{A human directs the robot towards a guest who is breaking a rule.}
-	\penaltyitem[4]{100}{A human tells the robot which guest is breaking a rule.}
+	\penaltyitem[4]{50}{A human directs the robot towards a guest who is breaking a rule.}
+	\penaltyitem[4]{100}{A human tells the robot which rule is being broken.}
 
 \end{scorelist}
 
--- a/tasks/EGPSR.tex
+++ b/tasks/EGPSR.tex
@@ -30,7 +30,7 @@ Similar to a smart speaker, the robot is asked to execute arbitrary commands req
 		\begin{itemize}
 			\item \textbf{Operators:} A \emph{Professional Operator} (the referee) commands the robot to execute a task.
 			Optionally, commands can be issued by a \emph{Non-Expert Operator}, i.e. a person from the audience with no robotics background.
-			In this case, the referee gives the goal of the command to the non-expert operator, who will then issue it to the robot in their own words. For example, the generated command might be "Bring me a coke from the kitchen." then the non-expert operator will be told "The robot should bring you a coke, which is found in the kitchen.", who then tells the robot "I want a coke. Go to the kitchen and get me one."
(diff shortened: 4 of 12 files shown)


tasks/GPSR.tex
```diff
@@ -28,7 +28,7 @@ \subsection*{Setup}
         \begin{itemize}
             \item \textbf{Operators:} A \emph{Professional Operator} (the referee) commands the robot to execute a task.
             Optionally, commands can be issued by a \emph{Non-Expert Operator}, i.e. a person from the audience with no robotics background.
-            In this case, the referee gives the command to the non-expert operator, who will then issue it to the robot (rephrasing is allowed).
+            In this case, the referee gives the goal of the command to the non-expert operator, who will then issue it to the robot in their own words. For example, the generated command might be "Bring me a coke from the kitchen." then the non-expert operator will be told "The robot should bring you a coke, which is found in the kitchen.", who then tells the robot "I want a coke. Go to the kitchen and get me one."
```
MatthijsBurgh, line 31: Again, why is the diff marked red here?

MatthijsBurgh (changes requested):

scoresheets/Restaurant.tex
```diff
@@ -6,21 +6,21 @@
 
 \small\begin{scorelist}
 	\scoreheading{Regular Rewards}	
-	\scoreitem[2]{100}{Detect calling or waving customer (per order).}
-	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training (per order).}
+	\scoreitem[2]{200}{Detect calling or waving customer and reach their table.}
```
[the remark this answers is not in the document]
johaq: Because as a referee I cannot verify whether the robot actually detected someone who wants to order without the robot reaching the table

scoresheets/Restaurant.tex
```diff
@@ -6,21 +6,21 @@
...
 
 	\scoreheading{Deus ex Machina Penalties}
-	\penaltyitem[2]{200}{Being guided to a table (per order).}
-	\penaltyitem[6]{100}{Bypassing manipulation (per object).}
-	\penaltyitem[6]{50}{Each object handover (per object).}
-	\penaltyitem[2]{25}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
+	\penaltyitem[2]{200}{Being guided to a table.}
+	\penaltyitem[6]{50}{Asking Barman to handover object.}
+	\penaltyitem[6]{50}{Handing over the object to the guest.}
+	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
```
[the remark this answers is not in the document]
johaq: Yes absolutely

scoresheets/Restaurant.tex
```diff
@@ -6,21 +6,21 @@
...
+	\penaltyitem[2]{100}{Not making eye-contact when taking an order.}
 
 	\scoreheading{Deus ex Machina Penalties}
-	\penaltyitem[2]{200}{Being guided to a table (per order).}
-	\penaltyitem[6]{100}{Bypassing manipulation (per object).}
-	\penaltyitem[6]{50}{Each object handover (per object).}
-	\penaltyitem[2]{25}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
+	\penaltyitem[2]{200}{Being guided to a table.}
+	\penaltyitem[6]{50}{Asking Barman to handover object.}
+	\penaltyitem[6]{50}{Handing over the object to the guest.}
```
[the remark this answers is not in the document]
johaq: Ah yes should include both. This needs to be adjusted to 3*100 for both anyway...

scoresheets/Restaurant.tex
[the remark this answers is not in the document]
MatthijsBurgh: Oh now I see your explanation.

I don't agree with it. As the robot could show a picture of the guest or similar. So I would go for separate items, but add the referee should be able to verify it in case of not reaching the guest close enough.

tasks/EGPSR.tex
```diff
@@ -29,9 +29,9 @@ \subsection*{Setup}
 	\item \textbf{People:}
 		\begin{itemize}
 			\item \textbf{Operators:} A \emph{Professional Operator} (the referee) commands the robot to execute a task.
-            Optionally, commands can be issued by a \emph{Non-Expert Operator}, i.e. a person from the audience with no robotics background.
-            In this case, the referee gives the command to the non-expert operator, who will then issue it to the robot (rephrasing is allowed).
-            If the robot consistently fails to understand the non-expert operator (e.g. after two retries), teams can default to a custom operator.
+			Optionally, commands can be issued by a \emph{Non-Expert Operator}, i.e. a person from the audience with no robotics background.
+			In this case, the referee gives the goal of the command to the non-expert operator, who will then issue it to the robot in their own words. For example, the generated command might be "Bring me a coke from the kitchen." then the non-expert operator will be told "The robot should bring you a coke, which is found in the kitchen.", who then tells the robot "I want a coke. Go to the kitchen and get me one."
```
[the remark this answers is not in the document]
johaq: Hm apparently this is marked as syntax error. Should probably change it to this ``text''

tasks/GPSR.tex
johaq: This one is weirder. Maybe changing the quotation marks work here as well

scoresheets/Restaurant.tex
```diff
@@ -18,9 +18,9 @@
 
 	\scoreheading{Deus ex Machina Penalties}
 	\penaltyitem[2]{200}{Being guided to a table.}
-	\penaltyitem[6]{50}{Asking Barman to handover object.}
-	\penaltyitem[6]{50}{Handing over the object to the guest.}
-	\penaltyitem[2]{100}{Being told/pointed where is a table/\textit{Kitchen-bar} (per order).}
+	\penaltyitem[3]{100}{Asking the Barman to handover object to the robot.}
```
MatthijsBurgh, line 21: Why 3 times? As you can only take an order from a guest 2 times, correct?

MatthijsBurgh (changes requested):

scoresheets/Restaurant.tex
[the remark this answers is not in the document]
johaq: Hm, maybe. It would need to be clear in that case. Like a cropped picture of just the guest or a bounding box.

scoresheets/Restaurant.tex
johaq: So, as it is written right now. If you get an order with 2 objects there are 2 pick and 2 place actions that can be skipped. But for an order with 3 objects there are 3 each that can be skipped. Therefore 3 penalties (or 6 combined as it was written before).
As I mentioned above I think this is bad because it makes getting an order with 3 objects even more punishing. So in my opinion for an order with 3 objects even if 3 pick and 3 place actions are skipped the penalty should only be applied twice to make it even with 2 objects orders. But I wanted some feedback for this because it seemed like a bigger change.

MatthijsBurgh: With your changes, bringing an order with attached container(like Tech United did in Portugal), will result in zero points for 1 and 2 item orders. And a 3 item order would result in 200 points.

I think that is penalizing too much.

johaq: No, how it currently is you'd get 0 points for a three item order and 200 points for a 2 item order. I would like to change it so your method would give 200 points in both cases

MatthijsBurgh: You will never a 1 item order?

johaq: Right, discrepancy is even worse then...
I already think having 2 and 3 item orders is bad because of the difference in difficulty 1 vs 3 is obviously worse.
@MatthijsBurgh Do you think it is still fine to change it to order is 2 to 3 objects and penalties can only be applied to two objects per order?

MatthijsBurgh: I agree that bringing an order without manipulation should result in the same amount of points for 2 and 3 item orders.

So when bringing an order is 600. Skipping manipulating the object is 100 points per skip. And skipping can happen both at the bar and the guest table. When limiting the reduction to only to 2 objects, so 4 skip. You end with 200 points.

So I don't think it matters whether the skips are at the bar or the guest table. So just 4 skips would also be fine by me.

I also agree with you we can drop 1 item orders. Or we should change the scoring, so that manipulation grants points instead of penalizing not manipulating.

scoresheets/Restaurant.tex
```diff
@@ -6,21 +6,22 @@
 
 \small\begin{scorelist}
 	\scoreheading{Regular Rewards}	
-	\scoreitem[2]{100}{Detect calling or waving customer (per order).}
-	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training (per order).}
+	\scoreitem[2]{100}{Detect calling or waving customer.}
+	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training.}
```
MatthijsBurgh, line 10:
```suggestion
	\scoreitem[2]{100}{Detect calling or waving customer}
	\scoreitem[2]{100}{Reach a customer's table without prior guidance/training}
```

scoresheets/Restaurant.tex
```diff
@@ -6,21 +6,22 @@
...
+
 	\scoreitem[2]{600}{Take and serve an order.}
 
 	\scoreheading{Bonus Rewards}
-	\scoreitem[2]{200}{Use an unattached tray to transport (per order).}
+	\scoreitem[2]{200}{Use an unattached tray to transport.}
```
MatthijsBurgh, line 15:
```suggestion
	\scoreitem[2]{200}{Use an unattached tray to transport}
```

scoresheets/Restaurant.tex
```diff
@@ -6,21 +6,22 @@
...
 	
 	\scoreheading{Regular Penalties}
-	\penaltyitem[2]{100}{Not making eye-contact when taking an order (per order).}
+	\penaltyitem[2]{100}{Not making eye-contact when taking an order.}
```
MatthijsBurgh, line 18:
```suggestion
	\penaltyitem[2]{100}{Not making eye-contact when taking an order}
```

scoresheets/SticklerForRules.tex
```diff
@@ -17,11 +17,11 @@
 	\scoreitem[4]{100}{Making eye-contact, politely clarify to the guest what rule is being broken.}
 
 	\scoreheading{Regular Penalties}
-	\penaltyitem[4]{100}{Approach a guest who is not breaking a rule.}
+	\penaltyitem[4]{100}{Talking to a guest about a rule they are not breaking.}
```
MatthijsBurgh, line 20:
```suggestion
	\penaltyitem[4]{100}{Talking to a guest about a rule they are not breaking}
```

scoresheets/SticklerForRules.tex
```diff
@@ -17,11 +17,11 @@
...
 
 	\scoreheading{Deus ex Machina Penalties}
-	\penaltyitem[4]{100}{A human directs the robot towards a guest who is breaking a rule.}
-	\penaltyitem[4]{100}{A human tells the robot which guest is breaking a rule.}
+	\penaltyitem[4]{50}{A human directs the robot towards a guest who is breaking a rule.}
```
MatthijsBurgh, line 23:
```suggestion
	\penaltyitem[4]{50}{A human directs the robot towards a guest who is breaking a rule}
```

scoresheets/SticklerForRules.tex
```diff
@@ -17,11 +17,11 @@
...
+	\penaltyitem[4]{100}{A human tells the robot which rule is being broken.}
```
MatthijsBurgh, line 24:
```suggestion
	\penaltyitem[4]{100}{A human tells the robot which rule is being broken}
```

scoresheets/Restaurant.tex
MatthijsBurgh, line 18:
```suggestion
	\penaltyitem[2]{100}{Not making eye-contact when taking an order}
```

MatthijsBurgh (approved):

MatthijsBurgh:
@johaq I approved these changes. But you might want to get another approval from an official TC member.

johaq:
Some more changes. Mainly gramar fixes and clarifications
-Specifiy that dishwasher tab needs to be placed inside hatch for bonus points
-In receptionist look at the person being described contradicts how introducing is described above
-Lots of grammar fixes in stickler for the rules

MatthijsBurgh (approved):

scoresheets/Receptionist.tex
```diff
@@ -4,7 +4,7 @@
 	\scoreheading{Main Goal}
 	\scoreitem[2]{250}{Introduce a new guest to every other guest and offer a seat}
 	\scoreitem{50}{Look at the person talking}
-	\scoreitem{50}{Look at the person being described}
+	\scoreitem{50}{Look at the person the robot is introducing the guest to}
```
PetervDooren, line 7: I think this wording works better for clarity
```suggestion
	\scoreitem{50}{Look at the person the robot is talking to}
```

tasks/Restaurant.tex
```diff
@@ -41,7 +41,7 @@ \subsection*{Setup}
 	\item \textbf{Objects:} 
 	\begin{itemize}
 	\item Objects to fulfill orders are located on the Kitchen-bar.
-	\item Orders have between one and three objects randomly chosen.
+	\item Orders have between two and three objects randomly chosen.
```
PetervDooren, line 44: there is technically no integer beteen two and three. slight grammar point
```suggestion
	\item Orders have two or three objects randomly chosen.
```

PetervDooren (approved):
All of these changes make sense and add either clarification or an adjustment to the scores.

We should check if we want to incorporate these for the version of 2023 as that one is already Final.

johaq:
Yes, I don't think these changes effect preparation much at all and make the rules a lot more clear and consistent

MatthijsBurgh:
@PetervDooren @johaq I will create a new release

Issue #805, opened 2023-07-31
Getting away from state machines (more open world)


PetervDooren:
## Is your idea/suggestion related to a problem? Please describe.
Currently many teams use Finite state machines to do their challenges. This is not a problem in and of itself. However it often leads to a robot going through the same steps making extensive use of its knowledge of where things will be for a given challenge. Rather we'd like to move to challenges that are formatted more like Restaurant and Stickler for the rules.

- Issue outlined in TC meeting 31/07/2023

PetervDooren:
For the most part this can be solved by ensuring the robot needs to react to at least some sensor information. For example the place where the breakfast must be served.

It can also be the way in which humans are instructed to interact with the robots. Now humans will follow the instructions of the robot even when the interaction is unnatural.

In general this is a problem that needs to be adressed per challenge. A solution that works for one might not work for another.

sunava:
Perhaps we should go back to visualizing things; therefore, Visbox was originally designed so that referees could see what the robot perceived. If this is not possible due to network issues or other reasons, it should appear on the robot's display where it perceived an object. Furthermore, any type of human-robot interaction should proceed in a way where the robot doesn't say, "please give me the milk," but rather, "please give me the milk on the top-right shelf, third row."

If the robot does not follow these given rules, the referee will not acknowledge it. Only in this way can we ensure that perception was genuinely used and that there is a certain knowledge base on the robot. Additionally, unknown objects that do not belong to the challenge should be reintroduced and should not be used. For example, in a "serve breakfast" challenge, there could be ice tea on the shelf. Therefore, for each challenge, we should clearly define what is allowed in terms of visualization and human-robot interaction to prevent situations where the robot simply says, "please give me the spoon," and the spoon is obviously part of the task, so it will be there.

sunava:
Template for pdf report instead of vizbox stuff discussion on 20.11.23 for perception.

sunava:
idea: check box  for "check report for locations cause HRI was not specific enough"

MatthijsBurgh:
I think you should focus on the desired robot behaviour, not its implementation.

sunava:
We decided to move this to the next iteration of the rulebook

PR #833, opened 2023-11-20
[state-machine] added rules for communication

sunava:
## Description
We have introduced new rules for communication and added visualization guidance in the "Serve Breakfast" challenge. These changes are aimed at fostering a more dynamic and less state-machine-dependent approach.

Communication Enhancements
Detailed Interaction: Robots are required to use specific and detailed language in their requests. Generic phrases like "please give me [object]" are unacceptable.
Example: A robot should say, "Please hand me the milk carton from the top-right shelf, third row," instead of a vague request.
Referee Role: Referees will enforce these communication standards strictly. Requests that lack specificity will be ignored.
Visualization Guidance
Perception Display: Robots must have a mechanism to display their internal perception data. This ensures referees can visually confirm what the robot is perceiving.

Closes issue #
Addresses issue: https://github.com/RoboCupAtHome/RuleBook/issues/805#issue-1829292526

- {Specific Communication Guidelines}
- {Advanced Visual Perception Standards: }

## Other comments
This is only for one challenge; this guideline will be done for every challenge.

### DIFF
--- a/tasks/ServeBreakfast.tex
+++ b/tasks/ServeBreakfast.tex
@@ -52,6 +52,8 @@ The robot has to set a table for breakfast for one person and prepare cereal for
 \subsection*{Additional Rules and Remarks}
 \begin{enumerate}[nosep]
 	\item \textbf{Safe placing:} Objects must be placed with care, namely the robot should place rather than throw or drop objects.
+	\item \textbf{Specific Communication Guidelines}: The robot must use precise and detailed language when interacting with humans. For example, instead of saying, "Please give me the milk," it should say, "Please give me the milk carton located on the top-right shelf, third row."
+	\item \textbf{Advanced Visual Perception Standards: }The robot must display its internal perception data on its own screen, or VizBox, enabling referees to visually verify what the robot detects during the task.
 	\item \textbf{Deus ex Machina:} The scores are reduced if human assistance is received, in particular for:
 	\begin{itemize}[nosep]
 		\item pointing to an object or telling to the robot where an object is or where to place it
@@ -74,6 +76,7 @@ The referee needs to:
 \begin{itemize}
 	\item Remove all objects from the table.
 	\item Place all objects in their default locations.
+	\item Referees will strictly enforce communication guidelines. Non-specific requests by the robot will be disregarded.
 \end{itemize}
 
 % \newpage


LeroyR:
**Detailed Interaction:** Generic Phrases like that are already disallowed afaik. If this is not already handled with Rule `3.8 Deus Ex Machina`. We should just expand the rule if it is lacking in regard to unspecific language.

**Perception Display:** Is this even possible? Standard Platforms can not be modified and the screen is in the front of the robot -> Referee potentially blocks the robot during manipulation.
For OPL, i dont want to enforce any design but we could request that the Robot has a display at the back.

VizBox/custom external display has the obvious networking issues: communication error -> score penalty? Seems harsh as this is not influenced by the team. We dont want additional scoring discussions because the robot displays different images too quickly or the Referees missed something.

Some years ago we had Tasks where the Team/Robot had to produce a `pdf-report` afterwards, which feels like a better solution.

sunava:
First, thank you for your feedback.

I've noticed that in the latest RoboCup, there were instances where referees provided items to robots even when the robots did not communicate as required. To address this, I propose we make it explicitly clear in the rules for each challenge what the specific communication requirements are. This will help ensure consistency in how referees respond to robot requests.

In particular, we should highlight that:

Robots must use detailed and specific language when requesting items.
Referees must adhere strictly to these guidelines and not respond to any requests that don't meet the specified criteria.
This approach will reinforce the importance of precise communication and ensure a fair and uniform standard across all challenges.

A few weeks ago, we discussed the possibility of allowing competitors to use the VizBox in our competitions. In the latest rulebook, the VizBox was actually a requirement, and several teams invested effort into supporting it. However, it turned out that the committee did not fully support this implementation.

After considering your suggestion, I find the idea of a PDF report more appealing than the VizBox or internal display approach. It seems like a more efficient and accessible way to handle the data presentation.

On that note, are there any existing rules or guidelines regarding the use of PDF reports? If there's a document or a set of standards about this, it would be helpful to review them to better understand how we can integrate this method into our competition.

sunava:
Closing this PR after talking to Alex, Leroy and Juan about it. -> I will look into a template of a pdf report.

ARTenshi:
Before closing this one, as a reminder, the PDF file was removed because it was more of a problem than a solution. First, there wasn't a standard format (among teams or tasks), and it was a burden for the referees to check every PDF after the slots and to remember what had happened; also, the PDF might not coincide with the reality -- while the PDF recognition might be correct, the robot might have taken the wrong object (especially when the objects are too close) and, especially with partial points, teams tend to reclaim points long after the task frame has finished.

Issue #806, opened 2023-07-31
The rulebook should mandate that we actually solve the technical challenges


PetervDooren:
## Is your idea/suggestion related to a problem? Please describe.
Somewhat related to #805. Many times we see teams make extensive use of prior knowledge to hard-code a solution that gives points at a tournament but which does not solve the technical challenge.

(Issue outlined in TC meeting 31/07/2023)

akinobu1998:
Ideas for manipulation tasks:
* Use unknown objects to prevent the robot from bypassing manipulation without perception
* Change furniture location before the task or use the random picking-up location for each test.
* Introduce non-expert helper (volunteer)


Currently, some robots just say the object name to bypass the manipulation, "Give me the xxx." for example. Especially for clean the table, by using some unknown objects as well as the known objects, at least we can prevent this behavior. For example, in storing groceries, store "unknown food" in to "food" or "unknown" category. It is not necessary to know the object name to manipulate objects.

As @PetervDooren mentioned, moving furniture should work, or use random pick-up locations each trial (announce some possible pick-up locations 2 hours before the test).

Introducing a non-expert helper can also make the deus ex machina harder. Non-expert helpers might not know how the robot grasps the object or even the object's name. The robot should clearly state the object that the robot wants to manipulate.

PetervDooren:
Your idea of using unknown objects would work by having the manipulation objects be a subset of a larger set, so the robot does not know a priori what needs to happen. However in clean the table we can achieve this by using a larger set of cutlery. like the bowl and a mug or cup. The rulebook already specifies that only two of spoon, fork, knife are present. 

I would not make deus ex machina harder. The robot is already asking for help, indicating it cant solve the problem. With the scoring strategy described in #808 teams are heavily disincentivised from using deus ex machina. So I think that is already fair.

ARTenshi:
I wouldn't make deus ex machina harder, either, but I agree with @akinobu1998 on making deus ex machina real (not just open loop voice requests).

akinobu1998:
For objects in clean the table, in 2023, only cutlery from YCB objects was used. To evaluate the robustness of manipulation, we can also use another type of cutlery  (and clean the table is a stage II task focused on manipulation). And, yes, it is already specified in the rulebook, we can achieve it in the object selection phase.

Thank you for your opinion, I would like to propose the idea of a non-expert helper to make the deus ex machina more natural no to make it hard.

PetervDooren:
[...]

PR #838, opened 2023-12-17
The rulebook should mandate that we actually solve the technical challenges #806

akinobu1998:

## Description

Closes issue #806

- Serve breakfast: referee decide the surface for the table instead of team
- Add instruction for Deus ex Machina

### DIFF
--- a/general_rules/ContinueRules.tex
+++ b/general_rules/ContinueRules.tex
@@ -15,6 +15,7 @@ The procedure to request human assistance while solving a task is as follows:
 		\item The particular goal or desired result
 		\item How the action must be carried out (when necessary)
 		\item Details about how to interact with the robot (when necessary)
+		\item Detailed information to identify the objects (e.g. object name, color, location) for picking and placing objects.
 	\end{compactitem}
 
 	\item \textbf{Supervise:} The robot must be aware of the human's actions, being able to tell when the requested action has been completed, as well as guiding the human assistant (if necessary) during the process.
--- a/general_rules/ContinueRules.tex
+++ b/general_rules/ContinueRules.tex
@@ -15,7 +15,7 @@ The procedure to request human assistance while solving a task is as follows:
 		\item The particular goal or desired result
 		\item How the action must be carried out (when necessary)
 		\item Details about how to interact with the robot (when necessary)
-		\item Detailed information enough to identify the objects (e.g. object name, color or location) for picking and placing objects.
+		\item Detailed information to identify objects for picking and placing (e.g. object name, color or location). The provided information needs to show that objects were perceived by the robot.
 	\end{compactitem}
 
 	\item \textbf{Supervise:} The robot must be aware of the human's actions, being able to tell when the requested action has been completed, as well as guiding the human assistant (if necessary) during the process.
--- a/general_rules/ContinueRules.tex
+++ b/general_rules/ContinueRules.tex
@@ -15,7 +15,7 @@ The procedure to request human assistance while solving a task is as follows:
 		\item The particular goal or desired result
 		\item How the action must be carried out (when necessary)
 		\item Details about how to interact with the robot (when necessary)
-		\item Detailed information to identify the objects (e.g. object name, color, location) for picking and placing objects.
+		\item Detailed information enough to identify the objects (e.g. object name, color or location) for picking and placing objects.
 	\end{compactitem}
 
 	\item \textbf{Supervise:} The robot must be aware of the human's actions, being able to tell when the requested action has been completed, as well as guiding the human assistant (if necessary) during the process.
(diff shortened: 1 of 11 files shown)


general_rules/ContinueRules.tex
```diff
@@ -15,6 +15,7 @@ \subsection{Procedure}
 		\item The particular goal or desired result
 		\item How the action must be carried out (when necessary)
 		\item Details about how to interact with the robot (when necessary)
+		\item Detailed information to identify the objects (e.g. object name, color, location) for picking and placing objects.
```
johaq, line 18: I think this is a very necessary change. But imo this needs to be a bit clearer. Does detailed information include one of the examples or all of them? 
I think the point of this change is so that teams don't use the Deus Ex Machina rules intended to skip manipulation to also skip perception. Now the requirements for this are different for different tests. In storing groceries the objects on the table are unknown so a robot showing me just a picture of an object it wants handed over is enough to show perception. In Serve Breakfast the robot know a bowl will be there so just showing a picture of the bowl does not show perception. Do you agree? Any idea how we can include this distinction. For example I think a robot trying to pick something up and failing and then asking for a handover always clearly shows perception.

johaq (commented):
I would also like to request changes to Section 3.8.1, especially the example given.
I think if we allow DEM rule use only if perception is shown the referee intentionally not following the robot's orders seems out of place

general_rules/ContinueRules.tex
akinobu1998: @johaq 
Thank you for your feedback, I just pushed a new commit.
I modified the example as well.
I have not mentioned particular task names or instructions because I think the rulebook should not force the team/robot to follow a particular procedure.

general_rules/ContinueRules.tex
```diff
@@ -15,7 +15,7 @@ \subsection{Procedure}
 		\item The particular goal or desired result
 		\item How the action must be carried out (when necessary)
 		\item Details about how to interact with the robot (when necessary)
-		\item Detailed information to identify the objects (e.g. object name, color, location) for picking and placing objects.
+		\item Detailed information enough to identify the objects (e.g. object name, color or location) for picking and placing objects.
```
johaq, line 18: I would change this to: 
Detailed information to identify objects for picking and placing (e.g. object name, color or location). Provided information needs to show that objects were perceived by the robot.

johaq (changes requested):
Example looks good.
I would add the perception bit to the rule section.

johaq (approved):

general_rules/ContinueRules.tex
johaq: Great. Thank you! I approved the changes.

myszzka96 (dismissed):
Looks good. One suggestion for Serve Breakfast:  currently the location of the surface for the test is to be defined 2h before the test. I would say this is likely to be done earlier. Maybe we should put "at least 2h before the test"? Otherwise all looks good

akinobu1998:
The last commit is minor changes, I just change 2 hours before to **at least** two hours before.

johaq (dismissed):

tasks/ServeBreakfast.tex
```diff
@@ -38,7 +38,7 @@ \subsection*{Setup}
 
 \subsection*{Procedure}
 \begin{enumerate}[nosep]
-	\item \textbf{Table selection:} Two hours before the test, the referees announce the surface that will be used as a table.
+	\item \textbf{Table selection:} At least two hours before the test, the referees announce the surface that will be used as a table.
```
MatthijsBurgh, line 41: Lets be consistent with either `2 hours` or `two hours`. I prefer the last one.

tasks/ServeBreakfast.tex
```diff
@@ -22,13 +22,13 @@ \subsection*{Setup}
...
 			\begin{itemize}
-				\item No people are involved in the test, unless the robot requires human assistance.
+				\item No people are involved in the test unless the robot requires human assistance.
 			\end{itemize}
 	\item \textbf{Furniture:}
 		\begin{itemize}
-			\item \textbf{Table:} The robot serves breakfast on any table or flat surface in the kitchen (the team is free to choose this location).
-			\item \textbf{Chairs:} Chairs may be placed around the kitchen table and won't be removed.
-			\item \textbf{Doors:} The robot does not need to open any doors for finding the breakfast items.
+			\item \textbf{Table:} The robot serves breakfast on the table or flat surface which is announced beforehand.
```
PetervDooren, line 29: Keep this to the 'dinner table'. Any arena is gonna contain that. 
```suggestion
			\item \textbf{Table:} The robot serves breakfast on the table which is announced beforehand.
```

tasks/ServeBreakfast.tex
```diff
@@ -68,6 +68,13 @@ \subsection*{OC Instructions}
 	\item Provide official cutlery and tableware for training.
 \end{itemize}
 
+At least 2 hours before the test:
+\begin{itemize}
+	\item Announce the surface that will be used as a table.
```
PetervDooren, line 73:
```suggestion
	\item Announce the table that will be used.
```

PetervDooren (commented):
minor comments. but looks good

PetervDooren (dismissed):

YuqianJiang (dismissed):

akinobu1998:
The conflict was resolved, and ready for merge.

PetervDooren (approved):

johaq (approved):

sunava (approved):

PR #844, opened 2024-02-26
EGPSR Overhaul

johaq:
1st draft for egpsr overhaul:

Instead of commands given at static location, the robot has to find and solve problems on its own. 
The possibilities are:
- An object is at the wrong location, bring it back to the correct one
- An object is on the floor, put it in the bin or bring it back to the assigned location
- A person can have a request for the robot, solve the request (taken from gpsr commands)

See https://github.com/johaq/CommandGenerator/tree/master for setup generation

### DIFF
--- a/scoresheets/EGPSR.tex
+++ b/scoresheets/EGPSR.tex
@@ -2,14 +2,12 @@ The maximum time for this test is 10 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[3]{800}{Executing the task associated with each command}
+	\scoreitem[3]{800}{Find and solve a problem}
 
 	\scoreheading{Bonus rewards}
-	\scoreitem[3]{200}{Understanding a command given by a non-expert operator}
+	\scoreitem[1]{800}{Move a piece of furniture to its original position}
 
 	\scoreheading{Deus Ex Machina Penalties}
-	\penaltyitem[3]{100}{Using a custom operator}
-	\penaltyitem[3]{100}{Bypassing speech recognition}
 	\penaltyitem[3]{800}{Instructing a human to perform parts of the task will apply a percentage penalty according to similar penalties in other Stage II tests.}
 \end{scorelist}
 
--- a/scoresheets/EGPSR.tex
+++ b/scoresheets/EGPSR.tex
@@ -2,7 +2,8 @@ The maximum time for this test is 10 minutes.
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[3]{800}{Find and solve a problem}
+	\scoreitem[3]{150}{Find and clearly state an encountered problem}
+	\scoreitem[3]{650}{Solve a problem}
 
 
 	\scoreheading{Deus Ex Machina Penalties}
--- a/scoresheets/EGPSR.tex
+++ b/scoresheets/EGPSR.tex
@@ -4,10 +4,9 @@ The maximum time for this test is 10 minutes.
 	\scoreheading{Main Goal}
 	\scoreitem[3]{800}{Find and solve a problem}
 
-	\scoreheading{Bonus rewards}
-	\scoreitem[1]{800}{Move a piece of furniture to its original position}
 
 	\scoreheading{Deus Ex Machina Penalties}
+	\penaltyitem[3]{200}{Asking for location of a problem}
 	\penaltyitem[3]{800}{Instructing a human to perform parts of the task will apply a percentage penalty according to similar penalties in other Stage II tests.}
 \end{scorelist}
 
--- a/tasks/EGPSR.tex
+++ b/tasks/EGPSR.tex
@@ -18,7 +18,7 @@ The arena is in its default state apart from problems set up for the robot to so
 \begin{enumerate}
     \item \textbf{Problems:}
 		\begin{itemize}
-			\item \textbf{Objects:} Objects that are not in their default location should be returned there. Objects on the floor are to be thrown in the trash or returned to their default location.
+			\item \textbf{Objects:} Objects that are not in their default location should be returned there (see \ref{rule:scenario_objects}). Objects on the floor are to be thrown in the trash or returned to their default location.
 			\item \textbf{Persons:} Some persons in the arena will have requests for the robot. They will raise their hand if the robot is in the same room.
 		\end{itemize}
 \end{enumerate}
--- a/tasks/EGPSR.tex
+++ b/tasks/EGPSR.tex
@@ -1,6 +1,6 @@
 \section{Enhanced General Purpose Service Robot}
 \label{test:egpsr}
-The robot is asked to find problems in the arena on its own and solve them.
+The robot is asked to maintain the household by cleaning up the arena and assisting people.
 
 
 \noindent \textbf{Main Goal:} Solve 3 problems in the arena.\\
@@ -14,7 +14,7 @@ The robot is asked to find problems in the arena on its own and solve them.
 % Setup
 % %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection*{Setup}
-Generally assume that the arena is supposed to be in its default state. Therefore, possible problems to solve are:
+The arena is in its default state apart from problems set up for the robot to solve:
 \begin{enumerate}
     \item \textbf{Problems:}
 		\begin{itemize}
--- a/tasks/EGPSR.tex
+++ b/tasks/EGPSR.tex
@@ -3,12 +3,8 @@
 The robot is asked to find problems in the arena on its own and solve them.
 
 
-\noindent \textbf{Main Goal:} Solve 3(5?) problems in the arena.\\
+\noindent \textbf{Main Goal:} Solve 3 problems in the arena.\\
 
-\noindent \textbf{Optional Goals:}
-\begin{enumerate}[nosep]
-	\item Maybe move furniture?
-\end{enumerate}
 
 \subsection*{Focus}
 \emph{Task planning}, \emph{object/people detection and recognition}, \emph{object feature recognition}, \emph{object manipulation}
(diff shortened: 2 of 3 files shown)


tasks/EGPSR.tex
```diff
@@ -1,14 +1,13 @@
 \section{Enhanced General Purpose Service Robot}
 \label{test:egpsr}
-Similar to a smart speaker, the robot is asked to execute arbitrary commands requested by an operator.\\
-\textbf{Note 1:} Due to this task's more open-ended nature, the maximum possible score is higher than the scores of the other \STWO{} tasks.\\
-\textbf{Note 2:} The commands in this task are more complex than in the GPSR task in \SONE.\\
+The robot is asked to find problems in the arena on its own and solve them.
 
-\noindent \textbf{Main Goal:} Execute 3 commands requested by the operator.\\
+
+\noindent \textbf{Main Goal:} Solve 3(5?) problems in the arena.\\
```
PetervDooren, line 6: robot has 10 min for this challenge. So I think 5 might be feasible. You can also not specify the number of problems and just go for as many as possible.

tasks/EGPSR.tex
```diff
@@ -19,19 +18,12 @@ \subsection*{Focus}
...
-		\end{itemize}
-	\item \textbf{People:}
+    \item \textbf{Problems:}
 		\begin{itemize}
-			\item \textbf{Operators:} A \emph{Professional Operator} (the referee) commands the robot to execute a task.
-			Optionally, commands can be issued by a \emph{Non-Expert Operator}, i.e. a person from the audience with no robotics background.
-			In this case, the referee gives the goal of the command to the non-expert operator, who will then issue it to the robot in their own words. For example, the generated command might be ``Bring me a coke from the kitchen." then the non-expert operator will be told ``The robot should bring you a coke, which is found in the kitchen.", who then tells the robot ``I want a coke. Go to the kitchen and get me one."
-			If the robot consistently fails to understand the non-expert operator (e.g. after two retries), teams can default to a custom operator.
+			\item \textbf{Objects:} Objects that are not in their default location should be returned there. Objects on the floor are to be thrown in the trash.
+			\item \textbf{Persons:} Any person inside the arena other than the referee and referee assistant have a command for the robot and can be approached.
```
PetervDooren, line 26: What happens when the referee or assistant are approached for a task. Do they say, "no I do not have a task"?

tasks/EGPSR.tex
sunava: I believe assigning a number here encourages teams to experience a sense of fulfillment. The number 5 appears to be suitable.

tasks/EGPSR.tex
johaq: People with requests will raise their hands. All other people will say no if approached.

tasks/EGPSR.tex
johaq: I think I'll just leave it at 3 for now. If teams perform well I think increasing the difficulty of tasks should come first before increasing the amount.

LeroyR (approved):

scoresheets/EGPSR.tex
```diff
@@ -2,14 +2,11 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[3]{800}{Executing the task associated with each command}
+	\scoreitem[3]{800}{Find and solve a problem}
```
LeanderVonSeelstrang, line 5: I would like to split this up into 

\scoreitem[3]{250}{Find and communicate a problem}
\scoreitem[3]{550}{Solve a problem}

This would increase the partial scoring and allow the viewers to better understand the reasoning of the robot + generally what is going on. Reasoning for 250: 200 for the location + 50 for the problem class

LeanderVonSeelstrang (approved):
I really like the general idea. 

As you know you need to add the names.md, so people can test the command generator. As my comment was only a suggestion and if the file to generate commands is provided I think this should be merged.

```
$ python generator.py
Traceback (most recent call last):
  File "/home/leander/Dokumente/Robocup/CommandGenerator/generator.py", line 83, in <module>
    names_data = read_data(names_file_path)
  File "/home/leander/Dokumente/Robocup/CommandGenerator/generator.py", line 11, in read_data
    with open(file_path, 'r') as file:
FileNotFoundError: [Errno 2] No such file or directory: '../names/names.md'
```

scoresheets/EGPSR.tex
johaq: 250 seems a lot for just finding. I'll change it to 150/650.

scoresheets/EGPSR.tex
```diff
@@ -2,14 +2,12 @@
 
 \begin{scorelist}
 	\scoreheading{Main Goal}
-	\scoreitem[3]{800}{Executing the task associated with each command}
+	\scoreitem[3]{150}{Find and clearly state an encountered problem}
```
PetervDooren, line 5: One thing I foresee teams doing is approaching any and all people in the arena and asking them for tasks.
We could say a team wasting their own time is penalty enough.

tasks/EGPSR.tex
```diff
@@ -19,19 +14,12 @@ \subsection*{Focus}
 % Setup
 % %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \subsection*{Setup}
+Generally assume that the arena is supposed to be in its default state. Therefore, possible problems to solve are:
```
PetervDooren, line 17: The way egpsr worded is quite strange in my opinion. It is made very clear what the task is but not how this task fits in a household. Essentially, this is a combination of cleanup and gpsr. The robot has to maintain a clean house and act as a smart speaker.

PetervDooren (changes requested):
I've a few points. Mostly question and word choice.

scoresheets/EGPSR.tex
johaq: Yes, I think in this case that is penalty enough. Also if the robot repeatedly talks to people without tasks or maybe the same person even they can be removed according to 3.7.13 3. Inactivity

johaq: Maybe to elaborate. Solving this by giving a penalty for approaching people without a task discourages approaching people in general

tasks/EGPSR.tex
johaq: Yes, agreed. I updated the wording of the objective and the setup. I hope it reads better now

scoresheets/EGPSR.tex
PetervDooren: True, lets see how this plays out at a tournament

PetervDooren (approved):

LeroyR (approved):

tasks/EGPSR.tex
```diff
@@ -19,19 +14,12 @@ \subsection*{Focus}
...
-		\end{itemize}
-	\item \textbf{People:}
+    \item \textbf{Problems:}
 		\begin{itemize}
-			\item \textbf{Operators:} A \emph{Professional Operator} (the referee) commands the robot to execute a task.
-			Optionally, commands can be issued by a \emph{Non-Expert Operator}, i.e. a person from the audience with no robotics background.
-			In this case, the referee gives the goal of the command to the non-expert operator, who will then issue it to the robot in their own words. For example, the generated command might be ``Bring me a coke from the kitchen." then the non-expert operator will be told ``The robot should bring you a coke, which is found in the kitchen.", who then tells the robot ``I want a coke. Go to the kitchen and get me one."
-			If the robot consistently fails to understand the non-expert operator (e.g. after two retries), teams can default to a custom operator.
+			\item \textbf{Objects:} Objects that are not in their default location should be returned there. Objects on the floor are to be thrown in the trash or returned to their default location.
```
LeanderVonSeelstrang, line 21: Do we need to clarify how the default location is defined? I assume the location the object had during the mapping slot? Is a reasonable target location of any sort also okay? 

At the moment, I would assume that an object:location dictionary is sufficient to find the correct default location and that no further reasoning is required.

tasks/EGPSR.tex
```diff
@@ -53,44 +39,25 @@ \subsection*{Procedure}
...
+	\item Provide commands to volunteers.
 \end{itemize}
 
 \subsection*{OC instructions}
 
 2h before the test:
 \begin{itemize}
-	\item Generate the robot commands (don't reveal them to the teams!).
-	\item Announce the location of the instruction point.
+	\item Generate the problems and commands (don't reveal them to the teams!).
```
LeanderVonSeelstrang, line 57: Is there a reason to allow for 2h possible leak time? Only the main egpsr referee should know them so much in advance. Should be okay, though.

tasks/GPSR.tex
```diff
@@ -1,7 +1,6 @@
 \section{General Purpose Service Robot}
 \label{test:gpsr}
-Similar to a smart speaker, the robot is asked to execute arbitrary commands requested by an operator.\\
-\textbf{Note:} Due to this task's more open-ended nature, the maximum possible score is higher than the scores of the other \SONE{} tasks.\\
+The robot is asked to understand and execute commands requiring a wide range of different abilities. \\
```
LeanderVonSeelstrang, line 3: Maybe "natural language commands" to clarify, that the commands won't follow a well defined fixed immutable structure.

LeanderVonSeelstrang (approved):
I guess I would clarify the "default location" and explicitly set the numbers of persons who know the generated commands in advance to exactly one person. Personally I think the commands should be phrased in natural language and should not have to follow a fixed pattern. 

Idk consider to change these things, but I would be okay with a merge with the current version, too.

tasks/EGPSR.tex
johaq: Default location is defined in 3.3.5, added reference to it. Yes that is all that is required for now. Let's first see how teams perform before we make it more difficult. Also "reasonable" target location is debatable

tasks/EGPSR.tex
johaq: Honestly, these things are more of a reminder to set the referee up with everything. In this case make sure each referee has the means to generate commands inside the arena.

tasks/GPSR.tex
johaq: No, GPSR commands for now. If teams score, make it more difficult

PetervDooren (dismissed):

sunava (dismissed):

LeroyR (approved):

PetervDooren (approved):

PR #850, opened 2024-03-06
Fix missing refs and catch them in CI

LeroyR:
Sorry, this is a big one.

### fix missing refs in the rulebook
- added bonus/penalties section to rulebook (for social survery)
- removed scoresheet refs to `start_signal` for now as we would need to add the section which refers `inspection` which refers `robot` - dont know what to move where so i removed the ref from the scoresheet for now

### Cleanup build
- build changed to using latexmk with -Werror on  texlife/texlife2022 image, to catch missing refs during ci
  - you can test the ci build locally with `latexdockercmd.sh`

### Add missing roadmap document


MatthijsBurgh:
I don't think this should be one PR. Please split it in multiple isolated PRs.

I.E.:
- Changing the CI
- Cleaning up the makefiles
- Checking for the missing refs
- Fixing the missing refs
- Adding the roadmap document to CI
- Actual rulebook changes

MatthijsBurgh:
> I could do one pullrequest for each commit in this branch tho

Not the perfect solution, but much better than one big PR. So please do that.

LeroyR:
One pullreq for each commit

Issue #841, opened 2024-01-08
New GPSR Command Generator


johaq:
I propose a new command generator for GPSR found inside my proposal for a template competition repo found here #840.

Mainly recreates old generator in python but reads in information from competition resources directly.

Also includes some changes based on feedback from previous years mainly removing unclear commands, and removing mentions of gender.

Can separate between commands including object manipulation and perception or commands only using hri and people perception. Generated batches of three commands are now more balanced based on that. 

This is a very early draft. I will be fairly busy the coming weeks so I need help in finding all the little errors.
Any help is appreciated.

Edit:
Also need to add QR code generation

MatthijsBurgh:
Maybe we should make it a submodule. So we don't copy the code each year.

johaq:
Yes, agreed. Very good suggestion.

jws-1:
Hello,

I was wondering if you are able to provide some example markdown files for the new command generator, if possible?

Thanks,

Jared

johaq:
There are files in the competition template repo: https://github.com/johaq/CompetitionTemplate
Which example files are missing?

jws-1:
Ah, I hadn't seen that. Thanks!

m-barker:
@johaq I believe the Question/Answer files are still missing. I.e., [this example](https://github.com/RoboCupAtHome/Bordeaux2023/blob/master/cmdgen/Questions.xml) but for the new format.

johaq:
Ah yes, I was thinking whether it makes sense to have dynamic questions based on the setup like:
"In which room is the TV?"
"What is the biggest fruit?" etc

I will work on this.

LeroyR:
@johaq Want to you transfer the ownership to RoboCupAtHome org?

Issue #84, opened 2015-04-26
Create BIG posters for challenges to appeal audience


LoyVanBeek:
To advertise the league and challenges, make posters for each challenge. 
In English and Chinese, with the time, a picture of the actions performed.

kyordhel:
Great idea. Questions as follows

- Who is going to do posters?
- Who is going to pay for the posters' printing?

LoyVanBeek:
I don't know, @robotijn asked me to :smiley: But I do think this is a good idea, for every league. 
I'll ask the LOC.

LoyVanBeek:
E-mail sent...

PR #856, opened 2024-04-05
Receptionist typo and remove references to age and gender

jws-1:
Small PR fixing a typo in the `Receptionist` task, and removing reference to age and gender (as has been done in GPSR, see #841).

### DIFF
--- a/tasks/Receptionist.tex
+++ b/tasks/Receptionist.tex
@@ -44,7 +44,7 @@ The robot has to take two new guests to the living room to introduce them and of
 	
 	\item \textbf{Switching Places:} Guests may switch places after they were seated.
 	
-	\item \textbf{Describing the First Guest:} Naming at least 4 characteristics of the first guest, i.e., color of clothes, color of hair, gender, and age, earns bonus points.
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
I don't think this is related to #84...

jws-1:
> I don't think this is related to #84...

Oops, I meant #841. I'll update.

MatthijsBurgh:
I don't really agree that #841 shows the removal of any age/gender reference.

I also think it is too big of a change to eliminate the robot making any age/gender statements, while the rulebook is already released.

(P.S. in my opinion it is fine for the robot to estimate age/gender of a person. Even for next year)

MatthijsBurgh:
Ok, I missed that one in #841 . Thanks for highlighting it.

My other argument is still there. I think this is too big of chance after the rulebook has been released.

LeroyR:
I also think this is too big of chance after the rulebook has been released, if this means the robot is not allowed to state gender or age of the person as teams may already implemented such behavior.
 
I also don't see any characteristics that could not be considered offensive, what if the person got coerced into wearing the specific outfit the person could be balding, height could be offensive too for a little person etc.

Feel free to reopen the typo fix.

PR #851, opened 2024-03-06
[Draft] organization doc restructure

LeroyR:
Rulebook release has a bunch of missing refs due to the rulebook splitup.

310188234-9d4f18fe-f1b6-4204-8e43-0ef1da19bfe5

This fixes that, but:

- readded bonus/penalties section to rulebook (for scoresheet refs)
- removed some ref for now as we need to add `start_signa`l but that refers` inspection` which refers `robot` sections. I dont know what to move where so removed the refs for now

PR #857, opened 2024-04-07
fix missing refs Alternative

LeroyR:
Alternative to #851 without restructuring as much.


LeroyR:
done fighting with this borked ci setup

Issue #855, opened 2024-04-05
Duplicate scoring entry in Storing Groceries


Shimarisu-kun:
The following scoring item is duplicated with different point deductions in Scoring Groceries:

"A human placing an object in the cabinet at a location clearly indicated by the robot"

The issue seems to have been introduced with commit 807610c.

MatthijsBurgh:
You are correct, see:
https://github.com/RoboCupAtHome/RuleBook/blob/5a3961f153ad5448bae4bbd7c18648cbfba7c054/scoresheets/StoringGroceries.tex#L24-L26

I leave it up to the current TC members, to fix this.

PR #860, opened 2024-04-08
Update StoringGroceries.tex

PetervDooren:

## Description

Closes issue #855 

-
-
-

## Other comments
Has overlap with  #859. Difference is this PR only removes the duplicate (taking the original penalty) without adjusting the other scores.


LeroyR:
Problem is -90 just does not make sense:
- Placing an object in the cabinet +15
- Placing an object next to similar objects on the cabinet +50

So it is better for the robot to just drop it on the floor

MatthijsBurgh:
I agree with @LeroyR the -90/-30 doesn't make sense compared to the points gained. I think does needs to be adjusted too.

But I agree that this PR is cleaner than #859

PetervDooren:
I also agree that -90 is a lot. It is also not made clear if this placing is from the robot to the cabinet or from the table to the cabinet. I am not saying this is not an issue. But it is not the issue being adressed in this PR.
Only question is whether the penalty for having a human place an object should be -30 or -45.

PR #831, opened 2023-10-31
Split rulebook

PetervDooren:
** Note: Your contribution is expected to meet the conventions and policies described in the [contribution guidelines](https://github.com/RoboCupAtHome/RuleBook/wiki/Guidelines:-Contributing) **
Made to replace #829 

## Description
Splits rulebook into the rules and organization of the competition. This makes for a briefer rulebook and hopefully a better intro for new teams.

Closes issue #

Changes proposed in this pull request:
- split rules and organization into two documents
-
-

## Other comments
If this request is marked as a draft, please describe your plans or any specific feedback you would like

Todo:

- [x] Edit the documents to be legible on their own.

PR #861, opened 2024-04-08
Add roadmap document

LeroyR:
There was an additional roadmap document introduced in #831.

This just adds it to ci. I did not look at the document itself.


MatthijsBurgh (changes requested):
Please keep the list of (organization, roadmap, rulebook, scoresheets) alphabetically.  I think this applies to all spots you made changes.

moriarty (approved):
SRTM. I'm not concerned with the alphabetical lists if they're not caught by a precommit hook or linter then the ordering is kinda arbitrary and hard to enforce.

LeroyR:
included with #905

PR #863, opened 2024-04-23
German Open Feedback for Receptionist

johaq:
- Changed scoring of guiding person to party area (this is just navigation unless the person gets lost. If the person gets lost that takes up a lot of time and not really part of the task description)
- Clarified that the host is also a guest for introduction
- Allow partial scoring for giving descriptions of people (per attribute)
- Allow bell ring to be a door knock as well

### DIFF
--- a/scoresheets/Receptionist.tex
+++ b/scoresheets/Receptionist.tex
@@ -2,10 +2,9 @@ The maximum time for this test is 5 minutes.
 
 \begin{scorelist}[startbutton=false]
 	\scoreheading{Main Goal}
-	\scoreitem[2]{15}{Navigate to the door, when the door bell rings}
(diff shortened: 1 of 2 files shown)


LeroyR (dismissed):

hawkina (approved):

LeroyR (approved):

akinobu1998:
Do we get minus points if the robot tells only incorrect attributes (and how do we judge it)?

johaq:
> Do we get minus points if the robot tells only incorrect attributes (and how do we judge it)?

No minus points. I would judge it based on how I would describe a person. 
For example an age estimate does not need to be 100% correct but in a reasonable range. So saying around 30 years old is fine. Saying between 25 and 50 is too broad imo

akinobu1998 (approved):

PR #864, opened 2024-04-23
German Open Feedback for Carry my Luggage

johaq:
- Adding penalty for rediscovering the operator by asking them to come back
- Clarify the DEM penalties are only applied if the robot scores afterwards

### DIFF
--- a/tasks/CarryMyLuggage.tex
+++ b/tasks/CarryMyLuggage.tex
@@ -59,8 +59,10 @@ The robot helps the operator to carry some luggage to a car which is parked outs
     \begin{enumerate}
      \item Natural interaction (e.g., waving and calling)
      \item Unnatural interaction (e.g., raising both hands and jumping)
+	 \item Coming back (e.g., asking the operator to step back in front of the robot)
      \item Touching the robot (e.g., pulling the robot's hand)
     \end{enumerate}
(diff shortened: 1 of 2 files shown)


tasks/CarryMyLuggage.tex
```diff
@@ -59,8 +59,10 @@ \subsection*{Additional Rules and Remarks}
     \begin{enumerate}
      \item Natural interaction (e.g., waving and calling)
      \item Unnatural interaction (e.g., raising both hands and jumping)
+	 \item Coming back (e.g., asking the operator to step back in front of the robot)
      \item Touching the robot (e.g., pulling the robot's hand)
     \end{enumerate}
+	Note: Penalties for human assistance are only applied if the robot continues to score after the interaction.
```
LeroyR, line 65: Can the penalties be applied multiple times? How do different DEM penalties stack - (e.g. up to XXXpts penalty) We should specify that.

Id suggest something like " Penalties can be applied multiple times upto a total of 250points" 
As the latest results suggest the task is still pretty hard

LeroyR (changes requested):

LeroyR (approved):

akinobu1998 (approved):

hawkina (approved):

PR #866, opened 2024-04-23
German Open Feedback for Restaurant

johaq:
- Adjusted DEM penalties to fit scoring item Serve the order
- Clarify usage of unattached trays
- Clarified mapping procedure (Note: I think allowing teams to start their mapping component at the front of the queue is good because I want to save as much time as possible at the start)
- Tweaking/coding before the test to fit test setup is always forbidden, removed here for clarification.

### DIFF
--- a/tasks/Restaurant.tex
+++ b/tasks/Restaurant.tex
@@ -77,7 +77,7 @@ This task focuses on
 		\item If a person from the audience (severely) interferes with the robot in a way that makes it impossible to solve the task, the teams may repeat the test immediately.
 		\item Each Deus Ex Machina penalty for skipping manipulation will only be applied twice per order so receiving an order with three objects is not more punishing.
 		\item If the robot detects a customer but does not reach their table, the robot must clearly show who was detected to receive points, i.e. displaying a picture of the person.
-		\item When a team is at the front of the queue, they are allowed to start their mapping procedure (the robot must remain in place). When it is their turn, they must bring the robot directly and in a straight line from the front of the queue to the start location. Once at the start location only slight movements are allowed (no moving back and forth, no full rotations etc.)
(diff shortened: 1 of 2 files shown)


LeroyR (approved):

LeroyR:
As far as i can tell most teams do some kind of debugging/coding on their start position hard to tell if that is general (e.g. fix the robot not starting, wrong behavior transitions) or specific (e.g. where the tables in the restaurant are located)

johaq:
Yes, that is why I removed this section. But coding in people locations etc before the test is obviously still forbidden but already covered in general rules

hawkina (dismissed):

tasks/Restaurant.tex
```diff
@@ -71,16 +71,18 @@ \subsection*{Additional rules and remarks}
...
+		\item If requested, the barman will place the order in a basket or tray for the robot to deliver it.
 		\item Upon arrival to the restaurant, only the team leader is allowed next to the robot for watching and charging.
 		\item If a person from the audience (severely) interferes with the robot in a way that makes it impossible to solve the task, the teams may repeat the test immediately.
 		\item Each Deus Ex Machina penalty for skipping manipulation will only be applied twice per order so receiving an order with three objects is not more punishing.
 		\item If the robot detects a customer but does not reach their table, the robot must clearly show who was detected to receive points, i.e. displaying a picture of the person.
+		\item When a team is at the front of the queue, they are allowed to start their mapping procedure (the robot must remain in place). When it is their turn, they must bring the robot directly and in a straight line from the front of the queue to the start location. Once at the start location only slight movements are allowed (no moving back and forth, no full rotations etc.)
```
akinobu1998, line 80: What if we say startup procedure or something?
This might mislead the team to create a map of the arena when they are waiting in the queue.

akinobu1998: Preparing in advance and starting the robot at the front of the queue is a good idea (no 30-second preparation is necessary anymore if they can prepare in advance).

> because I want to save as much time as possible at the start

Some teams might need to adjust the camera, mic, and software so we do not need to limit the mapping I think.

akinobu1998 (approved):

LeroyR (approved):

hawkina (approved):

PR #873, opened 2024-05-13
Portugal Open Feedback for Restaurant

johaq:
- Allow two team members next to robot when waiting in line (I see no reason why it HAS to be the team leader)
- Added and adjusted penalties:
    - Not making eye contact adjusted to -80. There are three parts to Take an order (300). Detect and get to a guest, understand the order, and make it back to the barman and report the order. So I think not making eye contact affects a third (understand an order) so should not give a full deduction of 100.
    - Added penalty for not reaching the bar. Can easily be decided by whether the referee has to move from behind the bar to interact.
    - ~~Added penalty for using custom operator since the default penalty (10% so 30p) is not high enough for this test.~~ (Note: this was a misreading on my part. Rule says 10% of maximum attainable score, so would be 205p)


akinobu1998 (approved):

LeroyR (approved):

MatthijsBurgh:
I think this is too close to the tournament to merge, especially the scoring. Two team members is fine

moriarty (approved):

hawkina (approved):

PR #902, opened 2024-09-09
Cleanup

LeroyR:
We currently have some unused tasks in the `tests` as well as the` tasks` folder. 
The repository layout could be confusing for .. anyone.

This moves all unused tasks into `tasks/old` for lookup.  They should just be deleted (we use git after all) but for _convenience_ i just moved the files for now. 

Also, the other commit is adding build instructions.


MatthijsBurgh:
We have version control, so we don't need an old folder. Just delete them. If you want stuff to be easy to find, tag the commit before deleting stuff.

akinobu1998 (approved):

tests/GPSR.tex
moriarty, line 1: wow that is confusing that this was in both tests and tasks.

moriarty (approved):

hawkina (approved):

Issue #900, opened 2024-09-02
Review Organization Document


LeroyR:
## Describe the bug
The Rulebook got split into rulesbook and organization document to reduce the size of the rulebook but currently there are some rule related information in the orga document. 
Also orga and rule doc depend on each other.

We should either undo the split or cleanup the current mess e.g. rule document should not depend on the orga and vice versa.

johaq:
Yes, we should move everything that the rulebook references in the organization document back to the rulebook document

PetervDooren:
As the person who made this mess I would recommend to merge them back into one document

PR #905, opened 2024-09-10
Restructure orga doc

LeroyR:
## Description

Alternative to #917 

Closes issue #900

We failed to correctly split the rulebook into two documents last year. 
See https://github.com/RoboCupAtHome/RuleBook/pull/851 for more info

This is looking into cleaning up the split. 

## Other comments

I think there are only a limited number of sections that should not be in the rulebook and can be exclusive to a different document as they have to influence on the **rulings**:

1. Introduction: Committee, Websites, Leagues
2. Team Registration and Qualification process
3. (Maybe) Setup and Preparation

Also there is a general rules section as there are some references to **RULES** e.g.  (e.g. external device, robots, for registration process) as i dont want to introduce any more cross references into the other document


johaq:
Regrading Setup and Preparation, at least Poster Session and Robot Inspection needs to be in the rules document.

sunava:
This branch has conflicts that must be resolved

Issue #912, opened 2024-10-11
Combine Storing Groceries and Clean the Table into one Stage 2 Task 


SparkRibeiro21:
Note: This comes as an alternative to combining Serve Breakfast and Clean the Table (so some comments may reference this comparison)

Combining SG with CT would provide a more natural task. In CT, the only "furniture interaction" is with the dishwasher. Whereas in a standard dinners there are more than dishes to tidy up. There are drinks, foods, fruits, snacks, trash, etc...
The idea of combining these two tasks, is to have a wider variety of objects on the table, such as foods or drinks for example that the robot would have to (similarly to SG) place in the correct cabinet shelf. For example: one object from two different categories (one fruit and one snack) are also on the table, and have to be placed near the same category in the cabinet. This would provide a more realistic "end of dinner" scenario rather than just having dishes on the table to tidy up. A bonus, could be having an object from a third category (like a drink) that had to be placed in the trash bin. Fully representing what happens in an actual dinner scenario: objects that have to be cleaned, objects that need to be stored and objects that need to be put in the trash.
This way SG would be naturally and smoothly merged into CT not being just removed from the tasks the robot should do, as what is done in SG will also be done in the new CT.
In addition (comparing to combining SB with CT proposal) it would not cause additional mess for referees that would have to change objects on the table.
This would allow teams to have "simpler" chores (placing objects in the cabinet) in CT. Since CT task has the lowest average point of every S2 task, with only 4 teams out of 17 (all S2 teams from the 3 sub leagues) having more than 55 points (which is a LOT LOWER than all other tasks). This would most likely increase the amount of chores robots can perform in CT, making it more interesting for audience.

Also putting SB as a second part of CT feels unnatural and since teams struggle so much do CT will virtually make SB disappear. 
SB is the only task where the robot must perform an action (pouring) that is not just pick and place of an object is done. In addition interactions with the cabinet are also being performed in EGPSR and Stickler (and a bit in GPSR) whereas performing actions such as pouring do not exist in any other task. Actually teams have been using it in finals performances to showcase their manipulation which commends actions such as pouring that are not just pick and place.
SB provides more show-off ability than SG, this can be observed from the livestream and from teams videos. The audience is more invested in robots pouring items into the bowl than placing items in a cabinet. Combining SG with CT would mean more show-off in Stage 1 but also in Stage 2 as teams, apart from loading the dishwasher can also perform more manipulation chores. 

[...]

PR #913, opened 2024-10-15
Combining Storing Groceries and Clean the Table into a Stage 2 task

SparkRibeiro21:
As described in https://github.com/RoboCupAtHome/RuleBook/issues/912 this is presented as an initial draft to combine the Clean the Table with Storing Groceries. Resulting into an updated Clean the Table where the robot also has to clean up remaining foods, drinks, fruits, ... left on the table in the cabinet (similarly to SG). I have also added a bonus task where a class of objects is announced as trash (for example: drinks) where the robot must place this object in the trash bin. Making this a much more realistic post-meal scenario where you have dishes to be cleaned, food to be stored and trash to be placed in a bin.

There are still some pending matters that I think new opinions may help with the overall task flow:

1. How many objects should the robot have to store?
 I have added four food objects to be cleaned up (three to be stored on the cabinet and one in the trash). But maybe this is too much or too little. Personally I was indecisive between three or four food objects.  

2. The points adjustment according to difficulty have not been updated yet.
The goal of this pull request is just to figure out the combination of SG and CT. Point adjustments still have to be made in a later pull request.

3. The amount of time HAS to be increased, but not sure of what it should be. 

4. Standardise the points received for picking and placing of objects.


moriarty:
@johaq brought this up and it was discussed in EC Meeting.

@mleonetti & @moriarty & @ARTenshi @myszzka96 have agreed

sunava (dismissed):

LeroyR:
> @johaq brought this up and it was discussed in EC Meeting.
> 
> @mleonetti & @moriarty & @ARTenshi @myszzka96 have agreed

Agree to which solution exactly? :smile:  I think we unanimously voted that Storing Groceries should not be our Stage 1 Task in its current form.

documents/rulebook.tex
```diff
@@ -73,7 +73,6 @@ \chapter{Tests in Stage I}
 \testtocentry{test:gpsr}
 \testtocentry{test:receptionist}
 \testtocentry{test:serve-breakfast}
-\testtocentry{test:storing-groceries}
```
LeroyR, line 76: Just add the combined task and don't remove storing groceries for now.

LeroyR (changes requested):

documents/rulebook.tex
SparkRibeiro21: Fixed

SparkRibeiro21:
**For reviewers:** Please read the description and tell me how many objects should the robot have ti store in both cabinet and trash bin!

_In Description: "How many objects should the robot have to store?
I have added four food objects to be cleaned up (three to be stored on the cabinet and one in the trash). But maybe this is too much or too little. Personally I was indecisive between three or four food objects."_

hawkina:
The time could be increased to 10 minutes? Maybe 7:30 might be enough.
Amount of objects: Let's say we want to focus on a realistic scenario. We could have: bowl/plate  + one cutlery item, then 3 food items like cereals, a fruit and a snack/drink so that we can declare of them to be thrown into the trash bin. This would mean 5 items per pick-place action. 
We could consider adding more items as a bonus and say, that for successful task completion we expect 5 out of all 3 categories,  (categories being: dishes, cutlery, food) but if someone manages to do more, they get an outstanding performance. 

It is a bit difficult to balance the necessary amount of items vs. have a nice scenario which makes sense.

ARTenshi:
> How many objects should the robot have to store?
> The amount of time HAS to be increased, but not sure of what it should be.

Given the current performance of service robots, I would say that ten objects in 10 minutes is fine.

> I have also added a bonus task where a class of objects is announced as trash.

Yes. I would take drinks and food items to the trash and some sides (ketchup, mustard, napkins) to the shelf.

LeroyR:
I am a fan of setting more stuff on the table than the Robots can handle but we have to make sure that a robot only able to cleanup does not score 5k points 😅

We could set multiple seats and  decrease the score for doing the same thing repeatedly, score times 0.5 for each repeat? 
(1st trash gives 100, 2nd gives 50, and then 25 for each additional tash item. 
Maybe a bonus for a each first item of 'category' or for cleaning up one seat completely. 

We can tweak the numbers and individual scores later. For now we just need to make sure that the task and every possible scoring  action is written down.

Issue #888, opened 2024-08-19
[DSPL] allowing external Microphones


hawkina:
**Issue**
This year it was very sponteniously decided during the setup days, to allow teams to use external microphones on the Toyota HSR. This rule was implemented as follows: 
- people are allowed to add an external microphone to the HSR via USB/AUX
- It can be mounted on the robot
- It can be a wireless microphone which can be handed to the referee
- It is allowed to keep using the build-in HSR microphone - this gave 15 extra points - correct me if I am wrong

**Points for discussion:** 
- I think we can all agree that keeping this rule makes sense moving forward since the HSR otherwise suffers a lot in the competition and teams otherwise have not much that they can do about it.
- Should using the build in mic give more bonus points since it is a huge difference and teams which manage to use it well consistently deserve these?
- are wireless microphones still allowed? These increase the audio quality very much, but if this is allowed in DSPL it should also be allowed in OPL.
- should there be a point deduction / penalty for using wireless microphones if they are handed to the referee/volunteer?

ARTenshi:
Maybe, towards moving the HSR to OPL league in the next couple of years, we should allow any modification next year for the DSPL as a bridge towards this goal.

hawkina:
> Maybe, towards moving the HSR to OPL league in the next couple of years, we should allow any modification next year for the DSPL as a bridge towards this goal.

Attaching something simple as a mic is one thing, modifying and taking apart the robot a whole other. I understand wanting to move towards OPL in the long term but for now this would defeat the entire purpose of even having DSPL. I think we can make it clear that the future would be a merge between DSPL and OPL but this should be a gradual process and we should probably announce it a few years in advance. I mean, whoever has the resources, can already do this and just participate in OPL instead.
Allowing all modifications is too big of a change with just one year warning time.

johaq:
So, I am against using wireless mics. They also don't seem to be necessary judging from team performances especially when allowing an external mic.

Wrt allowing general modifications: I am against allowing this generally. But I am certainly open to hearing more suggestions for possible modifications. But for now I think all of those modifications need to be in the rulebook and not a general permission.

sunava:
[...]

PR #919, opened 2024-11-25
allowing hsr microphone

sunava:
## A new subsection to the External Devices section that formalizes the use of external microphones on the Toyota HSR has been added.

Specifies the conditions under which external microphones can be used, including USB/AUX connections, wireless options, and the use of the built-in microphone.
Closes issue # [888](https://github.com/RoboCupAtHome/RuleBook/issues/888)


LeroyR (approved):

sunava:
> Do we have (do we need?) any macro or special layout to mark rules as HSR only?

i think for the start signal @johaq  wanted the rule to be hsr specific, microphone idk

LeroyR (changes requested):
Move to some HSR section.
Also this is not an external device :D

sunava:
> Move to some HSR section. Also this is not an external device :D

is it not :D upsi

PR #920, opened 2024-11-25
added microphone for hsr

sunava:
## A new subsection to the External Devices section that formalizes the use of external microphones on the Toyota HSR has been added.

Specifies the conditions under which external microphones can be used, including USB/AUX connections, wireless options, and the use of the built-in microphone.
Closes issue # [888](https://github.com/RoboCupAtHome/RuleBook/issues/888)

### DIFF
--- a/pages/general_rules/Robots-SPL.tex
+++ b/pages/general_rules/Robots-SPL.tex
@@ -16,6 +16,7 @@ Any laptop fitting inside the \MountingBracket{} is allowed to be used, regardle
 Furthermore, teams are allowed to attach the following devices to either the \HSR{} or the laptop in the \MountingBracket:
 \begin{itemize}
 	\item \textbf{Audio}: A USB audio output device, such as a USB speaker or a sound card dongle.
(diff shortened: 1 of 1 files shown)


LeroyR (approved):

johaq:
Are there potential issues with having a wireless microphone attached on the robot? Otherwise I think that's fine and the rule should just state that the mic is attached to the robot.

akinobu1998:
The connection method is not important I think.
Is the microphone attached permanently or can it be removed by human operators if the robot says to do so? We need to discuss more.

SparkRibeiro21 (approved):

hawkina:
> The connection method is not important I think. Is the microphone attached permanently or can it be removed by human operators if the robot says to do so? We need to discuss more.

It should be attached permanently. Otherwise we can go back to allowing "handing the micophone to the referee", and if I remember correctly, people were against this in prevous discussions. 

Wasn't the reason for not allowing wireless micophones potential cheatng? One usb receiver can connect to multiple microphones at the same time, so in theory, people could repeat the command to the robot, worded differently or directly into the microphone mproving the sgnal quality or something?

hawkina (approved):

PR #923, opened 2024-11-25
Storing groceries changes

LeroyR:
## Description

Storing Groceries changes to move the league forward. We want robots to start using the door handles and pick items directly from the shopping bag.
Also this incorporates pouring from the removed serving breakfast task by refilling a dry storage container.

- Increase time limit to 7 minutes.
- Opening door is now a main goal. 
  - Additionally we close both doors to have a more realistic scenario.
- Remove the team dependent setup (request door open, objects) instead the robot should just handle the situation.
  - Select the non heavy object if grasping is impossible for the robot
  - Requesting assistance for door opening
- Incorporate pouring by refilling a dry storage container with cornflakes.
  - Container is placed on the table for now. 
- Grasping objects from a shopping bag
  - placed next to table.


johaq:
I like this idea since it preserves two kinds of placing challenges, placing on a table AND placing inside a shelf. My questions is should we increase the time limit for this test? Potentially 7:30 instead of 5:00.

scoresheets/StoringGroceries.tex
```diff
@@ -1,24 +1,31 @@
...
 	\scoreitem[5]{15}{Perceiving object and categorizing it correctly}
 	\scoreitem[5]{50}{Picking up an object for transportation to the cabinet}
 	\scoreitem[5]{15}{Perceiving objects in shelf and saying on which layer the currently handled object should be placed}
 	\scoreitem[5]{15}{Placing an object in the cabinet}
 	\scoreitem[5]{50}{Placing an object next to similar objects on the cabinet}
+	\scoreitem{300}{Pouring cereal into the container}
 
 	\scoreheading{Bonus Rewards}
 	\scoreitem{200}{Opening the cabinet door without human help}
+	\scoreitem{50}{Picking up an object from the shopping bag}
```
hawkina, line 13: I'd suggest to also give 70 points here since it is more difficult. It is comparable to picking up the tiny/heavy objects.

tasks/StoringGroceries.tex
```diff
@@ -52,9 +56,11 @@ \subsection*{Procedure}
 		\begin{itemize}
 			\item whether the cabinet door should be closed
 			\item whether a heavy and/or a tiny object should be used in the test
+			\item whether the robot plans to pour the milk
```
hawkina, line 59: what about cereal?

hawkina (commented):
Generally I like the changes and the combination.
One question: is the container for the cornflakes going to be transparent? If so, then it might add an additional difficulty for perception and we might consider rewarding it/ giving alternative options. Or just let people know in advance so that they know to train for transperency.

hawkina:
> I like this idea since it preserves two kinds of placing challenges, placing on a table AND placing inside a shelf. My questions is should we increase the time limit for this test? Potentially 7:30 instead of 5:00.

I would vote for increasing the time limit to 7:30 mins.

scoresheets/StoringGroceries.tex
LeroyR: I would change the overall scores later. For now this is just to get the different score items into the rulebook

tasks/StoringGroceries.tex
LeroyR: As this is a main goal the cereal should be always present

LeroyR:
This is just a draft to show possible integrations. Details like scoring, container (where it is placed: inside cabinet so the robot has to take it out first or on the table) and time limit need further discussion.

---

> I like this idea since it preserves two kinds of placing challenges, placing on a table AND placing inside a shelf. My questions is should we increase the time limit for this test? Potentially 7:30 instead of 5:00.

I put it to 6:00 in the initial draft to accommodate pouring the cereals. Going up to 7:30 is also fine.

I'd also like to introduce either a bonus for the first item or a decreasing score for each additional item - f the robot stores two items it should be able to store more anyway with time being the limiting factor. The space inside the cabinet was not really limited (At least until we require storing items efficiently e.g. by stacking cans)

So my question is: Are 6 minutes not enough for a ~slow robot to get to the location, scan the shelf, pour cornflakes and store one item (add potentially opening the door as main goal). Faster robots can then store multiple objects to collect additional points.

Another open question: Do we still want this kind of opening the door?

sunava:
Here's a rewritten version for the PR comment:

- **Johannes**: Suggests adding extra time since very few teams managed to pick up all 5 objects.  
- **Luis & Vanessa**: Propose keeping the "storing groceries" task as it is, possibly adding time, and including the "pouring" task in Stage 2.  
- **Alina**: Supports the idea but prefers to focus on keeping "pouring" included due to limited available time.  
- **Akinobu**: Agrees with adding extra time and keeping "pouring" in Stage 2.

So, closing this for now.

LeroyR:
There is still ongoing discussion in this **DRAFT**, i dont see rationale for closing (just rename to storing grocerie changes)

Summary until now:
- My Suggestion to add extra task (pour, bag) and increase timelimit accordingly
- Multiple Suggestion to add extra time, even though we have a team finishing everything in 3 minutes
- No mention of opening door or using bag
- One vote for add pouring here
- One vote to add pouting to stage 2, even though we decided to **NOT** combine breakfast with cleanup

PR #924, opened 2024-11-30
[Draft] Alternative HRI task to Stage 2: Help Me Find.

SparkRibeiro21:
## Description
This is a rough first attempt at the "Help Me Find" task, that consists of helping a visually impaired person find navigate and pick an object the this person is trying to collect. To goal is to force robots to adapt to a person that has limited information regarding its environment, and therefore must develop a good perception of the person. The volunteer should be wearing a blindfold make this scenario more realistic and forcing the person to trust the information received from the robot. In addition, the robot must be able to describe objects not present in the competition dataset and understand object descriptions.

## Other comments
This is just a rough description... points and even task description still need a lot of thinking and adjusting, please provide feedback...

## Notes
This PR is made after the last TC meeting where this proposal has a lot of positive feedback. This PR is also to help the next EC meeting.

### DIFF
--- a/documents/rulebook.tex
+++ b/documents/rulebook.tex
@@ -101,6 +101,7 @@
 \testtocentry{test:egpsr}
 \testtocentry{test:restaurant}
 \testtocentry{test:stickler-for-the-rules}
+\testtocentry{test:help-me-find}
 
 
 % Uncomment to get aesthetic improvement
@@ -112,6 +113,8 @@
 \input{tasks/Restaurant}
 \newpage
 \input{tasks/SticklerForRules}
+\newpage
+\input{tasks/HelpMeFind}
 
 
 
--- a/scoresheets/HelpMeFind.tex
+++ b/scoresheets/HelpMeFind.tex
@@ -1,0 +1,27 @@
+\begin{scorelist}[timelimit=10]
+
+	\scoreheading{Regular Rewards}	
+	\scoreitem{50}{Correctly understanding where the objects are.}
+	\scoreitem{600}{Help the person safely navigate to the objects location.}
+	\scoreitem{200}{Correctly identify the requested object.}
+	\scoreitem{600}{Help the person pick the correct object.}
+	
+	\scoreheading{Bonus Rewards}
+	\scoreitem[10]{30}{Provide a thorough description of the objects.}
+	\scoreitem{300}{Help the person safely navigate to the exit.}
+
+	\scoreheading{Regular Penalties}
+	\penaltyitem{100}{Person bumping into something while being helped to navigate}
+	\penaltyitem{100}{Person picking the incorrect object}
+
+	% \scoreheading{Deus ex Machina Penalties}
+	% \penaltyitem[4]{50}{A human directs the robot towards a guest who is breaking a rule}
+	% \penaltyitem[4]{100}{A human tells the robot which rule is being broken}
+
+\end{scorelist}
+
+
+
+% Local Variables:
+% TeX-master: "Rulebook"
+% End:
--- a/tasks/HelpMeFind.tex
+++ b/tasks/HelpMeFind.tex
@@ -2,7 +2,7 @@
 \label{test:help-me-find}
 
 \subsection*{Description}
-A visually impaired person must find an object inside the house. However, this person has never been to this house. The robot must help the person navigate through the house and find its object.
+A visually impaired person must find and pick up an object inside the house. However, this person has never been to this house. The robot must help the person navigate through the house safely and find its object.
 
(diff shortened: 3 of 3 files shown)


hawkina:
The robot is expected to describe all the objects in can see on e.g. the table? Do we ensure that there are not too many obejcts? Otherwise the robot might spend too much time describing things. 

One cool solution would be ofcourse to list all objects it sees, and the user can then say which should be described.

ARTenshi:
I still don't like the story; visually impaired people have been able to move without external help, especially in their own houses. I might change it to be them guests, so the robot takes them from the front entrance to the couch and then describes the drinks and snacks on the closest table to offer one. 

In any case, regarding object description, I think there is very little HRI; the robot might just describe the objects it sees using a dictionary -- the base case would be the robot taking any object and giving it to the person, without any interaction. Gesture recognition and visual or physical interaction are desirable -- human-robot collaboration would be a plus.

LeroyR:
> I still don't like the story; visually impaired people have been able to move without external help, especially in their own houses. I might change it to be them guests, so the robot takes them from the front entrance to the couch and then describes the drinks and snacks on the closest table to offer some.

Currently it talks about visually impaired being at the house for the first time. I think we should change the story to simply be about someone wearing a blindfold for the purpose of whatever happens in the task.


> 
> In any case, regarding object description, I think there is very little HRI; the robot might just describe the objects it sees using a dictionary -- the base case would be the robot taking any object and giving it to the person, without any other interaction. Gesture recognition and visual or physical interaction are desirable -- human-robot collaboration would be a plus.
If the robot can guide me, blindfolded, to take the correct item is has to either instruct me carefully (where to place my hand) or drop it into my extended open hand, both sound like a good solution.

For Dialog: **I still have the open question of what exactly anyone considers good HRI (that we can _somehow_ measure.)** 

In what situations do we have a _real_ dialog between robot and human? The only thing that comes to MY mind is somehow making a [Taboo](https://en.wikipedia.org/wiki/Taboo_(game)) like game out of it to force robot/human to ask/answer multiple questions to get to the desired item. For the Objects we then give a description of the desired features to the blindfolded human so that saying "there is an apple" is not enough. :sweat_smile:

hawkina:
Somehow this challenge reminds me of a claw-machine.

> For Dialog: **I still have the open question of what exactly anyone considers good HRI (that we can _somehow_ measure.)**

I'd say when the robot has to do something interactively with the human. The "help to cook" scenario mentioned in the old themed finals seemed more fitting than this tbh. Or cleaning together. 
The cooking scenario could be:
- human tells the robot what to fetch from the cupboard
- the robot has to bring the item and either hand it over to the human or pour the contents into a bowl while the human is right next to the robot, potentially even adding things to the same bowl at the same time. 
- Maybe the human changed his mind about an ingredient, tells this to the robot, and the robot has to bring the item back and fetch another one
- should be possible to interrupt the robot to give it a different command
- robot should be able to open drawers and either pick-place from them or just open them to assist the user
- the robot could be mixing things while the human is holding the bowl or vice-versa

I'd always think of HRI as useful interactive assistance, but it always should go hand in hand with actually manipulating and doing things. If it's only conversation, then a robot is not needed but an Alexa is enough. 

tl;dr: bring humans and robots closer together.



Back to the challenge: 
I agree that the story is not that nice yet. 
- guiding a blindfolded person seems risky
How about this:
The person who enters the arena is a friend of the host, tasked to take care of the house while the host is on vacation. However, there was no time for the host to show his friend around, so now the robot has to do this. Robot greets the person at the door, shows them where they can hang their jacket and asks them what the person would like to do first, based on a list of tasks given by the host. 
these tasks can be:
- water plants
- check if all doors are closed
- feed the cat
- ...
Then the user gets to pick a task, the robot will guide the person to the objects and point them out. It might be necessary to open drawers or cabinet doors in order to access the objects. 
- watering plants: get the watering can from sink area (either handover or just tell person) then robot guides person to plants and watches them watering. if the person does not follow or does not tilt the can, the robot reminds them to do so. 
- feed cat: grab cat food and pour it into the feeding bowl. The robot could do this but maybe he wants to be supervised
- check if all doors are closed when the friend leaves. (including cupboards)
Focus could be: ensure the person does as told e.g. recognize if they are not doing things correctly. Include more talking. e.g. person can ask the robot for more information on things at any given time. e.g. interrupt navigation or whatever they are currently doing. 

Probably unrealistic to change it now but wanted to write down the ideas. (totally not influenced by the cats next to me...)

SparkRibeiro21:
Since EC decided on a different type of task and a new PR has been created for the new hri task, we should close this PR.

Issue #916, opened 2024-10-30
Improve Receptionist


LeroyR:
## Is your idea/suggestion related to a problem? Please describe.

Something about receptionist..

The median score is high -> the task is solved
Task is very static. e.g. 
- The Robots are telling the person where to stand instead of looking at them dynamically
- The introduction/getting to know name and drink is also very predictable, boring and 'hardcoded' 
We do not have any pose/gesture recognition

## Describe the solution you'd like
- Keep/ReIntroduce Subjective Interaction scoring (survey)? -> some 'interesting' interaction has to happen
- Have multiple people around/front of the robot 
  -  The robot has to recognize the speaking person (sound source or moving lips?) to score eye contact points
- Reward completion -> add one more guest and reduce scores?
  -  The robot could handle introduction during guidance to save time
- The 2nd guests could point at one of the persons and ask the robot for name/drink of that person 
- Smalltalk during guidance: I don't want to be guided to the seating area in silence somehow reward doing 'smalltalk'.
  - Add points for stating some trivia about the person which has to be obtained by talking. (job/time of arrival/country of origin etc)

ARTenshi:
Hello,

I like these two ideas:

> Have multiple people around/front of the robot

> The 2nd guests could point at one of the persons and ask the robot for name/drink of that person

Another point:

>The door opening points are VERY misadjusted. Opening the door is much more complex than opening the cabinet door in storing groceries, and teams get 200 points for the cabinet door and 100 points for the entrance door.

Yes, but opening the front door is a recurrent task; increasing the score would bias what happens inside the arena for teams able to open the door; we might consider those few points as a partial score considering all the chances that a team will have to open the door (almost for every test).

sunava:
I or a student of mine will do a PR soon

PR #930, opened 2024-12-17
[Receptionist] updates for the challenge

Julislz:
# Proposed Solution for issue #916 
with @sunava 
- the robot shows both guests where the drinks are and detects if their favorite drink is available
- There are passive guests in the background
- Points for finding one hobby/interest of each guest
- At the end, only the guests get introduced 
- optional goal 'describing the first guest' changes. describe first guest to second guest **before** reaching the living room (updated 10.01)
- guest paces from left to right during conversation with robot (update 10.01)
-  ~~Added optional goal picking up a drink~~ (updated 10.01)
- Added optional goal for finding similar interest between guests or guest and host
- added a procedural guide for quicker and easier understanding of the challenge
- Changed scoresheet accordingly

## Updated procedure summarized
Both guests arrive separately. The robot either opens the door for the guest or
waits for them at the starting point. It greets the guest and asks for their name.
The robot then guides the guest to the beverage area, where it asks for their
favorite drink and checks if the drink is available on the table.  The robot gathers
one interest of the guest at a freely chosen moment.
After showing the guest the beverage area, the robot escorts them to the living
room and offers a free seat. Once both guests are seated, the robot introduces
them to each other.

## Other comments
- ~~I would suggest making it an obligation to log what the robot is saying.
 This makes the process easier for the referee in a noisy arena to   check the output quickly after the challenge~~

- The passive guests and the host could be spectators if it is not too much effort.


LeroyR:
Why do we want to introduce grasping here?

We should not remove person recognition (e.g. describing the guest) for another recognize object.
There are other stage one tasks to show object recognition. 


The gaze is currently very simple (straight during interaction, down during navigation). I also dont want any more of "please step to my left so i can introduce you" for introduction.

The improvement should be making sure the recognition/tracking of humans really works e.g. follow the head while the person moves. Ideas:
- the person simply paces from left to right side of robot during introduction
- another guest blocks the gaze direction so the robot has to adjust its position to see the guest
- person hides behind a lamp
- a conversation with two people at the same time -> look at the speaking person

We will not collect any logs. Robot should speak loud and repeat on request.

Robocup wants to move away from 'state machines' in tasks, the procedure makes the task too static. e.g. why smalltalk at specific points.

SparkRibeiro21:
I really like this update to receptionist!
Just a couple of things:

_- the robot shows both guests where the drinks are and detects if their favorite drink is available_
I really like this, it allows teams to show object detection! i would not add bonus points for picking up the object. There is already picking objects in a lot of other tasks. no need for this also in receptionist.

I think describing the first guest to the second guest is weird because if I look to the second guest and point to the first guest, why do i need to describe it?

I think with the new additions asking for: "_one or two interests and a profession to ask about_"  seems a bit much, let's just ask for one hobby for now, if it goes well than we can add more next year.

_+ 50 pts identify favorite drink_
and add a new one +"50 show where drink is" similarly to what is done in DEM where the robot has to point or show in the face where the drink is. As a guest if a robot tells me there is a drink in a cabinet, I do not want to know there is a drink and still have to look for it, i want the robot to tell me where the drink is.

The door opening points are **VERY** misadjusted. Opening the door is much more complex than opening the cabinet door in storing groceries, and teams get 200 points for the cabinet door and 100 points for the entrance door. In the entrance door you have to identify the handle, pull down the handle, pull the door, ... whereas the cabinet door is much simpler. I believe no team even tries to attempt this (even though a couple of teams showed in their qualification videos from last year they could open doors), means that the number of points doesn't compensate the risk that is attempting to open the door.  

personally not a fan of the small talk...

The way you are thinking, we don't have to do any introductions to the host (we don't even have to detect or know who the host is) and just have to introduce the first and second guest right?

I don't also agree with the speaker log... the robots must talk clearly, loudly and find solutions to have their messages understood like show in a tablet... 

I think the people talking should always be making some kind of movement... robots should only get points for looking at the person if it is dynamically real time adjust to the person whom it is talking to. I believe one small step to the left and the right should be enough to check this... we often see robots looking to the person coordinates and if the robot moves, it still looks to the place where the person was.  

"
_+ 50 pts look at person talking
+ 75 pts introduce both guests
+ 50 pts look at the person talking_
"
you have +50 looking at the person talking twice

Julislz:
@LeroyR 

Firstly thanks for the feedback,
- we can absolutely remove the optional goal of grasping

'*"We should not remove person recognition (e.g. describing the guest) for another recognize object. There are other stage one tasks to show object recognition." *
- i removed that point because it seemed very unnatural to just state perceived attributes. however we can keep that goal.
   it surely is important for a human service robot to detect attributes of humans

*"the person simply paces from left to right side of robot during introduction"*
- The suggestion is great and i would add that. I actually thought that this was already part of the challenge. But you are right it should be described in the text. 

*" another guest blocks the gaze direction so the robot has to adjust its position to see the guest"*
- another person blocking the guest is pretty difficult to implement because usually the robot and guest are standing  very close 
 during the introduction and face each other directly. Additionally this scenario rarely happens in real life. when you 
 stand close to another person and talk to them other people usually move around you because it is very rude to go inbetween. 
 But that we could add. That behind the guest other people move around. to see if the robot loses focus because of the background   humans.
 *" person hides behind a lamp"*
- What do you mean with hide behind a lamp? does that mean the robot has to find the guest that it should guide through
the arena?

*"a conversation with two people at the same time -> look at the speaking person"*
- interesing idea. that could be included when both guests are sitting in the living room.
 However and this includes your last point with not collecting any logs: the robocup arena is to noisy because it is a big event.
 in Eidhoven there was a band playing during the challenges. Therefore it will never simulate a real household environment which 
 is fine but we need to adjust the challenges and ratings accourdingly.
 That is the reason why i would log conversation and why i think it is difficult for the robot to detect who is speaking.

Julislz:
@SparkRibeiro21 
Thanks for the feedback,
*"Let's just ask for one hobby for now, if it goes well than we can add more next year"*

sounds great!

*"i want the robot to tell me where the drink is."*
good idea i will add that

*"The door opening points are VERY misadjusted"*
i agree that door opening is much more difficult than opening a cabinet door. I did not raaise the score because
i also read some backlash about raising the points because it is a reoccuring task. I would also raise the points
and mybe lower the score for cabinet opening. However i think some more opinions about that from othere Members would be great to get a better view on this issue

*" The way you are thinking, we don't have to do any introductions to the host (we don't even have to detect or know who the host is) and just have to introduce the first and second guest right? "*
well you have to detect the host to find a free spot in the living room. And if one of the guests interests 
match with the hosts interests the robot should look at the host and guest while telling the similarity. 

*" I don't also agree with the speaker log... the robots must talk clearly, loudly and find solutions to have their messages understood like show in a tablet..."*
i have answered that issue earlier. but in addition to that answer: To log what the robot says was also ment as a help for the referee. Because as a referee you need to keep many things in mind and sometimes you miss detailes. With logs you can check later.

akinobu1998:
For door opening, in my opinion, 200 bonus is big enough so we might consider reducing the SG bonus.

akinobu1998:
For
> Before reaching the living room, the robot gathers one additional piece of information about the guest at a freely chosen moment.

Regarding @LeroyR's comment, 
> why smalltalk at specific points.

I like to propose the idea that the guest to speak about them in their own time and way.

The guest might talk about them at their own time, or the robot might ask about the guest if it is required.

I think a kind of question generator will be helpful for the guest, now the guest has to memorize their name, hobby, and profession and it might confusing without a memo or something. 

Do we provide a list of names, professions, and hobbies? I think it is not necessary.

SparkRibeiro21:
I think you should standardize the usage of the terms: "hobby" and "interest" as it may be confusing, I personally think interest is the way to go.

I think this (\scoreitem[2]{10}{tell position of favorite drink}) has to be more significant: it is 3 times less than to look in the direction of navigation which is basically just moving the neck forward...

I have to insist that opening the door is too undervalued, maybe put is 125. It is just disappointing that no teams has ever opened a door and a significant number of teams show that they can do it in their qualification videos...

Apart from these I think everything is OK

SparkRibeiro21:
Just some small details I found necessary to clarify:

1. Here is the only time you talk about asking for the job, i think it is better to keep it standardized: calling it interests. 

   \item \textbf{Smalltalk} ask each guest after his job or interests during guidance or at the drinking area

   Would change to:

   \item \textbf{Smalltalk} ask each guest for one interest during guidance or at the drinking area

2. Here, why does it says snacks area? It should only be beverages area as the robots will only provide drinks.

   \item \textbf{Arriving Guests:} Both guests have a name, favorite drink, and a interest. An arriving guest will either step in front of the robot or wait behind the door to step in if the door is closed. Guests have to be guided to the beverage ~~and snack~~ area and then the living room where the robot will introduce the guests to each other. Each of the guest will arrive separately. 

 3. I think it needs to be clarified in the tasks/Receptionist.tex file, that the list of names and drinks is given on setup days however there is no list of interests. For old teams this is common knowledge but for new teams it helps that this is stated in the rulebook.
 
4. These penalty items should be under \scoreheading{Penalties} and not under \scoreheading{Bonus Rewards}:

	\penaltyitem[4]{30}{Describe the first guest to the second guest (per incorrect attribute)}
	\penaltyitem[1]{30}{wrongly stating a similarity} 
 
 
 (I would do these changes myself and spare some work to @Julislz as these are small and quick fixes. However I believe this is being done in a forked repo and not in a branch of the original repo and this way we do not have permissions to make such edits)

tasks/Receptionist.tex
```diff
@@ -49,7 +49,7 @@ \subsection*{Procedure}
 	looks at the conversational partner. During navigation robot looks in the direction where it is going. Persistently gazing towards unrelated 
 	person or incorrect direction while moving during the task deducts points.
     
-    \item \textbf{Smalltalk} ask each guest after his job or interests during guidance or at the drinking area
+    \item \textbf{Smalltalk} ask each guest after one interest
```
SparkRibeiro21, line 52: I think you missed this edit...
Should be something like:
\item \textbf{Smalltalk} ask each guest one interest during guidance or at the drinking area

Julislz: no i actually removed that on purpose because i would not mind asking the guest in the living room for one interest. However if you want it changed back i could do that

SparkRibeiro21: oh, ok for me no problem. Then i would just replace the "after " by "for" so the sentence is semantically correct.

Julislz: oh right, thanks :)

SparkRibeiro21 (approved):

SparkRibeiro21:
@Julislz we just finished the TC meeting and this PR will be approved. 
The only thing missing are some small clarifications regarding the looking at people:

\scoreitem[1]{75}{Look at the person talking or talked about}

1. Why did you add the "or talked about". The robot should always be looking at the person he is talking to.
2. should be [2] in scoreitem, not [1] right? as the robot will have to talk with two guests.

And we have also decided that the looking at the person talking should never be a static looking at the detection coordinate, but a dynamic looking. So if the person moves a bit to the sides the robot should move its continuously move its neck to face the person. So this:

	\item \textbf{Looking at person/direction of navigation:} During verbal interactions and descriptions of people, robot 
	looks at the conversational partner. During navigation robot looks in the direction where it is going. Persistently gazing towards unrelated person or incorrect direction while moving during the task deducts points. 

has to be changed to:

	\item \textbf{Looking at person:} During verbal interactions and descriptions of people, robot 
	looks at the conversational partner. The conversational partner will make small movements to each side to confirm the robot is dynamically looking at the person. Points for looking at the person talking, will only be awarded if the robot proves to continuously look at the moving person. 
	\item \textbf{Looking at direction of navigation:} During navigation robot looks in the direction where it is going. Persistently gazing towards unrelated person or incorrect direction while moving during the task deducts points. 
	
(i separated the two "lookings". This way is easier to understand)

Once this is fixed, the PR is ready for approval.

Julislz:
great to hear :)

*1. Why did you add the "or talked about". The robot should always be looking at the person he is talking to.*
This is for introducing the first and second guest to each other. so when stating the name,drink and interest of the first guest the robot should look at the first guest. But i get why this is confusing, so i will remove it

*2. should be [2] in scoreitem, not [1] right? as the robot will have to talk with two guests.*
yes 

*And we have also decided that the looking at the person talking should never be a static looking at the detection coordinate, but a dynamic looking. So if the person moves a bit to the sides the robot should move its continuously move its neck to face the person*
yes i like that change

SparkRibeiro21 (dismissed):

tasks/Receptionist.tex
```diff
@@ -73,26 +84,28 @@ \subsubsection*{To Referee}
...
 	\item Open the door when requested by the robot.
 \end{itemize}
 
 \subsubsection*{To OC}
 
+
 During setup day:
 \begin{itemize}
-	\item Provide the dorbell sound (if not using door knocking).
+	\item  the list of names and drinks is given (a list of interests will not be provided)
```
LeroyR, line 98: This should also be a TODO list for referees:

```suggestion
        \item Provide the list of names and drinks.
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,30 @@
 \begin{scorelist}[startbutton=false,timelimit=5]
 	\scoreheading{Main Goal}
-	\scoreitem[2]{15}{Guide the guest to the other guests (navigate to the guest group)}
-	\scoreitem[2]{50}{Look in the direction of navigation or at the navigation goal}
-	\scoreitem[2]{50}{Introduce a new guest to every other guest and the host}
+	\scoreitem[2]{30}{show the guest around (navigate to the beverage area and living room)}
+	\scoreitem[1]{30}{Look in the direction of navigation or at the navigation goal}
+	\scoreitem[2]{20}{identify favorite drink}
+	\scoreitem[2]{20}{identify interest of guest}
```
LeroyR, line 6: what does `identify` mean? Is it fine for the robot to just ask for a hobby and listen for an answer?

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,30 @@
...
+	\scoreitem[2]{75}{Look at the person talking}
+	\scoreitem[1]{75}{Introduce both guests to each other}
+
 	
 	\scoreheading{Bonus Rewards}
-	\scoreitem[2]{100}{Open the entrance door for a guest}
-	\scoreitem[4]{30}{Describe the first guest to the second guest (per corect attribute)}
-	\penaltyitem[4]{30}{Describe the first guest to the second guest (per incorrect attribute)}
+	\scoreitem[2]{125}{Open the entrance door for a guest}
+	\scoreitem[1]{50}{find the same interest between two guests or a guest and the host}
```
LeroyR, line 15:
```suggestion
	\scoreitem[1]{50}{State a shared interest between two or more persons}
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,30 @@
...
 
 
 	\scoreheading{Penalties}
-	\penaltyitem{50}{Wrong guest information was memorized (continue with wrong name or drink)}
+	\penaltyitem[2]{50}{Wrong guest information was memorized (continue with wrong name or drink)}
+	\penaltyitem[2]{50}{interest was not or wrongly memorized}
 	\penaltyitem{50}{Persistent inappropriate gaze (away from conversational partner)}
-	\penaltyitem{10}{Persistent gaze not in the direction of the navigation while moving.}
+	\penaltyitem{25}{Persistent gaze not in the direction of the navigation while moving.}
+	\penaltyitem[2]{20}{wrongly identifying the favorite drink on the table}
```
LeroyR, line 24: Here are to many penalties. e.g. Why is `wrongly identifying the favorite drink on the table` a penalty if `identify favorite drink` is a score item. You already get no points for the (missed) scoreitem -> we can't apply a penalty to 0 points.

tasks/Receptionist.tex
```diff
@@ -4,13 +4,15 @@ \subsection*{Description}
...
 \textbf{Main goal:}
-    Introduce and usher two newcomers to a party and maintain appropriate gaze direction (at person, direction of navigation).
+    The robot must welcome and assist two newcomers at a party and maintain appropriate gaze direction while conversation (at person, direction of navigation).
 
 % \vspace{-5pt}
 \textbf{Optional goals:}
 \begin{enumerate}[nosep]
 	\item Open the entrance door for each arriving guest.
-	\item Describe the first guest for the second guest.
+	\item Opening the door for the arriving guests
```
LeroyR, line 13: Opening the door is listed twice

tasks/Receptionist.tex
```diff
@@ -73,26 +84,28 @@ \subsubsection*{To Referee}
...
 \end{itemize}
 
 \subsubsection*{To OC}
 
+
 During setup day:
 \begin{itemize}
-	\item Provide the dorbell sound (if not using door knocking).
+	\item  the list of names and drinks is given (a list of interests will not be provided)
 \end{itemize}
```
LeroyR, line 99: This misses `announce beverage location`

tasks/Receptionist.tex
```diff
@@ -73,26 +84,28 @@ \subsubsection*{To Referee}
...
 
+
 At least two hours before test:
 \begin{itemize}
 	\item Announce starting position.
-	\item Announce host's name and favorite drink.
-	\item Recruit volunteers as host and guests.
+	\item Announce host's name, favorite drink and interest.
+	\item Recruit volunteers as host and two guests and passive guests .
```
LeroyR, line 106: May add numbers here `Recruit x volunteers. (job listings)` ?

LeroyR (changes requested):
I still do not see why we want to add object recognition to this task. But i will approve the beverage area part if no one else voices any concerns.

We may need more clarification regarding:
- what if the robot ask interest of one of the passive guest and introduces them. (penalty or extra points or just nothing)
- Do passive guests answer questions?

 The task now feels more `statemachine` than before e.g. do this then this then that, maybe @moriarty @justinhart can comment if this is OK (as we had the goal of having less `statemachine` last year)

I like forcing the guests to be more active (e.g. evading the gaze, could add background? conversations) The Guest should definitely move after being seated so the robot has to re identify the guest later.

johaq (dismissed):

tasks/Receptionist.tex
```diff
@@ -23,36 +25,45 @@ \subsection*{Setup}
...
 	\end{itemize}
 
 \end{itemize}
 
 \subsection*{Procedure}
+Both guests arrive separately. The robot either opens the door for the guest or waits for them at the starting point. It greets the guest and asks for their name. The robot then guides the guest to the beverage area, where it asks for their favorite drink and checks if the drink is available on the table and where it stands. The robot finds out the interest of the guest at a freely chosen moment. After showing the guest the beverage area, the robot escorts them to the living room and offers a free seat. Once both guests are seated, the robot introduces them to each other.
 \begin{itemize}
-    \item \textbf{Introductions:} When introducing guests, the robot must clearly identify the person being introduced and state their name and favorite drink. Introducing two people means to introduce them to each other.
+	\item \textbf{Greeting guests:} the person paces a little to the left and right during conversation with the robot. other people might appear in the background
```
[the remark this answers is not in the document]
Julislz: was that also discussed in the TC meeting? Cause  i was specifically told that this should be added.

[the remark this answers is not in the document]
hawkina: It was specifically requested above in order to make sure the robot keeps eye-contact. It might be enough if one person does this just to see that the robot is capable of keeping tracking on them. If we hand out post-its with Name, Drink, Interest, it should be fine I think.

hawkina:
I think it overall looks fine. 
- I'd probably like to raise the points for opening the door even more, but have to compare to other challenges so that they can be raised equally across all tasks. 
- Not sore how we can avoid the "state machine" issue without introducing things like: "the robot can be interrupted during navigation by another person stepping in front of it and asking a question like: "is there still cola available?" or "Did Peter arrive yet?"  - this might make it a bit too difficult though. Maybe this could be done as a bonus challenge once the robot completes everything else, but then again, time limit is probably an issue.

hawkina (approved):

LeroyR:
> * Not sore how we can avoid the "state machine" issue without introducing things like: "the robot can be interrupted during navigation by another person stepping in front of it and asking a question like: "is there still cola available?" or "Did Peter arrive yet?"  - this might make it a bit too difficult though. Maybe this could be done as a bonus challenge once the robot completes everything else, but then again, time limit is probably an issue.

E.g keeping the current tasks, you would remove the descriptive procedure and design everything more open ended where guests can arrive any time with smalltalk, drinks and seating somehow happening dynamically.

I think it is more of an design issue instead of patching on interruptions or something.

We can not really change it this close to the draft release anyway **but** we already deferred #805 last year. We just kinda forgot?. This issues comes from not having a roadmap :sweat_smile:

LeroyR (approved):

tasks/Receptionist.tex
```diff
@@ -73,26 +83,31 @@ \subsubsection*{To Referee}
...
 \end{itemize}
 
 \subsubsection*{To OC}
 
+
 During setup day:
 \begin{itemize}
-	\item Provide the dorbell sound (if not using door knocking).
+	\item announce beverage location
+	\item the list of names and drinks is given (a list of interests will not be provided)
```
LeroyR, line 99: Names is covered under rule:scenario_names, drinks are a defined category.
You can remove this.
```suggestion
	
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,28 @@
 \begin{scorelist}[startbutton=false,timelimit=5]
 	\scoreheading{Main Goal}
-	\scoreitem[2]{15}{Guide the guest to the other guests (navigate to the guest group)}
-	\scoreitem[2]{50}{Look in the direction of navigation or at the navigation goal}
-	\scoreitem[2]{50}{Introduce a new guest to every other guest and the host}
+	\scoreitem[2]{30}{show the guest around (navigate to the beverage area and living room)}
```
LeroyR, line 3:
```suggestion
	\scoreitem[2]{30}{Show the guest around (navigate to the beverage area and living room)}
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,28 @@
...
+	\scoreitem[1]{30}{Look in the direction of navigation or at the navigation goal}
+	\scoreitem[2]{20}{confirm favorite drink}
```
LeroyR, line 5:
```suggestion
	\scoreitem[2]{20}{Confirm favorite drink}
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,28 @@
...
+	\scoreitem[2]{20}{confirm interest of guest}
```
LeroyR, line 6:
```suggestion
	\scoreitem[2]{20}{Confirm interest of guest}
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,28 @@
...
+	\scoreitem[2]{20}{tell position of favorite drink}
```
LeroyR, line 7:
```suggestion
	\scoreitem[2]{20}{Tell position of favorite drink}
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,28 @@
...
-	\penaltyitem[4]{30}{Describe the first guest to the second guest (per incorrect attribute)}
+	\scoreitem[2]{125}{Open the entrance door for a guest}
+	\scoreitem[1]{50}{State a shared interest between two or more persons}
+	\scoreitem[4]{30}{Describe the first guest to the second guest (per correct attribute)}
 
 
 	\scoreheading{Penalties}
-	\penaltyitem{50}{Wrong guest information was memorized (continue with wrong name or drink)}
+	\penaltyitem[2]{50}{Wrong guest information was memorized (continue with wrong name or drink)}
+	\penaltyitem[2]{50}{interest was not or wrongly memorized}
```
LeroyR, line 21:
```suggestion
	\penaltyitem[2]{50}{Interest was not or wrongly memorized}
```

scoresheets/Receptionist.tex
```diff
@@ -1,22 +1,28 @@
...
 	\penaltyitem{50}{Persistent inappropriate gaze (away from conversational partner)}
-	\penaltyitem{10}{Persistent gaze not in the direction of the navigation while moving.}
+	\penaltyitem{25}{Persistent gaze not in the direction of the navigation while moving.}
+	\penaltyitem[4]{30}{Describe the first guest to the second guest (per incorrect attribute)}
+	\penaltyitem[1]{30}{wrongly stating a similarity}
```
LeroyR, line 25:
```suggestion
	\penaltyitem[1]{30}{Wrongly stating a similarity}
```

tasks/Receptionist.tex
```diff
@@ -23,36 +24,45 @@ \subsection*{Setup}
 	\begin{itemize}
 	\item The test takes place in the living room.	
 	\item The robot starts inside the \Arena{} at a predefined location.
+	\item a table with drinks is prepared near the living room
```
LeroyR, line 27:
```suggestion
	\item A table with drinks is prepared near the living room
```

tasks/Receptionist.tex
```diff
@@ -23,36 +24,45 @@ \subsection*{Setup}
...
 	\end{itemize}
 
 \end{itemize}
 
 \subsection*{Procedure}
+Both guests arrive separately. The robot either opens the door for the guest or waits for them at the starting point. It greets the guest and asks for their name. The robot then guides the guest to the beverage area, where it asks for their favorite drink and checks if the drink is available on the table and where it stands. The robot finds out the interest of the guest at a freely chosen moment. After showing the guest the beverage area, the robot escorts them to the living room and offers a free seat. Once both guests are seated, the robot introduces them to each other.
 \begin{itemize}
-    \item \textbf{Introductions:} When introducing guests, the robot must clearly identify the person being introduced and state their name and favorite drink. Introducing two people means to introduce them to each other.
+	\item \textbf{Greeting guests:} the person paces a little to the left and right during conversation with the robot. other people might appear in the background
```
LeroyR, line 45:
```suggestion
	\item \textbf{Greeting guests:} The person paces a little to the left and right during conversation with the robot. Other people might appear in the background
```

tasks/Receptionist.tex
```diff
@@ -23,36 +24,45 @@ \subsection*{Setup}
...
+	
+	\item \textbf{Looking at person:} During verbal interactions and descriptions of people, robot 
+	looks at the conversational partner. The conversational partner will make small movements to each side to confirm the robot is dynamically looking at the person. Points for looking at the person talking, will only be awarded if the robot proves to continuously look at the moving person. 
+	
+	\item \textbf{Looking at direction of navigation:} During navigation robot looks in the direction where it is going. Persistently gazing towards unrelated person or incorrect direction while moving during the task deducts points. 
+    
+    \item \textbf{Smalltalk} ask each guest for one interest
```
LeroyR, line 52:
```suggestion
    \item \textbf{Smalltalk} Ask each guest for one interest.
```

tasks/Receptionist.tex
```diff
@@ -23,36 +24,45 @@ \subsection*{Setup}
...
+	\item \textbf{Greeting guests:} the person paces a little to the left and right during conversation with the robot. other people might appear in the background
+	
+	\item \textbf{Looking at person:} During verbal interactions and descriptions of people, robot 
+	looks at the conversational partner. The conversational partner will make small movements to each side to confirm the robot is dynamically looking at the person. Points for looking at the person talking, will only be awarded if the robot proves to continuously look at the moving person. 
+	
+	\item \textbf{Looking at direction of navigation:} During navigation robot looks in the direction where it is going. Persistently gazing towards unrelated person or incorrect direction while moving during the task deducts points. 
+    
+    \item \textbf{Smalltalk} ask each guest for one interest 
+    
+    \item \textbf{Finding the drink} the robot shows both guest the drinking area where it will ask for their favorite drink and tell the guest if and where (left, center right is enough) that drink is available on the table.
```
LeroyR, line 54:
```suggestion
    \item \textbf{Finding the drink} The robot shows both guest the drinking area where it will ask for their favorite drink and tell the guest if and where (left, center right is enough) that drink is available on the table. 
```

tasks/Receptionist.tex
```diff
@@ -73,26 +83,31 @@ \subsubsection*{To Referee}
 
 The referees need to:
 \begin{itemize}
-	\item Assign name and drink to 3 volunteers.
+	\item Provide the list of names and drinks on setup day
+	\item Assign name, drink and one interest to 3 volunteers.
 	\item Arrange (and re-arrange) people in the living room.
+	\item change the selection in the beverage area
```
LeroyR, line 89:
```suggestion
	\item Change the selection in the beverage area.
```

tasks/Receptionist.tex
```diff
@@ -73,26 +83,31 @@ \subsubsection*{To Referee}
...
 	\item Open the door when requested by the robot.
 \end{itemize}
 
 \subsubsection*{To OC}
 
+
 During setup day:
 \begin{itemize}
-	\item Provide the dorbell sound (if not using door knocking).
+	\item announce beverage location
```
LeroyR, line 98:
```suggestion
	\item Announce beverage location.
```

tasks/Receptionist.tex
```diff
@@ -73,26 +83,31 @@ \subsubsection*{To Referee}
 
 The referees need to:
 \begin{itemize}
-	\item Assign name and drink to 3 volunteers.
+	\item Provide the list of names and drinks on setup day
```
LeroyR, line 86:
```suggestion
	
```

tasks/Receptionist.tex
```diff
@@ -23,36 +24,45 @@ \subsection*{Setup}
...
 	
 	\item \textbf{Seating People:} The robot must point at a place or location where the guest can sit.
 	
 	\item \textbf{Switching Places:} Guests may switch places after they were seated.
 	
-	\item \textbf{Describing the First Guest:} Naming 4 characteristics of the first guest, i.e., color of clothes, color of hair, gender, and age, earns bonus points.
-	\item \textbf{Looking at person/direction of navigation:} During verbal interactions and descriptions of people, robot 
-	looks at the conversational partner. Robot can point at the person being introduce/described or alternate gaze between two people. During navigation robot looks in the direction where it is going. Persistently gazing towards unrelated 
-	person or incorrect direction while moving during the task deducts points. 
+	\item \textbf{Introductions:} When introducing guests, the robot must clearly identify the person being introduced and state their name favorite drink and a interest. Introducing two people means to introduce them to each other.
```
LeroyR, line 60:
```suggestion
	\item \textbf{Introductions:} When introducing guests, the robot must clearly identify the person being introduced and state their name favorite drink and an interest. Introducing two people means to introduce them to each other. 
```

LeroyR (changes requested):
Capitalize and stuff

tasks/Receptionist.tex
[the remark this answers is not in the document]
SparkRibeiro21: what we talked in the TC meeting is that is must be added one way or the other, a lot of robots don't do this and many times robots are talking to a person, the person moves aw bit and the robot keeps looking to an empty space. This just seems very weird. As @hawkina said, doing this with one person is enough to understand if the robot does dynamic tracking or not.

LeroyR (approved):

PR #933, opened 2025-01-06
Remove serve breakfast

LeroyR:
## Description

Because of scheduling issues and to relieve the teams of some stress we decided to remove a stage one task. We decided to remove serving breakfast from the list of stage one task.

## Other comments

We have concerns that the robots only interact with the environment by doing pick and place. Other actions (like wiping, open/closing drawers or using tools) should therefore be incorporated in other tasks.


akinobu1998 (dismissed):

SparkRibeiro21 (approved):

ARTenshi (dismissed):
OK

PR #935, opened 2025-01-06
Storing groceries changes

LeroyR:
## Description

Storing Groceries changes to move the league forward. We want robots to start using the door handles and pick items directly from the shopping bag.
Also this incorporates pouring from the removed serving breakfast task by refilling a dry storage container.

- Increase time limit to 7 minutes.
- Opening door is now a main goal. 
  - Additionally we close both doors to have a more realistic scenario.
- Remove the team dependent setup (request door open, objects) instead the robot should just handle the situation.
  - Select the non heavy object if grasping is impossible for the robot
  - Requesting assistance for door opening
- Incorporate pouring by refilling a dry storage container with cornflakes.
  - Container is placed on the table for now. 
- Grasping objects from a shopping bag
  - placed next to table.


LeroyR:
Reopen from 79cfe5a4c7de2976dcd65489d975b768fcccffb6 as chktex was stuck

ARTenshi:
>Opening door is now a main goal. 

Is it? It might be too hard to include this "Opening door is now a main goal" and making it mandatory at the same time.

LeroyR:
> > Opening door is now a main goal.
> 
> Is it? It might be too hard to include this "Additionally we close both doors to have a more realistic scenario" and making it mandatory at the same time.

We can keep it optional but i would still keep all doors closed.

Rather than the team telling the referee, the robot can just state that it could not open the doors and ask for help (like teams did for the chairs at the table during the last world cup)

Optional goals only enforce teams to solve a part of the main task first before attempting the optionals. But opening the door is the first action in the task anyway. Do we want to give 0 point for a team just opening the door but failing to grasp an item afterwards. We could say that the robot did nothing to solve the main part of storing groceries but then give 15 points for reaching the location?

tldr; I rather remove all "ask teams how to setup the arena" and then being an optional goal does not make much sense imho.

SparkRibeiro21:
I agree with what it is right now: We put it as a main goal and just add the DEM. This way the doors always start closed and robots can ask for help.

SparkRibeiro21 (approved):

akinobu1998 (approved):

ARTenshi (approved):
OK

PR #936, opened 2025-01-07
Updates clean table 2025

SparkRibeiro21:
## Description

As discussed in the TC meeting (06/01/2025) here are again the updates to the clean the table, hopefully this PR doesn't get closed instead of being edited...

Added two drinks to the table to be placed inside the trash bin. Creating a much more realistic scenario, where there is tableware and cutlery to be placed inside the dishwasher and drinks inside the trash bin (in the previous version there were also foods to be restored in the cabinet which were removed since Storing Groceries will remain in stage 1).
 
 Proposed changes:
 	
 	- Added "Wipe the table area where the drink was" to optional goals 
 	- Added two drinks will be on the table that must be placed inside the trash bin
 	- Added wiping object (i.e. sponge) must be placed somewhere near (by the referee) for wiping, ideally next to dishwasher tab
 	- Moved "Placing the items in the dishwasher correctly" from optional goals to main goals (already was in the main goals in the scoresheet, must have been forgotten in a previous update)
 	
 	- Standardized picking plate bowl and cup, there was a 10 point difference between them (dont think a 10 points makes a difference, this way is standardized, and is less bullet points in scoring and in DEM, not as confusing for refs)
 	- Added picking and placing the drinks scores to scoresheet
 	- Added DEM to picking and placing drinks 
 	- Increased points for placing objects correctly in a dishwasher as a human would by 25 (this should be the main focus of the task...)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ BUILD_DIR:=$(ROOT_DIR)/.build
 SILENT = @
 MSG      = echo
 HASRUBBER:=$(shell which rubber)
+XELATEX:=$(shell which xelatex)
 
 ## COLORS 
 RESET       = tput sgr0
@@ -51,6 +52,9 @@ ifdef HASRUBBER
 	$(SILENT) rubber --unsafe --pdf --force $$<
 else
 	$(SILENT) latexmk -Werror -shell-escape -silent -pdf -interaction=nonstopmode -outdir=$${BUILD_DIR} $$<
+endif
+ifdef XELATEX
+	$(SILENT) xelatex $$<
 endif
 
 ## Generate Summary Rule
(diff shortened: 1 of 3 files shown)


LeroyR:
Only remarks For Points:
- When combining points: I think we can could combine cutlery+plate for bonus point or just give bonus for the plate (needs two movements)
- We can decrease points for place (more like throwing randomly) inside the dishwasher to 25
- Wiping should give more points (it is pick for 50 + tool action **NEW**) maybe 2x100 or 1x150? 
- Pick and place the dishwasher tab is 100+200=300 points which seems rather high in comparison
  - but in sg picking up a tiny objects is 70+50=130

ARTenshi (approved):
OK

PR #938, opened 2025-01-14
Higher number of problems for egpsr

LeroyR:
## Description

EGPSR is boring for the audience as most of the time is spend searching for problems. This tries to deal with that by increasing the number of problems (depending on the arena size). 
To encourage solving different categories of problems (e.g. finding 3x trash is boring) solving the same category of problem now incurs a penalty which can be compensated by solving more problems.

Examples for reaching maximum score :
- (3 problems): Find and solve 3 different problem categories 
- (4 problems): Find and solve 2 different problem categories twice
- (5 problems): Find and solve 1 category once and another 4 times
- (5 problems): Find and solve 1 category 5 times and continue to find 8 additional problems of the other categories

* state the minimum number of generated problems is 8.
* solving the same Category of Problem incurs a penalty. 
* no limit on the number of solved problems

## Other comments
The scoresheet is rough. Maybe this should be more like table for referees to note down the problems and penalties or something.

Command Generator has to modified and moved to this namespace.

### DIFF
--- a/scoresheets/EGPSR.tex
+++ b/scoresheets/EGPSR.tex
@@ -1,15 +1,31 @@
 \begin{scorelist}[timelimit=10]
-	\scoreheading{Main Goal}
+	\scoreheading{Main Goal (can be repeated unlimited times)}
 	\scoreitem[3]{150}{Find and clearly state an encountered problem}
 	\scoreitem[3]{650}{Solve a problem}
 
+	\scoreheading{Penalties}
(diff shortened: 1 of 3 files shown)


johaq (approved):
Looks good

hawkina:
Would suggest to reduce some of the points a bit. 650 is a bit much if it is possible to get it 8x650...
should we list a few more example categories?

hawkina (approved):

SparkRibeiro21 (approved):

PR #953, opened 2025-03-18
[GO feedback] SG: reduce point for opening the second cabinet door

LeroyR:
## Description

During the German open FAQ teams we specified that asking the referee to open one door still enables the robot to score opening the second door. As the first door is much harder to open (need to use handle or rims ) than the second (possible grasp behind) It should reward a different number of points.

### DIFF
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -6,8 +6,8 @@
 	\scoreitem[5]{15}{Perceiving objects in shelf and saying on which layer the currently handled object should be placed}
 	\scoreitem[5]{15}{Placing an object in the cabinet}
 	\scoreitem[5]{50}{Placing an object next to similar objects on the cabinet}
-	\scoreitem[1]{200}{Opening the first cabinet door }
-	\scoreitem[1]{100}{Opening the second cabinet door }
+	\scoreitem[1]{200}{Opening the first cabinet door}
+	\scoreitem[1]{100}{Opening the second cabinet door}
 	\scoreitem{300}{Pouring cereal into the container}
 
 	\scoreheading{Bonus Rewards}
--- a/scoresheets/StoringGroceries.tex
+++ b/scoresheets/StoringGroceries.tex
@@ -6,7 +6,8 @@
 	\scoreitem[5]{15}{Perceiving objects in shelf and saying on which layer the currently handled object should be placed}
 	\scoreitem[5]{15}{Placing an object in the cabinet}
 	\scoreitem[5]{50}{Placing an object next to similar objects on the cabinet}
-	\scoreitem[2]{100}{Opening a cabinet door }
+	\scoreitem[1]{200}{Opening the first cabinet door }
+	\scoreitem[1]{100}{Opening the second cabinet door }
 	\scoreitem{300}{Pouring cereal into the container}
 
 	\scoreheading{Bonus Rewards}


ARTenshi:
I am against partial points, at least in bonus tasks – in this case, either you open both doors or none.

LeroyR:
> I am against partial points, at least in bonus tasks – in this case, either you open both doors or none.

This is not a bonus task nor is this solution giving partial points. Both doors are currently single normal scoring items.

Also, not giving partial points results in no team attempting to solve cool stuff.

hawkina (approved):

ARTenshi:
>not giving partial points results in no team attempting to solve cool stuff

But giving partial points causes exactly the same, as this is the point of this PR (no teams attempting the hard and novel action of opening the first door).

scoresheets/StoringGroceries.tex
```diff
@@ -6,7 +6,8 @@
 	\scoreitem[5]{15}{Perceiving objects in shelf and saying on which layer the currently handled object should be placed}
 	\scoreitem[5]{15}{Placing an object in the cabinet}
 	\scoreitem[5]{50}{Placing an object next to similar objects on the cabinet}
-	\scoreitem[2]{100}{Opening a cabinet door }
+	\scoreitem[1]{200}{Opening the first cabinet door }
+	\scoreitem[1]{100}{Opening the second cabinet door }
```
MatthijsBurgh, line 10:
```suggestion
	\scoreitem[1]{200}{Opening the first cabinet door}
	\scoreitem[1]{100}{Opening the second cabinet door}
```

hawkina:
I overall agree but would make it 300 for first door and 100 for second maybe.

MatthijsBurgh (changes requested):
I would change this to "Opening door by using the handle" and "Opening door by pulling the door". As a team should get the same points for opening the first and/or the second door by the handle.

johaq:
I don't think it is the purpose of the Rulebook to impose "how" to solve something but just "what" should be solved.
Why would a team that opens the first door by the edge of the door get less points than using the handle?

MatthijsBurgh:
The first comment in this PR clearly starts by highlighting the difference between opening a door by grasping the handle vs putting the robot gripper behind the door.

Of course, when both cabinet doors are closed, you can't put the gripper behind the door. But when the robot already opened one by the handle. It could go for the easy option for the second cabinet door or go for the difficult one, using the handle.

So I think @LeroyR wants to differentiate based on how the task is executed and not the number of executions

LeroyR:
> The first comment in this PR clearly starts by highlighting the difference between opening a door by grasping the handle vs putting the robot gripper behind the door.

There is - but we should not start to evaluate the difficulty of the chosen solution (don't want to open that can of worms for referee discussions). We should just design the challenge in better way. I don't think this should be our focus for now, but the door handles could be remounted for each team or we could put random obstacles on the ground to block the optimal position :sweat_smile: 

For now we just need to assign points related to the estimated difficulty. Opening the 2nd door is easier regardless if the robot opts to use a complicated solution (handle) or not.

> 
> Of course, when both cabinet doors are closed, you can't put the gripper behind the door. But when the robot already opened one by the handle. It could go for the easy option for the second cabinet door or go for the difficult one, using the handle.
> So I think @LeroyR wants to differentiate based on how the task is executed and not the number of executions

No. We had this discussion and came to the conclusion that the Rulebook should not dictate the solution.

SparkRibeiro21:
Overall I agree with a significant decrease in points for the second door, without stating how a robot should open the door.
Comparing to other stage 1 tasks (where robots have to "open furniture"): there is the door opening in receptionist for 200 points. Personally I do not think we should go much higher than this for the first cabinet door opening as the difficulty of both is comparable. For the second one, I think a drastic decrease in necessary, I would suggest anything between 50 and 100.

sunava (approved):

PR #955, opened 2025-03-18
Change furniture requirements

LeroyR:
I Noticed that we don't have requirements for a cabinet with doors. Also never really saw the cupboard with drawers in use.

So this changes the requirement of "cupboard with drawers" to:

-  chest of drawers, and 
- a high cabinet with doors


ARTenshi:
Let me check this in the next EC meeting as the requirements have been submitted several weeks ago

sunava (approved):

johaq (approved):

PR #956, opened 2025-03-19
(CI) use newer lint action

MatthijsBurgh:
Current one fails to run, but this doesn't show because of settings

The new action allows for checking only updated files in a PR. When we would do that, we could remove the `continue-on-error`. And enforce the linting on updated files. Which would slowly increase the linting on the entire rulebook.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: LaTeX linter (chktex)
-        uses: j2kun/chktex-action@v2.0.0
+        uses: j2kun/chktex-action@v2.1.0
         # Provide this output for context, but don't fail builds
         continue-on-error: true
         env:


LeroyR:
Disabled this via:
https://github.com/RoboCupAtHome/RuleBook/pull/937

MatthijsBurgh:
But the current version is broken.

So I suggest to use the new version, but always check all files. So the behaviour doesn't change. But at least it works again.

LeroyR:
The new version uses a Github bot setup to automatically suggest changes but will not update them: thus it opens change requests and blocks all pulls which then require manual intervention.

We ignored the output anyway. We can remove the action for now.
I am open to enable this again after all checks pass to test that behavior again.

LeroyR (dismissed):

MatthijsBurgh:
I will do some tests to check whether we can disable the review functionality

hawkina (approved):

Issue #950, opened 2025-03-18
[GO] Score for manipulation items feel too low


LeroyR:
Even if the max score of the receptionist vs e.g. storing groceries is comparable. The difference between a Team doing a single pick and place and a team doing no manipulation is just 100 points. 

I think the 'reduced points for repetition' as in egpsr may be a good solution allowing fast/stable/reliable robots to continue scoring while increasing the difference between at least one successful pick and place and none at all. Increasing the difference between some and no manipulation while not overvaluing manipulation focussed tasks. Currently all manipulation tasks are repetition heavy and a blanket increase would inflate the scores.

ARTenshi:
In theory, no team can score any points (including bonuses) unless they perform a main goal task. Also, I am against too permissive Deus ex Machina, at least for core skills in a given task (for example, bypass manipulation in a manipulation task or perception in a perception task).

sunava:
"optional goal "one time a full transporting without human help extra points"

PR #961, opened 2025-04-07
Increase points for first manipulation

LeroyR:

## Description

Closes issue #950

The Manipulation tasks (storing groceries, cleanup) are  focused on repetition even though the manipulation rate of most teams is pretty small. With the consolation points for navigation and recognition the difference between a team having grasped and placed something and the team only recognizing and showing bounding boxes can be _negative_:

Team a: nav + 5 bounding boxes:  90pts
Team b: nav + pick (wrong detection) + place (wrong location): 80pts

With deus ex machina this can increase even more. Due the repetitive nature a increase in the points for pick/place overvalues the individual task thus i opted for just adding a one time bonus.

This change:
- adds two 50 point (one time) boni for pick and place.
- Adds penalty for wrong detection (e.g showing 7 detection but 4 of them are wrong should score 3*15)

## Additional Remarks

- Is this our preferred solution?
- Should something like this also be added to cleanup?


SparkRibeiro21:
Agree with your suggestion. You should probably also add this to the Clean Up task.
It is a shame that more and more we are moving away from manipulation tasks. The removal of Serve Breakfast (going from 3 to 2 manipulation focused tasks) definitely makes manipulation not a priority for RoboCup@Home, as point wise it is better to focus on other areas...

johaq:
I agree with the change but the Items should be in the regular scoring section not in the bonus scoring imo.

LeroyR:
> I agree with the change but the Items should be in the regular scoring section not in the bonus scoring imo.

I had the concerns that leaving this is the main section makes it more _unclear_ for the referee.

As normal score item: i assume the ref (on handover) may mark down this **and** `Picking up an object for transportation to the cabinet` while marking `A human handing an object over to the robot` once.
Then during calculation it will be missed and the score is 50+50-50=50 instead of 0.

I think it is fine like this or maybe even:
```
+       \scoreitem{50}{Picking any Object (First time bonus}
```
?

We really need to clean up deus ex machina. :sweat_smile:

johaq (approved):

sunava (approved):

hawkina (approved):

PR #963, opened 2025-04-07
Clarify EGPSR scoring

LeroyR:
## Description

Clarify scoring of egpsr: repetition penalty is applied before deus ex machina penalties.

### DIFF
--- a/scoresheets/EGPSR.tex
+++ b/scoresheets/EGPSR.tex
@@ -7,6 +7,7 @@
 	\penaltyitem[1]{100}{Find repeated problem category}
 	\penaltyitem[1]{300}{Solving repeated problem category for the 2nd time}
 	\penaltyitem[1]{500}{Solving repeated problem category for the 3rd (or more) time}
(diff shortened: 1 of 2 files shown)


ARTenshi:
Given this PR for EGPSR, I would also suggest updating the following sentence:

"Some persons in the arena will have requests for the robot. They will raise their hand if the robot is in the same room."

to something like:

"They will raise their hand if the robot requests it."

The reason is that with the current phrasing, people might end up holding their hand up as soon as the robot enters the room and while the robot is still busy doing something else—like looking for trash or misplaced items. This way, the interaction feels a bit more natural and avoids unnecessary waving.

SparkRibeiro21:
@ARTenshi so the robot should announce it is searching for people with commands? It feels a bit weird, but I agree with you that having a person waving their arms while the robot is doing a different task is also not ideal.

johaq:
This is a good change.
Wrt waving people, I don't think we should hard require this. We actually want to encourage teams to find people on the fly and not with go to location -> announce -> look around. My usual instructions to the volunteers is start waving when you think the robot is looking for you.

johaq (approved):

PR #970, opened 2025-05-12
Updated scoresheets: More info for posterior analysis. Added time for…

SparkRibeiro21:
Updated score sheets: More info for posterior analysis. Added time for each goal columns. Referee now has to write: time for each goal, motive why task ended and number of missed attempts of a goal.

### DIFF
--- a/documents/scoresheets.tex
+++ b/documents/scoresheets.tex
@@ -6,7 +6,7 @@
 %%
 %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-\documentclass[11pt, a4paper]{book}
+\documentclass[10pt, a4paper]{book}
 
 \usepackage{soul}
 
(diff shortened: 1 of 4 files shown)


SparkRibeiro21:
Tried to improve score sheets to best encompass data for posterior analysis. This way when analyzing tasks after an event, decisions like: how much time do teams on average need to do something, is X easier than Y, how many teams have attempted something, can actually be discussed using data.

SparkRibeiro21:
[scoresheets.pdf](https://github.com/user-attachments/files/20163718/scoresheets.pdf)

LeroyR:
The current text size gets too small.

In (E)GPSR/Final, the "Point marking" area is often used to note partial scores or to clarify what exactly happened.

What do we think about using one sheet per try? Maybe just landscape-oriented sheets?

We might need a dedicated Time & Score Assistant to help fill out additional information. From my experience — especially in local competitions — we sometimes lack the manpower. I'd prefer to simplify scorekeeping rather than make it more complex.

One idea could be a dedicated time/action log, where an assistant notes key events (e.g., “0:10 entered arena,” “0:25 reached kitchen table”), and then fills out a simpler score sheet afterward.

Actually, writing this down makes me think it might be better to ensure we have an official video recording setup running. :sweat_smile:

SparkRibeiro21:
Not necessary anymore, since a lot of what was aim to be achieved with this PR, is already implemented in the new website. Such as statistics...

PR #995, opened 2025-09-28
First version of Receptionist 2026 task

SparkRibeiro21:
## Description

The goal is to make the Receptionist task much more natural task. 

New Updates for v2026
• Removed guest name had to be selected from a predetermined list of names. Now is any
popular English name.
• Removed favorite drink had to be selected from a predetermined list of objects (drinks). Now is any English named popular drink.
• Robots now have to check and inform the if the drink is not in the beverage area and move
on with the task
• Robots have more freedom to choose when to ask for the favorite drink and interest of the
guest.
• Guests will not meet the robot at starting point when they arrive, guests stay at the door.
The robot has to meet them at the door.
• Added doorbell sound when guest has arrived.(Generic audio classification)
• Added small penalty for all non essential questions that robot asks to help understand or
confirm information.
• Increased penalty for introduction with a wrong name, drink or interest.
• Decreased waiting time for door to be opened from 2mins to 1min.
• Alternative HRI penalty to understand guest, is now per information received and not
overall. (should have been changed in last years updates)
• MISSING: add list of possible special cases. (guest is not facing to the robot and talks
OR robot asks for name and person answers with favorite drink OR multiple guests at the
same time and they answer in random order, ...)
• Removed all mentions to points in text, so there is no difference between text and scoresheet that may lead to discussions

Notes for TC
• The host name and favorite drink is not used for anything.
• Webservice can be used as an interrupt.

For simplicity I added this pdf here, where everything in red is what was added and everything crossed out is what was removed. Sorry for the extra steps but personally i feel like this is the best way to analyze updates to existing tests. This way we don't miss anything.

[Receptionist_2026_v_0_1.pdf](https://github.com/user-attachments/files/22584615/Receptionist_2026_v_0_1.pdf)

### DIFF
--- a/scoresheets/Receptionist.tex
+++ b/scoresheets/Receptionist.tex
@@ -9,18 +9,20 @@
 
 	
 	\scoreheading{Bonus Rewards}
+	\scoreitem[2]{30}{Detect the doorbell sound as a signal that a guest has arrived}
 	\scoreitem[2]{200}{Open the entrance door for a guest}
 	\scoreitem[1]{50}{State a similarity between an interest between two or more persons}
 	\scoreitem[4]{30}{Describe the first guest to the second guest before reaching the living room (per correct visual attribute)}
 
 
 	\scoreheading{Penalties}
-	\penaltyitem[6]{20}{Wrong guest information was memorized (continue with wrong name, drink or interest)}
+	\penaltyitem[6]{40}{Wrong guest information was memorized (continue with wrong name, drink or interest)}
+	\penaltyitem[1]{10}{Non essential questions to help understand or confirm information (per question)}
 	\penaltyitem[4]{30}{Describe the first guest to the second guest before reaching the living room (per incorrect visual attribute)}
 	\penaltyitem[1]{120}{Introduce the wrong persons}
 	
 	\scoreheading{Deus Ex Machina}
-	\penaltyitem[2]{75}{Alternative HRI}
+	\penaltyitem[6]{20}{Alternative HRI}
 	\penaltyitem[2]{200}{Not recognizing people}
 
 	
--- a/tasks/Receptionist.tex
+++ b/tasks/Receptionist.tex
@@ -1,7 +1,8 @@
-\section{Receptionist}\label{test:receptionist}
+\section{Receptionist}
+\label{test:receptionist}
 
 \subsection*{Description}
-The robot has to guide two new guests to the living room, introduce them to each other, and offer them a seat.\\
+The robot has to take two new guests to the living room, offer a free place to sit and introduce them to eachother.\\
     
 \textbf{Main goal:}
     The robot welcomes and assists two newcomers at a party, offering them a seat and maintaining appropriate gaze direction during conversation (at person speaking, direction of navigation).
@@ -9,6 +10,7 @@ The robot has to guide two new guests to the living room, introduce them to each
 % \vspace{-5pt}
 \textbf{Optional goals:}
 \begin{enumerate}[nosep]
+	\item Detect doorbell sound as a signal that a guest has arrived.
 	\item Open the entrance door for each arriving guest.
 	\item Visually describe the first guest for the second guest before reaching the living room
 	\item Identify similarities between the guests and the host and incorporate them into the conversation.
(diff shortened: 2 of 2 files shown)


SparkRibeiro21:
Still missing the dialogue additions but made the following changes, as discussed at the 29th September TC meeting:

1) Removed interests
2) Since host was only used for interest (not used in introductions) and interest was removed, host was removed to simplify refereeing
3) Starting position is near where guests are sitting
4) not needing any confirmation or correction questions (non essential questions) awards bonus points instead of removing points
5) To get the offering drinks points, must tell guest if the favorite drink is in the beverage area and where it is, if not must state the list of all available drinks in the beverage area
6) In order for the doorbell to be played the robot must be waiting for a new guest at the start position  
7) passive guests (not the 2 being welcomed by the robot) may be anywhere in the arena including sitting and standing, in the sitting area
8) Improved clarification of how introductions must be made

discussed in TC meeting but not implemented:
- restructure guiding guests to beverage area - still needs to be further discussed...

adamgolding1:
Finding the drinks feels a bit awkward, if we take the person to the drinks area they can see the drinks themselves, maybe the robot can seat the guest and determine whether the drink they want is available, if the drink is not available then it can offer a suggestion? If the person wants to know where the bar is so they can see what is available for themselves then the robot can take them? Can discuss at TC meeting.

Minor typos.

Deus ex Machina -> Autonomous Skip

Score 
- replace Alternative HRI with Bypassing Automatic Speech Recognition

johaq (dismissed):

tasks/HumanRobotInteractionChallenge.tex
```diff
@@ -0,0 +1,139 @@
...
+\emph{\SysI{}}, \emph{\HRI{}}, \emph{\PerDet{}}, \emph{\PerRec{}}
+% human detection and recognition, safe navigation and human-robot interaction with unknown people.
+
+\subsection*{Setup}
+\begin{itemize}
+	\item \textbf{Location}: 
+	\begin{itemize}
+	\item The test takes place in the living room.	
+	\item The robot starts inside the \Arena{} at a predefined location near where guests are sitting.
+	\item A table with drinks is prepared near the living room.
```
LeroyR: We did remove the table stuff?

```suggestion

```

tasks/HumanRobotInteractionChallenge.tex
```diff
@@ -0,0 +1,139 @@
...
+It greets the guest and asks for their name and favorite drink.
+The robot escorts the guests to the living room and offers a free seat. 
+Once both guests are seated, the robot introduces them to each other. To introduce the two guests, the robot must (for both guests) look at a guest and state the other arriving guest name and favorite drink.
+The second guest will bring a bag for the host. The robot should ask for the bag to be handed over and move to the host. The host is already waiting and instructs the robot to follow him/her. The host will move to a random place in the house and inform the robot it can drop the bag there.   
+\begin{itemize}
+
+	\item \textbf{Doorbell:} A doorbell sound will be played when a guest is by the door ready to be met. The doorbell sound will not be provided beforehand and may be any traditional doorbell sound. As an alternative, if no doorbell sound is available, the guest can knock on the door. Waiting for the doorbell is optional, robots can bypass it and directly meet the guest at the door. In order for the doorbell sound to be considered detected, the robot must be waiting at the starting position and react instantly to the sound.
```
LeroyR: I still think this is kinda awkward regarding setup (Providing sound example, speaker system). Doorbell is one of those things a _deployed_ robot would just get wireless. I think just having the person knock make this much easier during the competition. 


IMHO being able to hear someone shouting at the robot and perceiving environment sound is something we can all imagine our ideal robot to have.  I think we should reintroduce stuff like sound source localization in next years revamped HRI task.
Currently the Robot are mostly tuned for interacting with close persons (and filtering the loud competition background noise) i think 30 Points are not going the influence the design of the robots.

Anyway ill approve this task regardless of choice here.

SparkRibeiro21:
> The bag idea is fine, but I would allow only a natural handover or bypass it (e.g., placing the bag directly in the gripper). Placing it on the floor and pointing at it seems odd.

This has already been removed before your comment...

scoresheets/HumanRobotInteractionChallenge.tex
```diff
@@ -0,0 +1,40 @@
...
+	\scoreitem{50}{Drop the bag on the correct area}
+	
+	\scoreheading{Bonus Rewards}
+	\scoreitem[2]{30}{Detect the doorbell sound as a signal that a guest has arrived}
+	\scoreitem[2]{200}{Open the entrance door for a guest}
+	\scoreitem[4]{20}{Describe the first guest to the second guest (per correct visual attribute)}
+	\scoreitem[4]{15}{Not asking non essential questions to confirm or correct information}
+
+	\scoreheading{Penalties}
+	\penaltyitem[4]{40}{Wrong guest information was memorized (continue with wrong name, drink or interest)}
```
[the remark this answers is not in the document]
SparkRibeiro21: Hi Leroy, i think I have now understood your perspective.
Following what was decided in a meeting: to have a higher penalty if you say a wrong name/drink than the points you gain.
I think the clearest way is to do as follows:

Regarding points gained, i think this is quite clear as is:

> \scoreitem[4]{30}{Introduce both guests to each other: correctly say name and favorite drink of each guest}
> \scoreitem[2]{50}{While introducing guests, look to the correct guest while talking about the other guest}
	
One score item for name and drink and a separate score item for looking at the correct guest. So you get +30 for every name and drink you get write and independently of that if you look at the correct person you also get +50 points

However, regarding the penalties:

> \penaltyitem[1]{120}{Introduce the wrong persons}
	
I will remove this as you are correct it does not make much sense. 

 > \penaltyitem[4]{40}{Wrong guest information was memorized (continue with wrong name, drink or interest)}

This will stay the same, so if teams decide to skip a name rather than say a wrong name, they get 0 points instead of -40.

However if any robot looks at the wrong person when saying the the other guest name and favourite drink. No penalty is applied, you just don't get any points for looking at the correct guest.
Do you feel this is clear now?

PR #998, opened 2025-10-06
First draft of GPSR 2026

adamgolding1:
Updating the GPRS Task
- Update section in Procedure for Command execution to include giving the instructions consecutively or one-by-one
- Update section in Additional Rules for passing command through LLM to get similar command
- Add to Deus ex Machina QR code will not be available

Question
If the commands are executed consecutively, how is the arena re-arranged between tasks?

### DIFF
--- a/tasks/GPSR.tex
+++ b/tasks/GPSR.tex
@@ -41,8 +41,9 @@ The robot is asked to understand and execute commands requiring a wide range of 
 \subsection*{Procedure}
 \begin{enumerate}[nosep]
 	\item \textbf{Instruction point:} At least two hours before the test, the referees announce the location of the \textit{Instruction Point}.
+	\item \textbf{Command generation:} Each command will be generated using the Command Generator. Once a command has been generated it will be entered into an LLM to re-generate a similar phrase.
 	\item \textbf{Test start:} The robot moves to the \textit{Instruction Point} when the arena door is open.
-	\item \textbf{Command execution:} The operator provides up to three commands issued consecutively to instruct the robot to perform the tasks. If the operator provides the commands one-by-one, the robot must return to the operator after completing each task. Once a command has been generated it will be entered into an LLM to re-generate a similar phrase. The operator will say "please start now" to initiate the task.
+	\item \textbf{Command execution:} The operator provides up to three commands issued consecutively to instruct the robot to perform the tasks. If the operator provides the commands one-by-one, the robot must return to the operator after completing each task. The operator will say "please start now" to initiate the task.
 	\item \textbf{Back to the instruction point:} The robot goes back to the \textit{Instruction Point} after completing all of the commands given by the operator.
 	\item \textbf{Pausing the Timer:} If commands are issued one-by-one, the referee pauses the timer as soon as the robot reaches the instruction point to reset the arena for the next command. The timer resumes once the referee signals the start of the next command.
 \end{enumerate}
--- a/tasks/GPSR.tex
+++ b/tasks/GPSR.tex
@@ -19,7 +19,7 @@ The robot is asked to understand and execute commands requiring a wide range of 
 \begin{itemize}[nosep]
     \item \textbf{Locations:}
         \begin{itemize}
-            \item \textbf{Task location:} The task takes place inside the \Arena{}. Commands may require the robot to leave the room. The \Arena{} is in its nominal configuration for this task.
+            \item \textbf{Task location:} The task takes place inside the \Arena{}. The \Arena{} is in its nominal configuration for this task. Commands may require the robot to leave the room.
             \item \textbf{Start location:} The robot starts outside the \Arena{}. When the door opens, it must navigate towards the \textit{Instruction Point}.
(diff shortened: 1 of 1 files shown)


ARTenshi:
As a reminder, the rulebook from 2023 and before already included:

"commands can be issued by a Naive Operator, i.e. a person from the audience with no background on robotics. The referee gives the command to the Naive Operator, who will then issue it to the robot (rephrasing is allowed). If the robot consistently fails to understand the naive operator (e.g. 3 times or more), teams can default to a custom operator."

"Commands for EGPSR are either Stage II tasks, complex commands requiring to perform chains of subtasks in sequence, or incomplete commands lacking relevant information to succeed."

Since rephrasing and multiple command attempts were already included in previous editions, I'm concerned this may not advance the competition's technical challenges.

A more challenging approach would be giving additional tasks while the robot is performing the first command. For example, the robot receives the first command at the instruction point; while executing it, a user calls the robot and requests a second task. The robot must then decide the order of execution—whether to finish the current task or pause it to handle the new request. The user (or an additional user) could then request a third task, requiring dynamic priority management.

Teams that can only handle one task at a time would complete their run after the first task (suitable for beginner teams), while more advanced teams could attempt to handle multiple concurrent requests for additional points.

Keeping the competition accessible to new teams remains important, but I think we should move forward on the challenges we proposed and not going back to what we have already done.

LeroyR:
I assume "going back" => returning to previously solved tasks but while i could see not having enough forward movement i dont agree that this is "going back".

> As a reminder, the rulebook from 2023 and before already included:
> 
> "commands can be issued by a Naive Operator, i.e. a person from the audience with no background on robotics. The referee gives the command to the Naive Operator, who will then issue it to the robot (rephrasing is allowed). If the robot consistently fails to understand the naive operator (e.g. 3 times or more), teams can default to a custom operator."

Which was Optional. "random/llm" rephrasing now mandatory. Previously you could generate every possible combination of commands for parsing.

> 
> "Commands for EGPSR are either Stage II tasks, complex commands requiring to perform chains of subtasks in sequence, or incomplete commands lacking relevant information to succeed."
> 
> Since rephrasing and multiple command attempts were already included in previous editions, I'm concerned this may not advance the competition's technical challenges.

Complex Subtasks and incomplete Task where not in the generator for the last 3 years. The incomplete commands (before that) never used or solved.
Rephrasing was Optional (now Mandatory).

> 
> A more challenging approach would be giving additional tasks while the robot is performing the first command. For example, the robot receives the first command at the instruction point; while executing it, a user calls the robot and requests a second task. The robot must then decide the order of execution—whether to finish the current task or pause it to handle the new request. The user (or an additional user) could then request a third task, requiring dynamic priority management.
> 

Optional planning/merging of multiple tasks would be new. 
Previous _Optionals_ into _Required_ is the step forward. 
We also decided to revamp the generated commands which may increase the execution difficulty. 
Other changes in the Rulebook are already massive which is why this task, in general, will not change too much.


> Teams that can only handle one task at a time would complete their run after the first task (suitable for beginner teams), while more advanced teams could attempt to handle multiple concurrent requests for additional points.
> 
> Keeping the competition accessible to new teams remains important, but I think we should move forward on the challenges we proposed and not going back to what we have already done.

LeroyR:
> Just to clarify, those "other changes" mostly concern the competition format, such as drastically reducing the number of tasks and allowing them to be repeated over several days. This simplification in the format should allow us to introduce more challenging tasks rather than "keeping" them as they are (with some improvements given the feedback from previous iterations).

While i personally agree (and also suggested more new tasks, multiuser, speech interruptions etc), the consensus was to not change e.g. receptionist substantially to avoid too many changes at once. I think adding a planning bonus here (give all commands instead of one at a time) while generating more interesting commands ensures this task is still interesting enough.

tasks/GPSR.tex
```diff
@@ -19,18 +14,14 @@ \subsection*{Setup}
...
     \item \textbf{People:}
         \begin{itemize}
-			\item \textbf{Professional Operator:} Typically the referee, issues standard commands to the robot.
-			\item \textbf{Non-Expert Operator (optional):} A person without robotics expertise (e.g., from the audience). The referee provides the intended command goal to this operator, who then issues the instruction using their own wording.
-
-            For example, the generated command might be \textit{"Bring me a coke from the kitchen."} then the non-expert operator will be told \textit{"The robot should bring you a coke, which is found in the kitchen."}, who then tells the robot \textit{I want a coke. Go to the kitchen and get me one."}
-
-            If the robot consistently fails to understand the non-expert operator (e.g. after two retries), teams can default to a custom operator.
+			\item \textbf{Professional Operator:} The referee issues standard commands to the robot.
+            If the robot consistently fails to understand after three retries, teams can default to a custom operator.
```
johaq: Three tries in total, so should be three tries or two retries. I think three tries is clearer

TKang-P:
> I think it is not clear enough how giving all commands at once works. Maybe we should include an example. Should there maybe be an advantage (in score) for getting all commands at once vs one by one?

Yes, I agree with this. Let the team choose it and give them a bonus for both correctly demonstrating plans and executing all three commands.

PR #1000, opened 2025-10-26
First draft of Restaurant 2026

TKang-P:
## Description
Updates for Restaurant 2026
To encourage teams to use barman rather than being stuck in the middle of the restaurant.
- increased the time and count to instruct the Barman from one minute for the task to two minutes per order.
- add some examples of instructions
- reduced the penalty score of "Being guided to a table"
- add penalty "Asking for directional confirmation" due to no awareness

I didn't include the barman's guidance/pointing procedure that automatically executed after 1 minute.
Because it will not change the robot's behavior and only make it look worse.

### DIFF
--- a/scoresheets/Restaurant.tex
+++ b/scoresheets/Restaurant.tex
@@ -5,6 +5,7 @@
 
 	\scoreitem[2]{200}{Understand and confirm the order received to the customer}
 	\scoreitem[2]{100}{Communicate the order to the barman}
+	\scoreitem[2]{200}{Picking up the requested item from the \textit{Kitchen-bar}}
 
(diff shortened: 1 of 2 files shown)


adamgolding1:
I think this looks ok, just a minor adjustment to replace the term Deus ex Machina to be consistent with new terminology

SparkRibeiro21 (approved):
Awesome :)

ARTenshi:
The term "facing the customer seating area" is ambiguous and potentially misleading because the robot’s field of view is limited. Even if the robot starts facing a few tables, customers may be seated at any table, including those outside its initial field of vision.

The starting position is usually next to the counter that may or may not be facing the costumers area.

PR #1021, opened 2026-01-13
Improve scoresheets

LeroyR:
## Description

- Improve scoresheets by introducing `scoremod` and `scorepen` commands.
  - add indented score modifies as used in pick and place
- Apply penalties as modifications to scoreitems
- Remove bonus goal rulings: no longer require "Main Goals" before "bonus goals". Rename to Extra goals in scoresheet.

## Other comments
This just marks the dependency of penalties that only apply to specific score items.
We already apply most penalties only to specific score items so you dont lose points for attempting to solve parts of the task. Example:
image
If the robot is not reaching the drop area and drops the bag this makes it more clear that the attempt does not result in -50. :smiley:

### DIFF
--- a/pages/general_rules/ContinueRules.tex
+++ b/pages/general_rules/ContinueRules.tex
@@ -1,70 +1,82 @@
-\section[Deus Ex Machina]{Deus Ex Machina: Bypassing Features With Human Help \\ \small Because the Show Must Go On}\label{rule:continue}
+\section[Human Assistance]{Bypassing Features With Human Help \\ \small Because the Show Must Go On}\label{rule:continue}
 
-Robots can't score unless they accomplish the main goal of a task.
-However, in many real-life situations, a minor malfunction may prevent the robot from accomplishing a task.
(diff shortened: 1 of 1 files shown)


scoresheets/DoingLaundry.tex
```diff
@@ -1,24 +1,37 @@
...
+      }
     \scoreitem{300}{Using the basket for transportation}
-    \scoreitem[6]{300}{Folding additional clothes (per item)}
-    \scoreitem[6]{200}{Stacking folded clothes neatly (per item)}
+      \scorepen{200}{%
+        Laundry is dropped or otherwise lost during transportation.
+      }
+    \scoreitem[5]{400}{Folding additional clothes (per item)}
+      \scorepen[5]{200}{%
+        Human assistance: Flattening/Arrange clothing before folding
```
johaq: see above

johaq (changes requested):
Looks good overall. Some small things

PR #1025, opened 2026-01-14
Updates to Pick and Place Challenge

SparkRibeiro21:
## Changes proposed in this pull request:
- Added auxiliary table with two common items (worldwide available set). Robots may choose to pick these (simpler) instead of the objects in the dining table and store them in the cabinet with a small penalty. 
- Added human assistance penalty for requesting environment changes (move chairs or decorations). Penalty is per item moved (per chair or per set of decorations)


SparkRibeiro21:
Regarding how we will fix the side table to the ground, I would say, the most important thing is not actually the table being taped to the ground, but it being in the same place for all teams, so probably some marks on the ground and making sure the table is in the correct place is enough.,

PR #1046, opened 2026-05-04
Restaurant updates from Portugal Open

SparkRibeiro21:
Updated the picking and placing scoring items from 2x200 to 4x100
20% reduction on all non manipulation tasks
Added +100 bonus for first picking and first placing (same as in PP to motivate teams to attempt manipulation)
Added scoresheet penalty for using alternative HRI for understanding the order (it was in the rulebook but not in scoresheet, hidden rule now is clearer)
Removed legacy remark of three items in order.

### DIFF
--- a/tasks/Restaurant.tex
+++ b/tasks/Restaurant.tex
@@ -68,6 +68,7 @@ Teams are allowed a single restart during the restaurant task. The following rul
 
 \subsection*{Additional rules and remarks}
 \begin{itemize}
+	\item \textbf{Alternative HRI:} Using an alternative HRI to understand an order causes a score reduction.
 	\item \textbf{Remarks:}
 	\begin{itemize}
 		\item This test occurs in a public area. Any physical contact with people or furniture results in immediate emergency stop.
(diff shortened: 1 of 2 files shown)


LeroyR (changes requested):
We need to also put the remarks regarding first pick/place into the task description

from PP:
```
First Pick Bonus: To encourage manipulation, the robot receives an additional bonus
for successfully picking the first object during the test. This bonus is awarded only once
```

slightly Off topic: Reading this it feels like the reasoning why this is here `To encourage manipulation,` should be a comment or something and not actually render into the rules

SparkRibeiro21:
> We need to also put the remarks regarding first pick/place into the task description
> 
> from PP:
> 
> ```
> First Pick Bonus: To encourage manipulation, the robot receives an additional bonus
> for successfully picking the first object during the test. This bonus is awarded only once
> ```
> 
> slightly Off topic: Reading this it feels like the reasoning why this is here `To encourage manipulation,` should be a comment or something and not actually render into the rules

Added this:

\item \textbf{First Pick/Place Bonus:} The robot receives an additional bonus for successfully picking an placing the first object during the test. This bonus is awarded only once.

(This way we do not talk about encouraging manipulation)

LeroyR (dismissed):

tasks/Restaurant.tex
```diff
@@ -68,6 +68,8 @@ \subsection{Restart}
 
 \subsection*{Additional rules and remarks}
 \begin{itemize}
+	\item \textbf{First Pick/Place Bonus:} The robot receives an additional bonus for successfully picking an placing the first object during the test. This bonus is awarded only once.
```
johaq: and placing

johaq (dismissed):
Small typo otherwise good

TKang-P (dismissed):

PR #1051, opened 2026-06-19
Introduce humanoid robot rules and safety crane guidelines

myszzka96:
Added rule for humanoid robots and safety crane usage in RoboCup@Home.

## Description

Due to some of the humanoids being a safety concern due to size weight this year, I propose to allow restricted use of cranes during competition. It has been discussed in early May  and this is a small change to accommodate that. 

- Allow for a crane with a team member as an operator

### DIFF
--- a/pages/general_rules/Robots.tex
+++ b/pages/general_rules/Robots.tex
@@ -21,6 +21,10 @@ Robots competing in the RoboCup@Home League must comply with security specificat
 \end{enumerate}
 The compliance with these rules will be verified during \RobotInspection{} (see \ref{sec:robot_inspection}).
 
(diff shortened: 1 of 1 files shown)


LeroyR:
Imho if the robot is not able to operate safely by itself it should not be allowed to participate.

If humanoids need special treatment they need their own, separate league.

I understand that this is required as people do not want to risk damaging the environment or the robot and we want to participate in the current humanoid wave. 
But why do they need a special crane operator? The crane system can just follow the robot. Touching should be forbidden. 

Just build a wheeled robot 🙃

myszzka96:
The crane looks something like this (https://funduinoshop.com/media/image/9a/b5/08/Unitree_Kran_Gesamt_600x600.png). The robot cannot actually pull it and it would not turn out very well. We do not have a full crane system over the arena aviable on site, unfortunatelly. All the operator is doing is pushing it behind the robot.

LeroyR:
Well I'd say it is in the teams responsibility to come up with solutions so they can show autonomous robots. 

I have neither a ceiling system nor is a crane operator included in my home. 

And at home s should show the current state of robots operating in my home, not in a lab with a crane.

johaq (approved):
For the record, I 100% agree with @LeroyR

PR #1056, opened 2026-08-18
Time and logic optimizations to restart rule

SparkRibeiro21:
## Description

The goal is to update the restart rule, removing some of its chaos of having to move outside the house, for both the team doing the task, the next team already setting up the robot, and all teams still in line. In addition, greatly improve time efficiency and reduce dead times between teams task attempts. 

## Changes proposed in this pull request:
- During a restart the robot no longer has to be moved outside the house. Now it is inside the house near the entrance.
- Removes the comotion of having to move the robot outside the house, and all the teams next in line have to move back whenever a restart happens. 
- Since no robot will move outside the house during a restart, the team next in line to compete, may start its setup, as soon as the previous team enter the arena, greatly reducing the setup time needed. Previously teams kind of had to wait for the robot to finish the task to start their setup.
- As soon as a task finishes, the referee will inform the next team it has 30 seconds to finish the setup. Only after will the referee do the referee discussion and arena preparation for the next task. Once the 30 seconds have passed (or the time necessary for the arena preparation if >30 seconds) the referee will move to the door and open the door.
- Significantly improves the dead times between tasks


pages/general_rules/Procedure.tex
```diff
@@ -116,7 +116,7 @@ \subsection{Restart Rule}\label{rule:restart}
 \begin{itemize}
     \item A restart may only be requested until the robot scores 40 points.
     \item The test time continues to run during the restart.
-    \item The team moves the robot back to the starting location.
+    \item The team moves the robot to the entrance (inside the house).
```
MrKeith99: To keep consistence in the rulesbook we should use same terminology.
```suggestion
    \item The team moves the robot to the entrance (inside the arena).
```

pages/general_rules/Procedure.tex
```diff
@@ -179,7 +179,7 @@ \subsection{Time limits}\label{rule:time_limits}
 	\item \textbf{Requesting time:} A robot (not the team) can request the referees to extend the inactivity time limit for a time-consuming process (e.g., 60~seconds). This time cannot exceed 3 minutes and can only be used once per test.
 
 	\item \textbf{Setup time:} Unless stated otherwise, there is no setup time.
-	Robots need to be ready to enter the \Arena{} no later than one minute after the door has been closed to the previous team.
+	Robots need to be ready to enter the \Arena{} no later than \SI{30}{\second} after the previous team ended their task. Teams can start their setup as soon as the previous team has entered the arena and the door has been closed.
```
MrKeith99, line 182: I have forgot to talk about this during the TC meeting, but this is not applicable to the current HRI task or tasks where the robot starts inside the arena.
Since it is a general rule, we can include the exception to specific tasks which do not start from entrance.

SparkRibeiro21: Yeah, this is a good point. 
On the first meeting, I asked to add a note: that for all tasks, we need to make an effort to standardise the task starting procedure, so it always starts outside the door. Otherwise, this problem still exists. It might not be optimal for the story logic of the task in some cases, but I believe this should be standard for all tasks. Robots always have to start outside and wait for door to be opened. 
If for some reason we don't do this for a task, I agree, the starting procedure for that specific task should be on the task itself.

johaq: Yes starting in front of the door will be a requirement for all tests

johaq (dismissed):

LeroyR (dismissed):

TKang-P (dismissed):

